PR #1, opened 2016-02-08
supermicro/x10sll-f: init

abbradar:
This adds necessary kernel modules to control IPMI and hardware random generator from TPM. I also have a watchdog module enabled, but here it is commented because enabling it by default is not a good idea IMO. I left it as an example, however.


ehmry:
Yea, I still haven't figured out how to use my watchdog.

PR #5, opened 2016-02-10
USB Armory configuration

washort:


ehmry:
Badass, I'm going to try it on my when I get the time.

PR #6, opened 2016-02-10
Add power-saving configuration for Lenovo T410.

peti:


lenovo/t410.nix
```diff
@@ -0,0 +1,77 @@
+{ config, pkgs, ... }:
+
```
Ericson2314: Missing `{`?

ehmry:
I did a quick force push, and I made an issue about this:
https://github.com/ehmry/nixos-hardware/issues/7

PR #9, opened 2016-02-24
WIP: Dell XPS 15 9550

Ericson2314:
Here is a dump of everything I know about getting my laptop running. It's highly unstructured however---and I was the one who emailed the list warning against precisely this lack of structure. So in the interest of not being a hypocrite :), let me know the best we to organize this.

I imagine that the 13" variant Dell xps-13-9350 would most/share all of these settings, there is also a question of how that should be handled.

### DIFF
--- a/dell/xps-15-9550.nix
+++ b/dell/xps-15-9550.nix
@@ -1,0 +1,35 @@
+{ config, pkgs, ... }:
+
+{
+  # BEGIN from generated hardware-configuration
+  imports =
+    [ <nixos/modules/hardware/network/broadcom-43xx.nix>
+      <nixpkgs/nixos/modules/installer/scan/not-detected.nix>
+      ../lib/kernel-version.nix
+    ];
+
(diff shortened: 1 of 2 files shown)


ehmry:
I don't think we should be too picky on structure for now, so long as we are making forward progress.

I would prefer that the kernel version not be pegged, we need some sort of `versionAtLeast` function.

ehmry:
Ok, I created a new module at `lib/kernel-version.nix`, it works like this:
```
{
  imports = [ ../lib/kernel-version.nix ];
  kernelAtleast =
    [ { version = "4.4"; msg = "Kernel does not support graphics hardware"; }
      { version = "4.5"; msg = "Kernel does not support audio hardware"; }
    ];
}
```

When the version at boot.kernelPackages is too low, a warning is printed during evaluation.

Could you use this for the trailing section of the module, or I am I missing something?

Ericson2314:
@ehmry That looks absolutely wonderful! I will gladly switch to using it.

Ericson2314:
@ehmry Ok this should be g2g! Used `kernelAtLeast` and moved comment to wiki file.

PR #25, opened 2017-09-03
General thinkpad configuration

Ma27:
__Note:__ this is just a WIP as this requires some testing against several machines, furthermore I'd like to deduplicate even more configurations.

However I'm happy about any feedback :-)


Ma27:
Just another thought: I'd even remove files completely if `general.nix` is enough hardware customization. It should be possible to load `generic.nix` by default and create a custom file for a model if needed.

lenovo/thinkpad/general.nix
```diff
@@ -0,0 +1,18 @@
...
+    enable = true;
+    emulateWheel = true;
+  };
+
+  hardware.enableAllFirmware = true;
+  services.tlp.enable = true;
+
+  services.xserver = {
+    synaptics.enable = false;
+    libinput.enable = true;
```
Profpatsch: libinput is questionable. I disabled it again on my machine, because it was buggy.

lenovo/thinkpad/general.nix
```diff
@@ -0,0 +1,18 @@
...
+
+with lib;
+
+{
+  hardware.trackpoint = mkDefault {
+    enable = true;
+    emulateWheel = true;
+  };
+
+  hardware.enableAllFirmware = true;
```
Profpatsch: Not sure if this should be a default. At least overridable (`mkDefault`). The modules should probably add hardware options more selectively.

Profpatsch: At the *very* least, `enableRedistributableFirmware` should be used to avoid unfree package by default.

Profpatsch:
Also compare https://github.com/openlab-aux/vuizvui/blob/master/modules/hardware/thinkpad.nix

Ericson2314:
CC @hSloan

lenovo/thinkpad/general.nix
Ma27: ah good to know. I've seen `enableAllFirmware` in other modules and it works on my ThinkPad, so I kept it for now :-)

lenovo/thinkpad/general.nix
Ma27: well, I have to admit that I'm currently using synaptics as well (because I didn't manage to get kinetic scrolling working, which is perfectly fine with synaptics).
However synaptics is deprecated in NixOS, so I changed it here and will try a migration to libinput on my machine in the next days...

lenovo/thinkpad/x250.nix
```diff
@@ -0,0 +1,5 @@
+{ config, pkgs, ... }:
+
+{
+  imports = [ ./general-intel.nix ];
+}
```
Ma27: I'd vote for removing `x250.nix` as all of its values are general defaults.
Only if a model requires special customizations, a special file should be added, right?

lenovo/thinkpad/general.nix
Ma27: ok, I switched my machine to redistributable firmware only and it seems to work as well :-)

lenovo/thinkpad/x250.nix
garbas: 👍

Ma27:
btw it would be very cool if some people could review/test this and tell me in this thread which devices still work or which improvements can be build.
Thanks in advance :-)

lenovo/thinkpad/general.nix
Ma27: @Profpatsch do you think we should work with `mkDefault false`?

Ma27:
@Ericson2314 @Profpatsch would be cool if you do do another review, I'm not 100% sure about the remaining parameters, so I could need some feedback :-)
(We can talk about this in person at NixCon as well if you attend...)

lenovo/thinkpad/general.nix
```diff
@@ -0,0 +1,22 @@
...
+  services.tlp.enable = true;
+
+  services.xserver = {
+    synaptics.enable = false;
+    libinput.enable = true;
+  };
+
+  environment.systemPackages = [ pkgs.acpi ];
+
+  sound.enableMediaKeys = true;
```
Profpatsch: Should probably also be `mkDefault`.

Ma27:
@Profpatsch absolutely agree with you regarding the `mkDefault` usage.

I actually agree with you, I don't like `libinput` either, but IIRC it's recommended over our packaged `synaptics` driver as it's deprecated: https://github.com/NixOS/nixpkgs/commit/9d2777a5a5f861a727d09c3ecdeffd8c72c143b7

lenovo/thinkpad/general.nix
```diff
@@ -12,11 +12,11 @@ with lib;
   services.tlp.enable = true;
 
   services.xserver = {
-    synaptics.enable = false;
-    libinput.enable = true;
+    synaptics.enable = mkDefault false;
+    libinput.enable = mkDefault true;
```
Profpatsch: Those two should be in one `mkDefault`, so that either both get set or neither, I’d say.

Ma27: done

Profpatsch:
cc @ehmry @garbas Merge?

lenovo/thinkpad/general-intel.nix
```diff
@@ -0,0 +1,6 @@
+{ ... }:
+
+{
+  boot.kernelModules = mkDefault [ "kvm-intel" ];
+  services.xserver.videoDrivers = [ "intel" ];
```
peterhoeg: The ```intel``` driver is deprecated and all the distributions have moved to ```modesetting```.

Mic92: on x250 I still have freezes with `modsetting`

peterhoeg: Should we then ```mkDefault``` then for those who do want to use modesetting without having to ```mkForce```?

Mic92: I would only set that this for x250 like it was before with the comment.

Ma27:
I currently use the same configuration on x270.
I'll test the non-deprecated config today to see if your hardware is the obly one and report what happened tonight:)

PR #28, opened 2017-12-22
treewide: drop hardware-configuration.nix boilerplate

yegortimoshenko:
hardware-configuration.nix is still relied upon for hard drive configuration, and it handles variations of hardware (custom configurations, subrevisions).

### DIFF
--- a/apple/macbook-pro-10-1.nix
+++ b/apple/macbook-pro-10-1.nix
@@ -5,20 +5,6 @@
     [ ../lib/kernel-version.nix
     ];
 
-  ## BEGIN from generated hardware-configuration
-  ## Probably better to just use a freshly generated hardware.configuration.nix
-  ## than this, but including for reference.
-  #imports =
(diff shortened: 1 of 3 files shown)


zimbatm:
sounds good

Ericson2314:
Just for the record, my idea with this was a sanity check: the hardware detection is by no means immune to regressions, it's good to have compare to a known-good state.

yegortimoshenko:
@Ericson2314 Perhaps you're right, but:

* model name can't cover all customizations that any specific machine might or might not have (e.g. fingerprint sensor, BIOS disabled devices)
* we might decide that some other module is better for the hardware at hand (e.g. out-of-tree `tp_smapi` vs. `thinkpad-acpi`), or kernel module name might change
* hardware configs currently generated by `nixos-generate-config` are by no means complete: if that were the case, this repo would have no meaning; which means, it's not only perceptible to regressions, but also to missing/buggy hardware detection, i.e. known-good state might not actually be good, and probably isn't, newer `nixos-generate-config` will probably identify more hardware and have less bugs

PR #29, opened 2017-12-22
treewide: modules -> profiles, separate dirs, list in README

yegortimoshenko:
```
Enforce proper naming: all paths are lowercase and hyphen-separated,
if there's a line of models (aspire, macbook-pro, thinkpad) it becomes
a subdirectory. Documentation for profiles is moved to README files in
respective directories.

Add an Org mode table that lists all available profiles and their paths.

Instead of fetching repo locally, use a Nix channel. Making hardware
profiles read-only should improve quality and amount of participation
long-term.
```

Preview how new README looks like: https://github.com/yegortimoshenko/nixos-hardware/blob/33d61be7e5935f1ee7bdc05c11e66387c9391058/README.org

Preview for per-profile README: https://github.com/yegortimoshenko/nixos-hardware/tree/33d61be7e5935f1ee7bdc05c11e66387c9391058/apple/macbook-pro/10-1


yegortimoshenko:
This is a major breaking change but I don't think it's important because it's not on a channel, most users have probably cloned the repo and made a bunch of changes. I think this should be done in order for `nixos-hardware` to be able to grow further.

Issue #35, opened 2017-12-24
Fix evaluation treewide


yegortimoshenko:
I've added a test to see if the tree evaluates properly: fc9b7ae0024066130c9cc3b699e8b103e7c21d62.

Evaluation should be fixed for all profiles in the tree.

yegortimoshenko:
Fixed via 721741b0c79b190259e415ea1606ced6744ee199, 95c8873c41ef57d820b027e16db0709c050579e1, f62ead1ce0874c3773fb8d762e5de3b336e8c4ac, 0344137b57a5429a8ac49557be38a806ed30ba3b, 63b3836fcf0ff03b5f96b173f5c06710f08b2c3e, a8b7b6ffe5813545e289404e504a21e2222f7ab5, dd87817ffc7c60929949183cdf9dc88c96e4bdb6, 667b1d8eb9c23d1de433d71caea5648a5acabadb, 097d47e3f80f3a24462669a05696065128773bdb.

PR #37, opened 2017-12-24
lib: drop, fixes #16

### LINKED ISSUE #16: Move inline hardwareNotes to README
https://github.com/NixOS/nixos-hardware/blob/master/lib/hardware-notes.nix#L21

Needs a description, it’s not very obvious what this is for (apart maybe from searching for examples).

yegortimoshenko:
Hardware notes are seemingly less useful than a `README`, and `kernelAtLeast` option, while potentially useful:

a) is not used anywhere in the tree at the moment,
b) doesn't properly work when there are multiple elements in the list,
c) assumes that there are no degradations ever.


yegortimoshenko:
@zimbatm `inversepath/usbarmory` is broken: it doesn't even evaluate (`host.nix` references to non-existing NixOS options, `default.nix` has a lot of hardcoded paths so that even if it did evaluate, it wouldn't work.

I've fixed evaluation in all profiles except `raspberry-pi/2` and `inversepath/usbarmory` in #35.

PR #38, opened 2017-12-25
add lenovo x250

Mic92:
Basically a meta profile that combines thinkpad + intel.

### DIFF
--- a/README.org
+++ b/README.org
@@ -30,6 +30,7 @@ should look like:
 | Lenovo ThinkPad T460s     | ~<nixos-hardware/lenovo/thinkpad/t460s>~   |
 | Lenovo ThinkPad X140e     | ~<nixos-hardware/lenovo/thinkpad/x140e>~   |
 | Lenovo ThinkPad X220      | ~<nixos-hardware/lenovo/thinkpad/x220>~    |
+| Lenovo ThinkPad X250      | ~<nixos-hardware/lenovo/thinkpad/x250>~    |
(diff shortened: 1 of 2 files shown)


zimbatm:
@Mic92 can you add a `users: Mic92` comment? That will help other users ping you in case they want to change the profile.

Mic92:
In nixos we also have `meta.maintainers` for nixos modules

yegortimoshenko:
Perhaps it makes sense to only enable `services.thinkfan` for ThinkPads where fan control doesn't work properly out of the box.

zimbatm:
@yegortimoshenko I told you, generalization doesn't work on this repo. Only users of a specific hardware can test it properly.

yegortimoshenko:
@zimbatm I disagree: apparently it works pretty well, this profile inherits ThinkPad profile, generic laptop profile, generic PC profile, and Intel CPU profile. And there is only option that is overrided specifically in this new profile, which also might go away and moved into model profiles if we decide that many ThinkPads have decent fan control by default.

zimbatm:
The problem is not right now, it's when you want to change a setting in one of the generalized profiles. This will inevitably lead to changes that haven't be tested on the affected hardware.

yegortimoshenko:
@zimbatm It does work on @Mic92's hardware, that daemon just might not be beneficial in case of ThinkPad X250.

zimbatm:
It's fine, only the future will tell if my prediction are true. If you keep steering this repo it might also not happen as much as I expect it to be.

PR #39, opened 2017-12-25
README: add links to docs

zimbatm:


yegortimoshenko:
Sure, but I would link the model name itself to link to the docs instead of adding another column. I'm going to rebase that on top of master.

zimbatm:
Good idea. Any idea on how to implement it without destroying the ASCII formatting of the table?

yegortimoshenko:
Merged in 008eb393030b414953596e2434c8b80c3762b10b.

yegortimoshenko:
@zimbatm: I don't think there is a way to do that without destroying ASCII formatting, but if you use Org mode, at least Emacs version hides link formatting, and it looks nice. There is also a Vim port but I don't know if it does that or not.

zimbatm:
Ack. As a note, I quite like the AsciiDoctor format because it's close to MarkDown but it better specified and has support for all the useful things like TOC, tables, link references, ..

yegortimoshenko:
I'm going to check it out. Thanks!

PR #40, opened 2017-12-25
add Travis-CI to the repo

zimbatm:
todo:
* [ ] setup Travis-CI on the repo
* [ ] test that it works

### DIFF
--- a/tests/eval-test.sh
+++ b/tests/eval-test.sh
@@ -7,10 +7,11 @@ find=(find . -name default.nix)
 for profile in `${find[@]}`; do
     echo evaluating $profile >&2
 
-    nixos-rebuild \
-	-I nixos-config=tests/eval-test.nix \
-	-I nixos-hardware-profile=$profile \
-	dry-build
+    nix-build '<nixpkgs/nixos>' \
+        -I nixos-config=tests/eval-test.nix \
+        -I nixos-hardware-profile=$profile \
+        -A system \
+        --dry-run
 
     if [ $? -ne 0 ]; then
 	exit 1
--- a/tests/eval-test.sh
+++ b/tests/eval-test.sh
@@ -1,19 +1,15 @@
 #!/bin/sh
 
-cd $(dirname $0)/..
+set -e
 
-find=(find . -name default.nix)
+cd "$(dirname "$0")/.."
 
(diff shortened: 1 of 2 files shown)


yegortimoshenko:
I don't think it would work because this test needs NixOS, and I don't think Travis CI uses NixOS (they are likely to use something like Debian with Nix installed).

tests/eval-test.sh
```diff
@@ -1,18 +1,15 @@
...
 for profile in `${find[@]}`; do
-    echo evaluating $profile >&2
+  echo evaluating $profile >&2
 
-    nixos-rebuild \
-	-I nixos-config=tests/eval-test.nix \
-	-I nixos-hardware-profile=$profile \
-	dry-build
-
-    if [ $? -ne 0 ]; then
```
yegortimoshenko: Without this, test would never fail.

zimbatm: `set -e` takes care of this

zimbatm:
fixed the test script so that it works on any system that has nix installed

Mic92:
In doubt, it is possible to create VMs on non-nixos systems: `nix-build '<nixpkgs/nixos>' -A vm -k -I nixos-config=./vm.nix`

tests/eval-test.sh
yegortimoshenko: Great! Didn't know about `set -e`.

yegortimoshenko:
I think shell scripts should be portable. Also, changing evaluation rules makes it more complex to reason about. Let's leave it at `#!/bin/sh`, `set -e`, and `find`.

yegortimoshenko:
Sorry for force-pushing on top of your work, for reference, it's in fc032be7355ef8d1226fa77f0ef3a0c153e36100. Again, I think it makes the script more complex and I don't see any point in making it less verbose.

zimbatm:
@yegortimoshenko it's fine to disagree but please stop merging things into master while we are having a discussion. Anyways, follow-up in https://github.com/NixOS/nixos-hardware/pull/44

yegortimoshenko:
@zimbatm I've only reverted your pull request to a previous state. The point of this pull request seemingly was to make Travis CI work and it works, so I've merged it.

PR #44, opened 2017-12-25
improve test scripts

zimbatm:
* validated with shellcheck
* allow to evaluate just one for convenience
* reduce verbosity and improve debugging output

### DIFF
--- a/tests/eval-one.sh
+++ b/tests/eval-one.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+set -u
+here=$(cd "$(dirname "$0")" && pwd)
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 <profile>"
+  exit 1
+fi
+
+profile=$1
+logfile=$(mktemp)
+
+# shellcheck disable=SC2039
+echo -n "evaluating $profile: "
+
+nix-build \
+  '<nixpkgs/nixos>' \
+  -I "nixos-config=$here/eval-test.nix" \
+  -I "nixos-hardware-profile=$profile" \
+  --dry-run \
+  --no-out-link \
+  --show-trace \
+  -A system > "$logfile" 2>&1
+ret=$?
+
+if [ "$ret" -gt 0 ]; then
+  echo ERROR
+  echo ===================================================================== >&2
+  cat "$logfile" >&2
+else
+  echo OK
+fi
+
+rm "$logfile"
+exit "$ret"
--- a/tests/eval-test.sh
+++ b/tests/eval-test.sh
@@ -1,15 +1,11 @@
 #!/bin/sh
+set -efu
+cd "$(dirname "$0")/.." || exit 1
 
-set -e
+echo "### Evaluating all profiles ###"
+echo
 
-cd "$(dirname "$0")/.."
-
+# shellcheck disable=SC2044
 for profile in $(find . -name default.nix); do
-  echo evaluating $profile >&2
-
-  nix-build '<nixpkgs/nixos>' \
-    -I nixos-config=tests/eval-test.nix \
-    -I nixos-hardware-profile=$profile \
-    -A system \
-    --dry-run
+  ./tests/eval-one.sh "$profile"
 done
--- a/tests/eval-test.sh
+++ b/tests/eval-test.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 set -efu
-cd "$(dirname "$0")/.." || exit 1
+cd "$(dirname "$0")/.."
 
 echo "### Evaluating all profiles ###"
 echo


tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
yegortimoshenko: Evaluating just one is not a proper test, because profiles can be inherited. It can be done via `nixos-rebuild -I nixos-config=tests/eval-test.nix nixos-hardware-config=path/to/default.nix dry-build`. I think it's a bad idea to allow to eval just one via a script.

tests/eval-test.sh
```diff
@@ -1,15 +1,11 @@
 #!/bin/sh
+set -efu
+cd "$(dirname "$0")/.." || exit 1
```
yegortimoshenko: This can't really fail.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+  -I "nixos-config=$here/eval-test.nix" \
+  -I "nixos-hardware-profile=$profile" \
+  --dry-run \
+  --no-out-link \
+  --show-trace \
+  -A system > "$logfile" 2>&1
+ret=$?
+
+if [ "$ret" -gt 0 ]; then
+  echo ERROR
```
yegortimoshenko: `nix-build` already prints error, we don't need to print that twice.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# shellcheck disable=SC2039
+echo -n "evaluating $profile: "
+
+nix-build \
+  '<nixpkgs/nixos>' \
+  -I "nixos-config=$here/eval-test.nix" \
+  -I "nixos-hardware-profile=$profile" \
+  --dry-run \
+  --no-out-link \
```
yegortimoshenko: `--no-out-link` doesn't do anything when `--dry-run` is present.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+profile=$1
+logfile=$(mktemp)
+
+# shellcheck disable=SC2039
+echo -n "evaluating $profile: "
+
+nix-build \
+  '<nixpkgs/nixos>' \
+  -I "nixos-config=$here/eval-test.nix" \
+  -I "nixos-hardware-profile=$profile" \
```
yegortimoshenko: We shouldn't allow profiles that have spaces in names.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+  --dry-run \
+  --no-out-link \
+  --show-trace \
+  -A system > "$logfile" 2>&1
```
yegortimoshenko: I don't see the point of having a log file instead of stderr, and I think it hides a lot of potentially useful information. Also, the idea of stderr is exactly to have an error log.

tests/eval-test.sh
```diff
@@ -1,15 +1,11 @@
...
 
-set -e
-
-cd "$(dirname "$0")/.."
+echo "### Evaluating all profiles ###"
```
yegortimoshenko: This is redundant.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 <profile>"
+  exit 1
+fi
+
+profile=$1
+logfile=$(mktemp)
+
+# shellcheck disable=SC2039
```
yegortimoshenko: What's the point of using a tool when it requires magic comments to do its job? I think `shellcheck` is very overreaching, and we shouldn't use it.

tests/eval-one.sh
```diff
@@ -0,0 +1,35 @@
...
+echo -n "evaluating $profile: "
+
+nix-build \
+  '<nixpkgs/nixos>' \
+  -I "nixos-config=$here/eval-test.nix" \
+  -I "nixos-hardware-profile=$profile" \
+  --dry-run \
+  --no-out-link \
+  --show-trace \
```
yegortimoshenko: `--show-trace` might be useful and should make it into `master`.

tests/eval-test.sh
```diff
@@ -1,15 +1,11 @@
...
+echo
 
+# shellcheck disable=SC2044
```
yegortimoshenko: Not sure why you ignore this check: https://github.com/koalaman/shellcheck/wiki/SC2044

In this case, it could be rewritten as `find . -name default.nix -exec tests/eval-one.sh {}`. But again, I think we shouldn't accept files with weird names into this repo, and current shell script makes sure that we won't.

tests/eval-one.sh
Mic92: No, in the error case the log file contains both stderr and stdin.

tests/eval-test.sh
```diff
@@ -1,15 +1,11 @@
 #!/bin/sh
+set -efu
```
Mic92: `set -o pipefail` also comes handy.

yegortimoshenko: It is a bashism.

tests/eval-test.sh
Mic92: `exit 1` should not be necessary, since `set -e` is used.

tests/eval-one.sh
yegortimoshenko: It doesn't matter, I think `nix-build` only ever writes to `stderr`. What matters is that CI builds may stutter, and with this approach, we would never know why.

tests/eval-one.sh
Mic92: Travis will evaluate all of them in the end, or not?

yegortimoshenko: Travis evaluates every profile on each commit/PR.

tests/eval-one.sh
yegortimoshenko: Also `2>&1` is a bashism.

tests/eval-one.sh
zimbatm: It doesn't hurt to escape the argument in either case

tests/eval-one.sh
zimbatm: It catches a lot of common mistakes and educates the users about shell corner-cases.

tests/eval-one.sh
zimbatm: The idea here is that I can test my profile without having to wait on all the others to evaluate. Travis is testing all the profiles anyways.

tests/eval-one.sh
zimbatm: @yegortimoshenko With this approach it's easy to find the failing profile and get to the point instead of going through pages of repeating out links. They might be interesting but not very useful in practice.

tests/eval-test.sh
zimbatm: There is no solution that doesn't depend on word splitting, without depending on bash 4.  `find -exec` doesn't early-return nor propagate the exit status. I don't think that profiles should contain special characters in any case.

tests/eval-one.sh
zimbatm: Look at the Travis CI output logs and you will understand.

yegortimoshenko: I don't understand: https://travis-ci.org/NixOS/nixos-hardware/builds/321451543#L5488

tests/eval-one.sh
yegortimoshenko: I don't think we need that for fifteen lines of shell script.

tests/eval-one.sh
yegortimoshenko: You shouldn't assume that your profile isn't inherited by any other profiles. Profile might be inherited by someone from another directory, e.g. in case of brand renaming (e.g. IBM -> Lenovo).

tests/eval-one.sh
yegortimoshenko: It does. We should enforce that there is no white spaces in names.

tests/eval-one.sh
yegortimoshenko: Right, you only need logs if something goes wrong. Your way of doing things makes certain errors opaque and non-debuggable.

tests/eval-one.sh
Mic92: How is this relevant? Travis will catch this eventually.

yegortimoshenko: It's relevant in that it's not a test and thus should not be exposed. It's possible to see someone with merge rights using `eval-one` test and then pushing to `master`.

yegortimoshenko:
Resolved in #66 with a fundamentally different approach.

yegortimoshenko:
@zimbatm I shouldn't have behaved so aggressively over this, I'm sorry.

PR #46, opened 2017-12-26
x230: init

makefu:
This PR introduces X230 to the nixos-hardware repo.

For Battery Management and power control `acpi_call` is used instead of `tp-smapi`. Additionally,  `va` and `vdpau`acceleration is enabled.
`tpm-rng` provides an extra entropy source for rngd (which is enabled by default).

### DIFF
--- a/lenovo/thinkpad/x230/default.nix
+++ b/lenovo/thinkpad/x230/default.nix
@@ -1,0 +1,24 @@
+{ config, lib, pkgs, ... }:
+
+with lib; {
+  imports = [
+    ../.
+    ../../../common/cpu/intel
+  ];
+
+  boot = {
+    kernelModules = [
+      "acpi_call"
+      "tpm-rng"
+    ];
+    extraModulePackages = with config.boot.kernelPackages; [
+      acpi_call
+    ];
+  };
+
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiIntel
+    vaapiVdpau
+    libvdpau-va-gl
+  ];
+}
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,2 @@
 lenovo/thinkpad/x250 @Mic92
+lenovo/thinkpad/x230 @makefu
--- a/README.org
+++ b/README.org
@@ -32,6 +32,7 @@ should look like:
 | Lenovo ThinkPad T460s     | ~<nixos-hardware/lenovo/thinkpad/t460s>~   |
 | Lenovo ThinkPad X140e     | ~<nixos-hardware/lenovo/thinkpad/x140e>~   |
 | Lenovo ThinkPad X220      | ~<nixos-hardware/lenovo/thinkpad/x220>~    |
+| Lenovo ThinkPad X230      | ~<nixos-hardware/lenovo/thinkpad/x230>~    |
 | Lenovo ThinkPad X250      | ~<nixos-hardware/lenovo/thinkpad/x250>~    |
 | [[file:microsoft/surface-pro/3][Microsoft Surface Pro 3]]   | ~<nixos-hardware/microsoft/surface-pro/3>~ |
 | [[file:raspberry-pi/2][Raspberry Pi 2]]            | ~<nixos-hardware/raspberry-pi/2>~          |


yegortimoshenko:
@makefu I've only read your POSIX sh array implementation yesterday!

And I am going to use this config as well :-)

lenovo/thinkpad/x230/default.nix
```diff
@@ -0,0 +1,26 @@
+{ config, lib, pkgs, ... }:
+
+with lib; {
+  imports = [
+    ../.
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop/hdd
```
yegortimoshenko: My ThinkPad X230 has a 128GB SSD. I inherit HDD/SSD profiles only when hard drive is soldered in, or proprietary, or otherwise it can be determined that there can only be this particular kind of storage.

You can import this profile in `/etc/nixos/configuration.nix`, though.

lenovo/thinkpad/x230/default.nix
```diff
@@ -0,0 +1,26 @@
...
+  ];
+
+  boot = {
+    kernelModules = [
+      "kvm-intel"
```
yegortimoshenko: Nitpick: `kvm-intel`  is recognized by `nixos-hardware-config`, so it's better to not duplicate it here.

lenovo/thinkpad/x230/default.nix
```diff
@@ -0,0 +1,26 @@
...
+      "acpi_call"
+      "tpm-rng"
```
yegortimoshenko: `tpm-rng` is certainly a good idea, but our `rngd` doesn't use it: https://github.com/NixOS/nixpkgs/commit/64311899db7c455c5f07f7ff30c5a52c50b8fbae

Perhaps we should enable `services.tcsd`?

lenovo/thinkpad/x230/default.nix
```diff
@@ -0,0 +1,26 @@
...
+    ];
+    extraModulePackages = with config.boot.kernelPackages; [
+      acpi_call
```
yegortimoshenko: Have you seen `thinkpad_acpi` module? It might be better for ThinkPads.

http://www.thinkwiki.org/wiki/Thinkpad-acpi
https://github.com/torvalds/linux/blob/0192f17529fa3f8d78ca0181a2b2aaa7cbb0784d/drivers/platform/x86/Kconfig#L399

lenovo/thinkpad/x230/default.nix
```diff
@@ -0,0 +1,26 @@
...
+    ];
+  };
+
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiIntel
+    vaapiVdpau
+    libvdpau-va-gl
```
yegortimoshenko: I always assumed this is on with `services.xserver.videoDrivers = [ "intel" ];`, but seems it's not :-( Do you think it's safe to enable it for other computers with modern Intel integrated graphics?

lenovo/thinkpad/x230/default.nix
yegortimoshenko: Also see https://github.com/NixOS/nixpkgs/issues/33033.

makefu: actually i've copy-pasted this from the x220 profile but i was unsure myself if i really need this in the profile. i remove the import

lenovo/thinkpad/x230/default.nix
makefu: actually it does. i tried running rngd -vvvv and the extra entropy source shows up only when tpm-rng is loaded

lenovo/thinkpad/x230/default.nix
makefu: but this would assume that everybody starts with nixos-hardware-config, right? this config assumed "clean slate" but i can remove this part and just put it into my customized config

lenovo/thinkpad/x230/default.nix
makefu: i had some issues with thinkpad_acpi, especially with tlp  but i am willing to give it another whirl.

lenovo/thinkpad/x230/default.nix
yegortimoshenko: Interesting!

makefu: regarding tcsd, i had an extra import prepared ( `common/tpm` ) which would enable tcsd but i was unsure if there is actually any other way to use the tpm. i personally use the tcsd

lenovo/thinkpad/x230/default.nix
yegortimoshenko: I've left it there for reverse compatibility, I'll start removing things like these soon, but I'll have to ping everyone who has ever contributed to X220 profile, because it's a breaking change. There are also a few places that explicitly specify boot loader to use, which I'm also going to drop.

lenovo/thinkpad/x230/default.nix
makefu: I also had to double-check it. Enabling it greatly increases battery life time when watching videos.
Regarding other computers, I have no computers to test unfortunately.

For checking if vdpau and va is enabled you can run `nix-shell -p libva.bin vdpauinfo --run "vainfo; vdpauinfo"`

lenovo/thinkpad/x230/default.nix
yegortimoshenko: Indeed, assumption is that users run `nixos-generate-config` because the long-term plan is to move to "dynamic dispatch" of sorts and to generate all of this in `nixos-generate-config` based on discovered hardware. Explicitly specifying model has its limitations: some hardware might be defunct, something may have been customized (e.g. SSD, or fingerprint sensor), or maybe our assumptions are incorrect. 

It's just that `nixos-generate-config` is not advanced enough to do that yet.

lenovo/thinkpad/x230/default.nix
yegortimoshenko: If you have issues, then `acpi_call` is better, at least for now. I'll test both `acpi_call` and `thinkpad_acpi` later (probably right after New Year) and I'll ping you with my results.

lenovo/thinkpad/x230/default.nix
yegortimoshenko: This is a very good find, and possibly all nixos-hardware users now will have better battery life. Thank you a lot!

I'm going to test this on a few machines that I have that are not from Lenovo...

lenovo/thinkpad/x230/default.nix
yegortimoshenko: Seemingly, the correct way to use built-in TPM is to enable `tcsd`. I think it's a good idea to enable tcsd by default for all systems with TPM, but for now I think there is no need for a new `common/` profile, we'd have to check how it works on systems other than ThinkPads (e.g. Supermicro boards).

lenovo/thinkpad/x230/default.nix
makefu: update:
`tlp stat` without `acpi_call` loaded:
```
...
+++ ThinkPad Battery Features
tp-smapi   = inactive (unsupported hardware)
tpacpi-bat = inactive (kernel module 'acpi_call' not installed)

+++ Battery Status
/sys/class/power_supply/BAT0/manufacturer                   = LGC
/sys/class/power_supply/BAT0/model_name                     = 45N1029
/sys/class/power_supply/BAT0/cycle_count                    = (not supported)
/sys/class/power_supply/BAT0/energy_full_design             =  93240 h
/sys/class/power_supply/BAT0/energy_full                    =  69500 h
/sys/class/power_supply/BAT0/energy_now                     =  62310 h
/sys/class/power_supply/BAT0/power_now                      =  13781 [mW]
/sys/class/power_supply/BAT0/status                         = Discharging

Charge                                                      =   89.7 [%]
Capacity                                                    =   74.5 [%]

+++ Suggestions
* Install acpi-call kernel module for ThinkPad battery features
```

`thinkpad_acpi` gets loaded automatically by other modules without explicitly adding it to the kernelModules.

lenovo/thinkpad/x230/default.nix
makefu: i am unsure but it seems like the RNG of the tpm will only be used if `tpm-rng` is enabled even with tcsd enabled.
I will recheck however, the train to Lepizig will take another 3 hours anyway ;)

`rngd -vvv` with `tpm-rng` loaded:
```
Available entropy sources:
	Intel/AMD hardware rng
	DRNG
```

yegortimoshenko: I also think so, it seems both `tpm-rng` and `tcsd` is required to make use of this hardware. My point was that if just `tpm-rng` is loaded, we still need a daemon to read from TPM, because NixOS `services.rngd` module explicitly prohibits reading from TPM.

makefu: Loading `tpm-rng` seems to be one way to have both `tcsd` and `rngd` get along together (according to some guy on the internet at least https://jpalczewski.github.io/en/posts/tpm-and-rngd/ )

makefu:
as discussed i've removed the disk import and the explicit `kvm-intel` in kernelModules

yegortimoshenko:
Thank you!

PR #54, opened 2018-04-06
Add builtins.fetchGit to README.org

yegortimoshenko:


yegortimoshenko:
No problem, I've turned this into a pull request.

Inline backtick equivalent is `~~`, to make a code block, one could use `#+BEGIN_SRC` and `#+END_SRC`.

yegortimoshenko:
(As I've opened this pull request, I'm not sure you are subscribed to it, so just in case, cc @Mic92)

Mic92:
Rollbacks do not work with `fetchgit`?

Mic92:
The advantage of `fetchgit` is that one does not have to manually add channels.

Mic92:
Also proposed here: https://github.com/NixOS/rfcs/pull/22/files#diff-3a4972879b71cb7724e90e002641cf0cR130

yegortimoshenko:
There is nothing comparable to channel rollback with `builtins.fetchGit`, `nixos-rebuild` rollback still works. But it's important to have channel rollback too, if someone updated `nixos-hardware` and that resulted in a breakage, but they still want to use newer `nixos` channel.

I'm concerned that this eval time fetches on every rebuild, while `nixos-rebuild --upgrade switch` explicitly updates channels. I think that we would break user expectations by recommending that, and cause slow eval. It also couples location of the overlay (in that case, link) with its interface, making testing new profiles/patches that are not present on master branch very inconvenient, while with current approach one could just point NIX_PATH to it, e.g. `sudo nixos-rebuild -I nixos-hardware=https://github.com/Mic92/nixos-hardware/archive/patch-1.tar.gz switch`

Mic92:
I will still use `fetchgit` because I don't want manually setup channel every time,

PR #55, opened 2018-04-19
apple: disable services.mbpfan

yegortimoshenko:
Causes overheating on MacBook Air 4,x and 6,x.

### DIFF
--- a/apple/default.nix
+++ b/apple/default.nix
@@ -7,6 +7,4 @@
 
   hardware.facetimehd.enable = lib.mkDefault
     (config.nixpkgs.config.allowUnfree or false);
-
-  services.mbpfan.enable = lib.mkDefault true;
 }


Ericson2314:
It definitely helped with several 2015 MacBook Pros I had for work (forget if they are 10,x or 11,x) so please make it conditional.

yegortimoshenko:
Good to know. Actually I was wrong about 6,x, overheating only ever occurs on 4,x. I think this might be upstream issue, or maybe some sensor on one of my Airs is malfunctioning. Sorry for overreacting.

yegortimoshenko:
Hm, actually no, MacBook Air 6,x also overheats with `services.mbpfan` (over 100 celsius). I don't plan to merge this in current state but would like to keep it as a reminder to resolve the issue.

Mic92:
Merge or not to merge?

sevenfourk:
for me fan crazy unexpectedly starts to run at 5-6k rpms, I almost immediately jump out of the chair.

macbook-pro-12,1.

will disable it for now, to see if machine will keep cool without it.

Mic92:
@Ericson2314 should we maybe mention it in the README instead of making it conditional?

Mic92:
Seems stalled.

PR #57, opened 2018-05-15
intel: videoDrivers: intel -> modesetting

yegortimoshenko:
This seems to be the current best practice.

https://www.phoronix.com/scan.php?page=article&item=intel-modesetting-2017&num=1
https://jlk.fjfi.cvut.cz/arch/manpages/man/modesetting.4

### DIFF
--- a/common/cpu/intel/default.nix
+++ b/common/cpu/intel/default.nix
@@ -2,8 +2,7 @@
 
 {
   hardware.cpu.intel.updateMicrocode = lib.mkDefault true;
-  services.xserver.videoDrivers = [ "modesetting" ];
-
+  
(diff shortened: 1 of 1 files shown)


common/cpu/intel/default.nix
```diff
@@ -2,7 +2,7 @@
 
 {
   hardware.cpu.intel.updateMicrocode = lib.mkDefault true;
-  services.xserver.videoDrivers = [ "intel" ];
+  services.xserver.videoDrivers = [ "modesetting" ];
```
Mic92: This is the default. You can delete this line as well.

yorickvP:
Anything blocking this?

PR #58, opened 2018-05-15
add XPS 13 (9360)

yorickvP:


dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
...
+    ../../../common/pc/laptop
+  ];
+  boot.kernelModules = ["kvm-intel"]; # should this be in common/cpu/intel?
+  boot = {
+    loader = {
+      systemd-boot.enable = lib.mkDefault true;
+      efi.canTouchEfiVariables = lib.mkDefault true;
+    };
+
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
```
yegortimoshenko: Does it only work with the latest kernel? (Current LTS kernel is 4.14.x)

dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
...
+  };
+  
+  # intel huc, guc. qca6174 (old?)
+  hardware.enableRedistributableFirmware = true;
+
+  # 4k screen, use bigger console font
+  i18n.consoleFont = "latarcyrheb-sun32";
+
+  # touchpad
+  services.xserver.libinput.enable = lib.mkDefault true;
```
yegortimoshenko: See https://github.com/NixOS/nixos-hardware/blob/master/common/pc/default.nix

dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  networking.wireless.enable = lib.mkDefault true;
```
yegortimoshenko: This conflicts with iwd and NetworkManager:

https://github.com/NixOS/nixpkgs/blob/release-18.03/nixos/modules/services/networking/iwd.nix#L12
https://github.com/NixOS/nixpkgs/blob/release-18.03/nixos/modules/services/networking/networkmanager.nix#L265

dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
...
+  hardware.bluetooth.enable = lib.mkDefault true;
```
yegortimoshenko: This module should really be called `services.bluez` :-)

Bluetooth works without that, say, in GNOME.

dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
...
+      systemd-boot.enable = lib.mkDefault true;
+      efi.canTouchEfiVariables = lib.mkDefault true;
+    };
+
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
+    initrd.availableKernelModules = [ "xhci_pci" "nvme" "usb_storage" "sd_mod" "rtsx_pci_sdmmc" ];
+    # touchpad goes over i2c
+    blacklistedKernelModules = [ "psmouse" ];
+
+    kernelParams = [ "i915.enable_fbc=1" "i915.enable_psr=2" ];
```
yegortimoshenko: This should probably be moved to `common/cpu/intel/kaby-lake/default.nix`.

dell/xps/13-9360/default.nix
```diff
@@ -0,0 +1,42 @@
+{ lib, pkgs, ... }:
+let
+  firmware_qca6174 = pkgs.callPackage ./firmware_qca6174.nix {};
+in
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+  ];
+  boot.kernelModules = ["kvm-intel"]; # should this be in common/cpu/intel?
```
yegortimoshenko: The policy up until now was to not include options that are generated by `nixos-generate-config`.

dell/xps/13-9360/default.nix
yorickvP: everything higher than 4.11 should work without kernel patches, but not all of them work as well as others. 4.14.x should be fine, but I haven't tested that in forever.
Is https://patchwork.kernel.org/patch/9676073/ in 4.14?

dell/xps/13-9360/qca6174-firmware.nix
```diff
@@ -0,0 +1,26 @@
...
+
+stdenv.mkDerivation rec {
+  name = "${target}-firmware-${version}";
+  version = "${branch}-00042";
+
+  branch = "4.4.1.c1";
+  target = "QCA6174";
+
+  src = fetchurl {
+    url = "https://github.com/kvalo/ath10k-firmware/raw/master/${target}/hw3.0/${branch}/firmware-6.bin_RM.${version}-QCARMSWP-1";
```
Mic92: This hash will break if `https://github.com/kvalo/ath10k-firmware` is updated. It is better to reference a concrete commit.

Mic92: Also we can add this to nixpkgs directly.

yegortimoshenko: Fixed in master.

PR #60, opened 2018-05-31
Add basic support for 6th-gen X1

azazel75:
Hello, 

this is a PR to support X1, as with the same name there are multiple "generations" and at least the last generation comes with tree display flavors, I chose to make nested directories to differentiate between those.


azazel75:
@yegortimoshenko I think I'm done

Mic92:
Still interested in reviewing this? @yegortimoshenko

lenovo/thinkpad/x1/6th-gen/default.nix
```diff
@@ -0,0 +1,73 @@
...
+
+  # Enable S3 suspend state: you have to manually follow the
+  # instructions shown here: https://delta-xi.net/#056 in order to
+  # produce the ACPI patched table. Put the CPIO archive in /boot and
+  # then enable the following lines
+  # boot.kernelParams = [
+  #   "mem_sleep_default=deep"
+  # ];
+  # boot.initrd.prepend = [
+  #   "/boot/acpi_override"
```
garbas: this should probably be ``"${./path/to/acpi_override}"`` since nix will prepend this to initird

@azazel75 did you manage to get suspend working? 

once i apply this patch and uncomment above 2 lines i still don't see ``S3`` supported in acpi.
```
dmesg| grep -i "acpi: (supports"
[    0.188263] ACPI: (supports S0 S4 S5)
```

I also updated the script to generate patch to use `sed` since applying patch is broken with BIOS >= 1.24

azazel75: Hi @garbas, 
I can patch the line you commented, it's the same thing... in your case the absolute path is calculated by nix, I suppose!? (I'm a newbie)

About the patch, I had also to manually patch the "disassembled" code because the `patch` binary falied to apply some (final) hunks... I don't know which version my BIOS is at but I spawned up  Windows just three weeks ago and let it apply an upgrade to the BIOS using the installed Lenovo software. I was worried that it may break the ACPI S3 stuff but it still works after that.  Are you sure that the kernel patches the table on startup? My early kernel logs tell me that:
```
$ dmesg  | grep DSDT
[    0.000000] ACPI: DSDT ACPI table found in initrd [kernel/firmware/acpi/dsdt.aml][0x2338b]
[    0.000000] ACPI: Table Upgrade: override [DSDT-LENOVO-SKL     ]
[    0.000000] ACPI: DSDT 0x000000004FFC0000 Physical table override, new table: 0x000000004DF1E000
[    0.000000] ACPI: DSDT 0x000000004DF1E000 02338B (v02 LENOVO SKL      00000001 INTL 20180313)
[...]
$ dmesg | grep -i "acpi: (supports"
[    0.132924] ACPI: (supports S0 S3 S4 S5)
```

azazel75: Hi @garbas, I just checked my bios updates booting in Windows. It has upgraded my BIOS to 1.25 and S3 is still working

garbas: @azazel75 i forgot to send you the link to the changes of the script i did -> https://github.com/fiji-flo/x1carbon2018s3/pull/1

Looks like suspend works for you.

regarding the `"/boot/acpi_override"`, during the build you can not access `/boot` but with `"${/boot/acpi_override}"` nix will prepend it to the file to `/nix/store` and then it can append it to initrd (https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/kernel/make-initrd.sh#L40).

being said that i'm still missing something since i dont see any override for the DSDT. need to fi

azazel75: I must say that I've hadn't any problem evaluating my configuration with "/boot/acpi_override", why do you say that you can not access `/boot`?
Yes I saw the line in  `make-initrd.sh` and with `/boot/acpi_override` as $PREP it should work. In fact, mine works flawlessly

garbas: @azazel75 hmm, that is weird. it complained not to have permissions to open that file for me.

Here is my config: https://github.com/garbas/dotfiles/blob/master/nixos/grayworm.nix.
I'm running on nixos-unstable channel.

garbas:
@azazel75 did you see #62 ? is there something to be included from that pull request? maybe at least as a section that is commented.

also I loved that you provided links where you got suggestions for configuration from, it makes it much easier to review/use.

azazel75:
>>>>> "Rok" == Rok Garbas <notifications@github.com> writes:

    Rok> @azazel75 did you see #62 ? is there something to be included
    Rok> from that pull request? maybe at least as a section that is
    Rok> commented.

Uh, I didn't. I'm a bit astonished by the duplicate PR....anyway:

    >   boot.extraModprobeConfig = "options thinkpad_acpi experimental=1 fan_control=1";


This probably isn't so useful without activating "thinkfan" daemon, but
I'm not using it and the fan works on high load as expected. Maybe it
works better? I dunno http://www.thinkwiki.org/wiki/How_to_control_fan_speed

For the `experimental=1` parameter it lacks some reference about what it
will do IMHO.

    >   # enable thermal trottle
    >   services.thermald.enable = true;

Same here, is it really useful/needed? I tend to run as few daemons as
possible in order to save battery. Again, it lacks references
I wonder how much functionality of thermald/thinkfan is really needed in
latest kernel and if/how much of it is covered also by tlp

    >   # enable sound
    >   sound.enable = true;
    >   hardware.pulseaudio.enable = true;

This seems opinionated

    >   # handle lid close hibernate, suspend, or ignore
    >   systemd.extraConfig = "";
    >   services.logind.extraConfig = ''
    >     HandleLidSwitch=hibernate
    >     LidSwitchIgnoreInhibited=yes
    >   '';

this seems useless if S3 is working (on this topic: I'm using unstable
too and If I remember correctly I had to enable read for others on the
`/boot/acpi_override` file)

    >   # Thinkpad power services with some sample values
    >   services.tlp.enable = true;
    >   services.tlp.extraConfig = ''
    >       DEVICES_TO_DISABLE_ON_STARTUP="bluetooth"
    >       START_CHARGE_THRESH_BAT0=75
    >       STOP_CHARGE_THRESH_BAT0=90
    >       START_CHARGE_THRESH_BAT1=75
    >       STOP_CHARGE_THRESH_BAT1=90
    >       CPU_SCALING_GOVERNOR_ON_BAT=powersave
    >       ENERGY_PERF_POLICY_ON_BAT=powersave
    >     '';

- disabling bluetooth seems opinionated
- BAT1 refers to non existent second battery


    Rok> also I loved that you provided links where you got
    Rok> suggestions for configuration from, it makes it much easier to
    Rok> review/use.



-- 
Alberto Berti -  Information Technology Consultant

"gutta cavat lapidem"

Mic92:
I suppose this is good to merge?

azazel75:
The `powersave` setup in tlp's `extraConfig` is actually useful, I've enabled it

Mic92:
I suppose it should be also added to the table in the README.

azazel75:
Thanks @Mic92: for the README see #69

PR #62, opened 2018-06-29
thinkpad x1 gen6

denten:
First stab at Thinkpad x1 gen6 profile.


lenovo/thinkpad/default.nix
```diff
@@ -1,10 +0,0 @@
-{ lib, pkgs, ... }:
```
michaelpj: This breaks every other configuration in this directory, since they all import this.

lenovo/thinkpad/x1-gen6/default.nix
```diff
@@ -0,0 +1,41 @@
+# pulled out all the x1 specific things
+# for nixos-hardware
+
+{ config, pkgs, ... }:
+
+{
+
+  boot.kernelModules = [ "acpi" "thinkpad-acpi" "acpi-call" ];
+  boot.extraModulePackages = [ config.boot.kernelPackages.acpi_call ];
+  boot.extraModprobeConfig = "options thinkpad_acpi experimental=1 fan_control=1";
```
michaelpj: Maybe improve the `acpi_call` module in the parent directory and import that?

lenovo/thinkpad/x1-gen6/default.nix
```diff
@@ -0,0 +1,41 @@
...
+
+  environment.systemPackages = with pkgs; [
```
michaelpj: If you import the default module this is implied by that (in `common/pc/laptop`)

lenovo/thinkpad/x1-gen6/default.nix
```diff
@@ -0,0 +1,41 @@
...
+
+  # enable thermal trottle
+  services.thermald.enable = true;
+
+  # enable sound
+  sound.enable = true;
+  hardware.pulseaudio.enable = true;
+
+  # handle lid close hibernate, suspend, or ignore
+  systemd.extraConfig = "";
```
michaelpj: Did you mean to include this?

lenovo/thinkpad/x1-gen6/default.nix
```diff
@@ -0,0 +1,41 @@
...
+  services.logind.extraConfig = ''
+    HandleLidSwitch=hibernate
+    LidSwitchIgnoreInhibited=yes
+  '';
+
+  # Thinkpad power services with some sample values
+  services.tlp.enable = true;
+  services.tlp.extraConfig = ''
```
michaelpj: I wonder if it makes sense to put these in the default thinkpad module, since presumably they apply generally?

yegortimoshenko: This is already enabled in `common/pc/laptop` profile.

lenovo/thinkpad/x1-gen6/default.nix
```diff
@@ -0,0 +1,41 @@
...
+  environment.systemPackages = with pkgs; [
+    tlp
+  ];
+
+  # enable thermal trottle
+  services.thermald.enable = true;
+
+  # enable sound
+  sound.enable = true;
+  hardware.pulseaudio.enable = true;
```
michaelpj: These probably shouldn't be here, since people may want to make their own choices about sound settings. Generally any direct settings should probably have `mkDefault` so that other people's settings will override them.

lenovo/thinkpad/x1-gen6/default.nix
borisbabic: I think there are some differences between generations. AFAIK The oldest thinkpads on the list should behave differently with `CPU_SCALING_GOVERNOR_ON_BAT=powersave` since powersave behaves differently depending on the scaling driver. Using powersave with pstate in newer generations isn't like powersave in the previous generations. 

BTW. I don't think disabling bluetooth on startup should be a default. IMHO making some tweaks here by default is okay but making tradeoffs like this is better left to be opt in.

michaelpj:
I think they're both maybe a little opinionated? It would be good if we had clearer guidelines about which options you should set vs. leave to users to set. I also think it would be nice if we could share as much as possible between the thinkpad configurations otherwise there's a ton of copy pasting.

garbas:
@michaelpj i found things not to be to opinionated in #60. i think @yegortimoshenko has an idea about guidelines, but it would be nicer to see them written down.

sharing between thinkpad, with current setup this is definitely possible. i personally see this as an optimization step. once we notice that things duplicate we move the to "lower" placing default.nix.

garbas:
closing this since this PR was reviewed in #60 (which just got merged) and integrated non-opinionated configuration from this PR.

PR #63, opened 2018-07-13
Add a note about rcu_sched problems

boothead:
How I fixed my XPS15 after it being dead for two months


yegortimoshenko:
Thank you!

Mic92:
I like how we can express workarounds in a crisp, brief manner instead of burying the solution in a forum on page 3.

roberth:
Hold on, this is the XPS 15 9550 page, not the XPS 15 9560, right? Although similar, they are different models.

boothead:
Right you are @roberth - Also I notice I've screwed up the rendering (I thought I was writing markdown)

PR #65, opened 2018-07-17
Add readme section on using fetchgit

michaelpj:
I like to use `fetchgit` to get the repository, rather than a channel, since it makes my `configuration.nix` more self-contained.

### DIFF
--- a/README.org
+++ b/README.org
@@ -2,6 +2,8 @@ NixOS profiles covering hardware quirks.
 
 ** Setup
 
+*** Using channels
+
 Add and update ~nixos-hardware~ channel:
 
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
Also see #54.

michaelpj:
Well, I think having the alternatives in the readme is fine so long as any drawbacks are documented.

I also think that while not having channel rollback is a shame, it's wrong to say that you don't have rollbacks - `nixos-rebuild` rollbacks are 90% of what I want!

Mic92:
`fetchTarball` is a bit faster on small repositories and pinning is also possible: https://github.com/nix-community/NUR#pinning

PR #67, opened 2018-07-17
Some suggestions for CONTRIBUTING

michaelpj:
Just a couple of things that I think we want people to do, but it's worth saying explicitly.


Mic92:
Merged in https://github.com/NixOS/nixos-hardware/pull/77

Issue #41, opened 2017-12-25
Link with the ArchLinux wiki


zimbatm:
The [ArchLinux wiki](https://wiki.archlinux.org/) is a wonderful source of information. It has a great deal of information on hardware support and is actively kept up2date. Let's piggy-back on it.

I can think of two things that could be useful:
* profile naming: there exists a lot of hardware out there. When in doubt with the profile name, that the name of the ArchLinux wiki page.
* profile info: link back to the ArchLinux wiki page. This will be useful to motivate various configuration decisions.

I think that we should piggy-back on it. For example if there is an ArchLinux wiki page for a specific hardware.

What do you think?

zimbatm:
Also I quite like the hardware report that they are giving. Eg: https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_Edge_E455 . It gives a good idea of what has been tested to work.

yegortimoshenko:
I agree, we should link to it in per-model README file whenever it makes sense to do so, but there are other great resources that we should also link to, [ThinkWiki](https://www.thinkwiki.org/wiki/ThinkWiki) for ThinkPads, [EveryMac](https://everymac.com/) for Macs...

Also, the point of this thing is so that people don't have to configure anything themselves, so we shouldn't rely on the user to check the ArchWiki or any other resource. This should be mostly for maintenance.

Mic92:
I think we already do this where appropriate. So let's close it.

PR #68, opened 2018-07-23
apple/macbookpro/11-5: init

rehno-lindeque:
Similar to [macbookpro/10-1](https://github.com/NixOS/nixos-hardware/tree/585a13d8b929aba086bed5f1fa3317afd30c39b3/apple/macbook-pro/10-1) and [macbookpro/12-1](https://github.com/NixOS/nixos-hardware/tree/585a13d8b929aba086bed5f1fa3317afd30c39b3/apple/macbook-pro/12-1).

Includes (minimal) external wiki references as suggested by #41, but this commit can be dropped if it is considered out of scope.


rehno-lindeque:
Also added a fix for a spurious wakeup signal that I judged to be important enough to include in this initial PR.

apple/macbook-pro/11-5/default.nix
```diff
@@ -0,0 +1,30 @@
...
+in
+
+{
+  imports = [
+    ../.
+    ../../../common/pc/laptop/ssd
+    <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix>
+  ];
+
+  boot.loader.systemd-boot.enable = true;
```
Mic92: I think mkDefault should be used for boolean options: https://github.com/NixOS/nixos-hardware/pull/67/files#diff-0660552a9aaaac3267076efad7bf5c1bR9

apple/macbook-pro/11-5/default.nix
```diff
@@ -0,0 +1,30 @@
...
+  boot.loader.efi.canTouchEfiVariables = true;
```
Mic92: This should be only done for the first installation. Otherwise a new boot entry is created for every nixos-rebuild switch. This can fill up the nvram, which is also limited in terms of write-cycles.

apple/macbook-pro/11-5/default.nix
rehno-lindeque: Fair enough, seems reasonable to make this optional.

apple/macbook-pro/11-5/default.nix
rehno-lindeque: Interesting, thank you. How about if I put these two lines as instructions in the README.md file instead?

I suppose [macbook-pro/10-1](https://github.com/NixOS/nixos-hardware/blob/master/apple/macbook-pro/10-1/default.nix#L11) could also do with a PR to fix.

Would be nice to update the [docs](https://github.com/NixOS/nixpkgs/blob/80b5f5227e4a801bc9da70ba68d59aad54618e7c/nixos/modules/system/boot/loader/efi.nix#L11) too.

rehno-lindeque: Are these instructions better suited? https://nixos.wiki/wiki/Bootloader#Keeping_kernels.2Finitrd_on_the_main_partition

rehno-lindeque: Ok, sorry for the noise: Seeing as this config is already automatically [generated by nixos-generate-config](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-generate-config.pl#L534) I suppose it really shouldn't be in this module at all. 

I'll remove the boot loader completely if there's no objections.

Mic92:
Looks good. Only merge conflicts needs to be resolved.

rehno-lindeque:
OK, that ought do it

PR #70, opened 2018-08-13
lenovo/thinkpad/x260: Init

rvolosatovs:
Added a profile for x260.
- TLP config includes a fix for BTRFS(the first line), which e.g. is not present in config for x1 6th gen. AFAIK this should not influence the performance too much, yet it should prevent possible bugs/data corruption, which BTRFS users may encounter otherwise, hence I think it's a good idea to put it there.
- While working on this I noticed that most of the other ThinkPad profiles do not include the `acpi_call`, which is required, e.g. for TLP to work. Did not include the fixes, as that is out of scope of this PR.
- Maybe it would be a good idea to extract the TLP config into a (shared) function, which would generate powersaving config given the battery count?

### DIFF
--- a/lenovo/thinkpad/x260/default.nix
+++ b/lenovo/thinkpad/x260/default.nix
@@ -7,6 +7,6 @@
 
   # https://wiki.archlinux.org/index.php/TLP#Btrfs
   services.tlp.extraConfig = ''
-    SATA_LINKPWR_ON_BAT=max_performance
+    SATA_LINKPWR_ON_BAT=med_power_with_dipm
   '';
 }
--- a/lenovo/thinkpad/x260/default.nix
+++ b/lenovo/thinkpad/x260/default.nix
@@ -1,0 +1,20 @@
+{
+  imports = [
+    ../.
+    ../acpi_call.nix
+    ../../../common/cpu/intel
+  ];
+
+  # See https://linrunner.de/en/tlp/docs/tlp-faq.html#battery https://wiki.archlinux.org/index.php/TLP#Btrfs
+  services.tlp.extraConfig = ''
+    START_CHARGE_THRESH_BAT0=75
+    STOP_CHARGE_THRESH_BAT0=80
+    START_CHARGE_THRESH_BAT1=75
+    STOP_CHARGE_THRESH_BAT1=80
+
+    SATA_LINKPWR_ON_BAT=max_performance
+
+    CPU_SCALING_GOVERNOR_ON_BAT=powersave
+    ENERGY_PERF_POLICY_ON_BAT=powersave
+  '';
+}
--- a/lenovo/thinkpad/x260/default.nix
+++ b/lenovo/thinkpad/x260/default.nix
@@ -5,16 +5,8 @@
     ../../../common/cpu/intel
   ];
 
-  # See https://linrunner.de/en/tlp/docs/tlp-faq.html#battery https://wiki.archlinux.org/index.php/TLP#Btrfs
+  # https://wiki.archlinux.org/index.php/TLP#Btrfs
   services.tlp.extraConfig = ''
-    START_CHARGE_THRESH_BAT0=75
-    STOP_CHARGE_THRESH_BAT0=80
-    START_CHARGE_THRESH_BAT1=75
-    STOP_CHARGE_THRESH_BAT1=80
-
     SATA_LINKPWR_ON_BAT=max_performance
-
-    CPU_SCALING_GOVERNOR_ON_BAT=powersave
-    ENERGY_PERF_POLICY_ON_BAT=powersave
   '';
 }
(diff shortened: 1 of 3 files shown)


lenovo/thinkpad/x260/default.nix
```diff
@@ -0,0 +1,20 @@
+{
+  imports = [
+    ../.
+    ../acpi_call.nix
+    ../../../common/cpu/intel
+  ];
+
+  # See https://linrunner.de/en/tlp/docs/tlp-faq.html#battery https://wiki.archlinux.org/index.php/TLP#Btrfs
+  services.tlp.extraConfig = ''
+    SATA_LINKPWR_ON_BAT=max_performance
```
yegortimoshenko: Good catch! We should probably apply this for all profiles, not just ThinkPad X260.

lenovo/thinkpad/x260/default.nix
```diff
@@ -0,0 +1,20 @@
...
+
+    START_CHARGE_THRESH_BAT0=75
+    STOP_CHARGE_THRESH_BAT0=80
+    START_CHARGE_THRESH_BAT1=75
+    STOP_CHARGE_THRESH_BAT1=80
```
yegortimoshenko: Is that specific to this model? Is the rationale to save up battery charge cycles?

lenovo/thinkpad/x260/default.nix
```diff
@@ -0,0 +1,20 @@
...
+
+    CPU_SCALING_GOVERNOR_ON_BAT=powersave
+    ENERGY_PERF_POLICY_ON_BAT=powersave
```
yegortimoshenko: Huh, apparently that's not a default. Maybe let's make it into one?

lenovo/thinkpad/x260/default.nix
rvolosatovs: See [here](https://linrunner.de/en/tlp/docs/tlp-faq.html#battery)
>Factory settings for ThinkPad battery thresholds are as follows: when plugged in the battery starts charging at 96%, and stops at 100%. These settings are optimized for maximum runtime, but having a battery hold a lot of power will decrease its capacity over the years. To alleviate this problem, the start/stop charge thresholds can be adjusted – at the cost of a more or less reduced battery runtime.

These values are rather conservative defaults from the TLP manual.

lenovo/thinkpad/x260/default.nix
rvolosatovs: What do you mean by this?

Mic92:
@yegortimoshenko anything that needs to be done here?

lenovo/thinkpad/x260/default.nix
```diff
@@ -0,0 +1,20 @@
...
+    START_CHARGE_THRESH_BAT0=75
+    STOP_CHARGE_THRESH_BAT0=80
+    START_CHARGE_THRESH_BAT1=75
+    STOP_CHARGE_THRESH_BAT1=80
+
+    SATA_LINKPWR_ON_BAT=max_performance
+
+    CPU_SCALING_GOVERNOR_ON_BAT=powersave
+    ENERGY_PERF_POLICY_ON_BAT=powersave
+  '';
```
yegortimoshenko: I've tested this config on i7-6500U laptop (Purism Librem 13v3) for several days, which is the same CPU generation that X260 comes with (Skylake), but supposedly somewhat more performant. Unfortunately, this config (specifically, `CPU_SCALING_GOVERNOR_ON_BAT=powersave` and `ENERGY_PERF_POLICY_ON_BAT=powersave`) causes severe performance degradation to the point where, e.g. GNOME would not run smoothly anymore.

As of charge thresholds, that seems to be opinionated as well, and I can see other people (myself included)  would like to leverage full battery capacity instead. So maybe let's stick with TLP defaults on that as well.

rvolosatovs:
@yegortimoshenko rebased and addressed your comment

lenovo/thinkpad/x260/default.nix
```diff
@@ -0,0 +1,12 @@
+{
+  imports = [
+    ../.
+    ../acpi_call.nix
+    ../../../common/cpu/intel
+  ];
+
+  # https://wiki.archlinux.org/index.php/TLP#Btrfs
+  services.tlp.extraConfig = ''
+    SATA_LINKPWR_ON_BAT=max_performance
```
Mic92: There is a discussion on this: https://wiki.archlinux.org/index.php/Talk:TLP#Btrfs
Should we make only enable this on kernel lower then 4.15?

Mic92: Was this reproducible with any other file system then BTRFS?

yegortimoshenko: @Mic92 I think we can set it to `med_power_with_dipm` on all laptops starting with 4.15 kernels, just in case.

PR #78, opened 2018-09-01
Fix unusable Lenovo Thinkpad X1 profile

azazel75:
Just a couple of commits to fix the X1 profile.  The first fixes the reference to a file that was renamed and the second  fixes acpi override file.


yegortimoshenko:
Thank you!

PR #81, opened 2018-10-19
init dell xps 9370

Moredread:
I'm not sure if TLP sets battery thresholds correctly with acpi_call on Dell notebooks, but I guess it doesn't hurt to enable it.

I haven't found out how to set those thresholds manually, the last time I tried I think the instructions I found online were for older Dell models and didn't work for the 9370 but I'm not sure anymore.


dell/xps/13-9370/default.nix
```diff
@@ -0,0 +1,16 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel/kaby-lake
+    ../../../common/pc/laptop
+    ../../../lenovo/thinkpad/acpi_call.nix
+    ../../../lenovo/thinkpad/cpu-throttling-bug.nix
```
yegortimoshenko: Hm, interesting. So neither this nor acpi_call are specific to ThinkPads?

Moredread: Correct, but I only played around with acpi_call a while ago, without any specific result. The throttling bug seems to hit most Dell and Thinkpads (and maybe other vendors) models with the latest Kabby Lake CPUs as far as I understand it.

Moredread: Oh, I think I misremembered what I used acpi_call for. I used it to disable the dedicated graphics card on my old Thinkpad, so I haven't tested it on my Dell. But from https://github.com/mkottman/acpi_call it seems to be compatible with other models as well in principle.

Ma27:
when this is a module used among several vendors, shouldn't we move it out of the `lenovo/` tree?

yegortimoshenko:
Of course we should move it into `common`.

Moredread:
I've added a commit to this PR that moves both files

Moredread:
I'm not sure why it fails on Travis, locally it builds fine and I don't find a reference to `lenovo/thinkpad/acpi_call.nix` in my branch. Maybe a cache issue?

Mic92:
It is https://github.com/NixOS/nixos-hardware/blob/master/lenovo/thinkpad/t450s/default.nix#L6
which which was added in the meantime but is not included yet in your own branch.

Moredread:
Oh sorry, I'll rebase my branch asap.

Mic92:
I fixed it in: https://github.com/NixOS/nixos-hardware/pull/82

PR #84, opened 2018-10-30
lenovo/thinkpad/x230: add coreboot sub-profile

yegortimoshenko:
ACPI call is not supported by Coreboot, also it pulls in `mei_me` module which causes warnings if laptop runs on cleaned ME image (which, while orthogonal to Coreboot, typically goes hand in hand).

### DIFF
--- a/lenovo/thinkpad/x230/coreboot/default.nix
+++ b/lenovo/thinkpad/x230/coreboot/default.nix
@@ -1,0 +1,3 @@
+{
+  imports = [ ../common.nix ];
+}
--- a/lenovo/thinkpad/x230/default.nix
+++ b/lenovo/thinkpad/x230/default.nix
@@ -1,19 +1,6 @@
-{ config, lib, pkgs, ... }:
-
 {
   imports = [
-    ../.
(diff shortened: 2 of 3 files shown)


lenovo/thinkpad/x230/default.nix
```diff
@@ -1,19 +1,6 @@
...
-
-  boot = {
-    kernelModules = [
-      "tpm-rng"
-    ];
-  };
-
-  services.xserver.deviceSection = lib.mkDefault ''
-    Option "TearFree" "true"
-  '';
```
yegortimoshenko: This is gone as it's not specific to X230 config.

lenovo/thinkpad/x230/coreboot/default.nix
```diff
@@ -0,0 +1,3 @@
+{
+  imports = [ ../common.nix ];
```
Mic92: should this variant be also added to the README?

Mic92:
Closed due lack of responsiveness. If someone else wants to add this profile, feel free to open it again.

PR #86, opened 2018-10-30
common/pc: disable ath3k if non-free firmware is not available

yegortimoshenko:
`ath3k` requires non-free firmware.

This improves boot speed if `enableRedistributableFirmware` is false.

### DIFF
--- a/common/pc/default.nix
+++ b/common/pc/default.nix
@@ -1,5 +1,9 @@
-{ lib, ... }:
+{ config, lib, ... }:
 
 {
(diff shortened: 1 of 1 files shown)


Moredread:
Isn't this better placed in NixOS itself, as it applies to the module as a whole, independent of hardware?

Moredread:
I see this is still open. I have changed my mind, I think it fits nicely here. Thumps up from me XD (if it still is the case with needing unfree firmware in newer kernels)

PR #88, opened 2018-12-27
lenovo x1 carbon gen6: update S3 suspend

bhipple:
With UEFI >= 1.30, there's an explicit option to enable S3 power management.
Once this is selected, S3 is enabled and "deep" is selected as the default
/sys/power/mem_sleep value without requiring any kernel boot parameters.

### DIFF
--- a/lenovo/thinkpad/x1/6th-gen/default.nix
+++ b/lenovo/thinkpad/x1/6th-gen/default.nix
@@ -2,7 +2,9 @@
 # standing with kernel 4.6.11 is the following wiki page:
 # https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_X1_Carbon_(Gen_6). The
 # TrackPoint and TouchPad issues there seem to have been fixed already.
-
+#
+# Enable the lower-power S3 suspend state by upgrading the BIOS to version >= 1.30,
+# then manually selecting Linux in the power management section.
 { config, pkgs, ... }:
 {
   imports = [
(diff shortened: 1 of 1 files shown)


bhipple:
CC @azazel75 @FruitieX @hoopla12 I see from https://discourse.nixos.org/t/nixos-on-x1-carbon/1627 that you're using this laptop; if you haven't done so already, you may want to upgrade your UEFI version to the latest one that supports the efficient sleep by default.

azazel75:
@bhipple , well done! I've actually updated several days ago but hadn't found the time yet to  submit a patch. +1 for me

azazel75:
Just one note, are you sure that it isn't better to reference the firmware upgrade as "BIOS version >= 1.30" instead of "UEFI"? The archlinux wiki too refers to it as "BIOS update" https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_X1_Carbon_(Gen_6)#Enabling_S3  and  you can actually check your version entering the BIOS configuration utility

lenovo/thinkpad/x1/6th-gen/default.nix
```diff
@@ -2,7 +2,9 @@
 # standing with kernel 4.6.11 is the following wiki page:
 # https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_X1_Carbon_(Gen_6). The
 # TrackPoint and TouchPad issues there seem to have been fixed already.
-
+#
+# Enable the lower-power S3 suspend state by upgrading UEFI to version >= 1.30,
```
azazel75: Maybe it's better to use BIOS instead of UEFI here?

bhipple: Yeah that's a good point; upgraded comment accordingly.

Issue #90, opened 2018-12-30
Listing derivations (nix-env -qa) does not work


zarelit:
When I try to list available packages with `nix-env -qaP` the evaluation breaks in `nixos-hardware/default.nix` and I don't know what's wrong. The following logs are taken as root but it behaves the same with an unprivileged user.

**nix-env -qaP --show-trace**
```
[root@tsundoku:/home/zarel]# nix-env -qaP --show-trace
error: while evaluating 'buildProfile' at /nix/store/rkb722vhq7xkhid0nbnl55i2r59csd77-nixos-hardware/nixos-hardware/default.nix:18:18, called from /nix/store/rkb722vhq7xkhid0nbnl55i2r59csd77-nixos-hardware/nixos-hardware/default.nix:51:20:
while evaluating the attribute 'system' at /nix/var/nix/profiles/per-user/root/channels/nixos/nixos/default.nix:33:3:
while evaluating the attribute 'config.system.build.toplevel' at /nix/var/nix/profiles/per-user/root/channels/nixos/nixos/modules/system/activation/top-level.nix:273:5:
while evaluating 'foldr' at /nix/var/nix/profiles/per-user/root/channels/nixos/lib/lists.nix:39:20, called from /nix/var/nix/profiles/per-user/root/channels/nixos/nixos/modules/system/activation/top-level.nix:137:12:
while evaluating 'fold'' at /nix/var/nix/profiles/per-user/root/channels/nixos/lib/lists.nix:42:15, called from /nix/var/nix/profiles/per-user/root/channels/nixos/lib/lists.nix:46:8:
while evaluating anonymous function at /nix/var/nix/profiles/per-user/root/channels/nixos/nixos/modules/system/activation/top-level.nix:137:52, called from /nix/var/nix/profiles/per-user/root/channels/nixos/lib/lists.nix:45:14:
while evaluating anonymous function at /nix/store/vs30cml2ma72nd6vavfzc7ljrqw3p356-nixos-18.09.1819.76aafbf4bf4/nixos/pkgs/build-support/replace-dependency.nix:20:1, called from /nix/var/nix/profiles/per-user/root/channels/nixos/nixos/modules/system/activation/top-level.nix:138:7:
[...]

PR #92, opened 2019-01-04
fix nix-env evaluation

Mic92:
nix-env picked up `default.nix`, which contained our hardware profiles.
This is not only cpu-intensive to evaluate but also fails because
it does some import-from-derivation.
By moving default.nix to release.nix it should be no longer loaded


Mic92:
fixes https://github.com/NixOS/nixos-hardware/issues/90

kalbasit:
would this break configurations loading the nixos hardware with `import <nixos-hardware>`? in other words, do I have to explicitly specify release.nix?

yegortimoshenko:
> would this break configurations loading the nixos hardware with import <nixos-hardware>? in other words, do I have to explicitly specify release.nix?

I don't think so; this has never been supported API.

PR #93, opened 2019-01-08
add raspberry-pi 0 and 3

matthewbauer:
Based off of 2.

### DIFF
--- a/raspberry-pi/0/default.nix
+++ b/raspberry-pi/0/default.nix
@@ -1,0 +1,32 @@
+{ lib, pkgs, ...}:
+
+{
+  boot = {
+    consoleLogLevel = lib.mkDefault 7;
+    extraTTYs = [ "ttyAMA0" ];
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi;
+    kernelParams = [
+      "dwc_otg.lpm_enable=0"
+      "console=ttyAMA0,115200"
+      "rootwait"
+      "elevator=deadline"
+    ];
+    loader = {
+      grub.enable = lib.mkDefault false;
+      generationsDir.enable = lib.mkDefault false;
+      raspberryPi = {
+        enable = lib.mkDefault true;
+        version = lib.mkDefault 1;
+      };
+    };
+  };
+
+  nix.buildCores = 1;
+
+  nixpkgs.config.platform = lib.systems.platforms.raspberrypi;
+
+  # cpufrequtils doesn't build on ARM
+  powerManagement.enable = lib.mkDefault false;
+
+  services.openssh.enable = lib.mkDefault true;
+}
--- a/raspberry-pi/3/default.nix
+++ b/raspberry-pi/3/default.nix
@@ -1,0 +1,30 @@
+{ lib, pkgs, ...}:
+
+{
+  boot = {
+    consoleLogLevel = lib.mkDefault 7;
+    extraTTYs = [ "ttyAMA0" ];
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi;
+    kernelParams = [
+      "dwc_otg.lpm_enable=0"
+      "console=ttyAMA0,115200"
+      "rootwait"
+      "elevator=deadline"
+      "cma=32M"
+    ];
+    loader = {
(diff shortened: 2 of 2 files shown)


Mic92:
Please also add a link to the README: https://github.com/NixOS/nixos-hardware#profiles

raspberry-pi/0/default.nix
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  nix.buildCores = 1;
+
+  nixpkgs.config.platform = lib.systems.platforms.raspberrypi;
+
+  # cpufrequtils doesn't build on ARM
+  powerManagement.enable = lib.mkDefault false;
+
+  services.openssh.enable = lib.mkDefault true;
```
Mic92: I would move the content to `raspberry-pi/default.nix` for all versions and only import the default.nix here for better sharing. It is still good to list every generation explicit however because then people don't need to change anything in case we change something for a particular model.

Mic92:
Have you tested rpi 0 so?

Mic92:
the wiki recommend the latest kernel for rpi2/rpi3:
https://nixos.wiki/wiki/NixOS_on_ARM

Mic92:
cc @samueldr

raspberry-pi/3/default.nix
```diff
@@ -0,0 +1,30 @@
...
+      grub.enable = lib.mkDefault false;
+      generationsDir.enable = lib.mkDefault false;
+      raspberryPi = {
+        enable = lib.mkDefault true;
+        version = lib.mkDefault 3;
+      };
+    };
+  };
+
+  nix.buildCores = 4;
```
mweinelt: While the device has 4 cores it has less than a 1GB of memory and will run out of memory like this.

raspberry-pi/0/default.nix
```diff
@@ -0,0 +1,32 @@
...
+      raspberryPi = {
+        enable = lib.mkDefault true;
+        version = lib.mkDefault 1;
+      };
+    };
+  };
+
+  nix.buildCores = 1;
+
+  nixpkgs.config.platform = lib.systems.platforms.raspberrypi;
```
yegortimoshenko: `hostPlatform`?

raspberry-pi/3/default.nix
```diff
@@ -0,0 +1,30 @@
...
+
+  nixpkgs.config.platform = lib.systems.platforms.aarch64-multiplatform;
+
+  services.openssh.enable = lib.mkDefault true;
```
yegortimoshenko: This is a hardware profile, so it should not include service configurations, like SSH.

yegortimoshenko: This is a hardware profile, so it should not include service configurations, like SSH.

mweinelt:
How about enabling zram-swap by default, to compensate for the imbalance between processor cores and available memory?

Mic92:
Looks like this PR got stalled. If someone wants to take this over, please feel free to make a new pull request.

edrex:
I just installed on an RPi3 and the existing wiki documentation about what kernel to use is out of date. Mainline works after fix for https://github.com/NixOS/nixpkgs/issues/97064. Guess I should get out of my comfort zone and make the edits

The line between what should go in nixpkgs vs nixos-hardware seems fuzzy. Are people concerned about community fragmentation weakening the center with moving to flakes?

edrex:
OT, I know. I'm posting here because I'm wondering if I should pick up this PR and dust it off. It would basically be the contents of the [arm config template on the wiki](https://nixos.wiki/wiki/NixOS_on_ARM#NixOS_installation_.26_configuration) (which is what currently says to use old kernel on pi3).

Mic92:
this is relevant in this regard: https://github.com/NixOS/rfcs/pull/70

Mic92:
> OT, I know. I'm posting here because I'm wondering if I should pick up this PR and dust it off. It would basically be the contents of the [arm config template on the wiki](https://nixos.wiki/wiki/NixOS_on_ARM#NixOS_installation_.26_configuration) (which is what currently says to use old kernel on pi3).

I can test it.

PR #97, opened 2019-02-06
dell/xps/13-9380: init

kalbasit:
This is pretty much a copy of XPS 9370, without kaby-laky and the throttle bug. I was getting lockups with the kaby-laky changes and the throttle bug did not seem to have an impact either way.


dell/xps/13-9380/README.wiki
```diff
@@ -0,0 +1,9 @@
+= Dell XPS 13 9380 =
+
+== Battery drain when sleeping ==
```
grahamc: You're my hero!

PR #102, opened 2019-02-28
t450s: fix trackpoint configuration at boot

ashkitten:
as i discussed in detail in [this issue on nixpkgs](https://github.com/NixOS/nixpkgs/issues/56512), the t450s (and likely other models) have a problem where they do not create the sysfs attributes for the trackpoint device early enough in the boot process for udev to configure them. this pull request adds a (sadly device-specific) method for triggering udev to configure the device once the sysfs path appears.

### DIFF
--- a/lenovo/thinkpad/t450s/default.nix
+++ b/lenovo/thinkpad/t450s/default.nix
@@ -6,4 +6,15 @@
     ../../../common/pc/laptop/acpi_call.nix
     ../.
   ];
+
+  systemd.paths.trackpoint = {
+    pathConfig = {
+      PathExists = "/sys/devices/rmi4-00/rmi4-00.fn03/serio2";
+      Unit = "trackpoint.service";
+    };
+  };
+
+  systemd.services.trackpoint.script = ''
+    ${config.systemd.package}/bin/udevadm trigger --attr-match=name="${config.hardware.trackpoint.device}"
+  '';
 }


lenovo/thinkpad/t450s/default.nix
```diff
@@ -6,4 +6,15 @@
     ../../../common/pc/laptop/acpi_call.nix
     ../.
   ];
+
+  systemd.paths.trackpoint = {
+    pathConfig = {
+      PathExists = "/sys/devices/rmi4-00/rmi4-00.fn03/serio2";
```
Mic92: Could you also add a brief comment here, what this line does as you did in the commit comment?
Otherwise looks good to me.

ashkitten: yeah np, i'll add that when i've got my t450s back in a working state (i've had to order a replacement touchpad for unrelated reasons)

ashkitten:
by the way, i'd like to be able to reuse this code without copy-paste for other devices that need it, but i don't know nix well enough to know do that (obviously the device path could be different)

Mic92:
You can just put it into a different file and use the `imports` feature.
In case other models require modifications, you can also create an option for that like here: https://github.com/Mic92/dotfiles/blob/master/nixos/vms/modules/retiolum.nix

ashkitten:
hi, sorry to say i won't be able to work on this pull request and help it to be merged as i would feel bad abandoning it afterwards. my touchpad and by extension trackpoint is Officially Not Working anymore and i've not been able to repair it.

PR #103, opened 2019-03-24
pcengines/apu: expose GRUB over serial

tomfitzhenry:
Prior to this commit, an apu's boot would go through stages of varying visibility:
1. BIOS: exposed to serial
2. GRUB: not exposed to serial
3. Linux: exposed to serial

This commit changes (2) to be visible over serial, so the entire boot is visible over serial.

Tested on nixos 18.09 on a https://www.pcengines.ch/apu1d.htm

References:
* https://www.gnu.org/software/grub/manual/grub/grub.html#Serial-terminal
* https://github.com/NixOS/nixpkgs/issues/37013#issuecomment-379226915

### DIFF
--- a/pcengines/apu/default.nix
+++ b/pcengines/apu/default.nix
@@ -1,3 +1,8 @@
 {
   boot.kernelParams = [ "console=ttyS0,115200n8" ];
+  boot.loader.grub.extraConfig = "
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks!

PR #107, opened 2019-05-08
Added basic support for the t470s

JorisE:
Also noted that the t480s was already supported but not on the list of profiles.


Mic92:
thanks!

PR #110, opened 2019-07-01
dell/xps/13-9380: also import intel profile

Mic92:


Mic92:
There is no AMD as an option: https://www.dell.com/en-uk/shop/laptops-notebooks-and-2-in-1-laptops/xps-13-laptop/spd/xps-13-9380-laptop

dell/xps/13-9380/default.nix
```diff
@@ -2,6 +2,7 @@
 
 {
   imports = [
+    ../../../common/cpu/intel
```
Mic92: Since all of them also have NVME, I wonder if we should have a NVME profile?

kalbasit: You joined the 9380 fan-base? :grinning: 

Yea I think that's fair to say.

Mic92: I was not aware that this is a cult already. The screen is nice. The battery in my old laptop lasted a longer.

Mic92: The NVME profile might be worth another PR.

PR #114, opened 2019-07-18
Initial 15-7590

tomberek:
Still working out bugs with Graphics, but the Wifi only working with 5.1 is an issue that will have to be readdressed.

### DIFF
--- a/dell/xps/15-7590/README.wiki
+++ b/dell/xps/15-7590/README.wiki
@@ -1,0 +1,46 @@
+= Dell XPS 15 7590 =
+*Mostly copied from 15-9550
+
+== Tested Hardware ==
+
+* CPU: Intel(R) Core(TM) i9-9980HK
+* RAM: 32 GB
+* HDD: 1 TiB SSD
+* Screen: 15" 4k (3840✕2160)
+* Input: Touchscreen and trackpad.
+
+== Firmware Configuration ==
+
+Not much tweaking of NixOS itself was needed. But we currently cannot automate the firmware setup, so this must be done by hand.
+
+=== Before installation ===
+
+These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as changing any other simple setting.
+
+* ''Disable Intel hardware RAID and use AHCI instead.'' Intel doesn't seem to provide a working linux driver for this.  (If you just have SSD it's pointless and just slows things down needlessly anyways.)
+
+=== Wifi ===
+Wifi does not work with kernels older than 5.1 (firmware not present) or newer (https://bbs.archlinux.org/viewtopic.php?id=247705)
+```
+  # Use the systemd-boot EFI boot loader.
+  boot.loader.systemd-boot.enable = true;
+  boot.loader.grub = {
+    device = "nodev";
+    efiSupport = true;
+    efiInstallAsRemovable = true;
+  };
+  boot.loader.efi.canTouchEfiVariables = true;
+  boot.kernelPackages = pkgs.linuxPackages_5_1;
+```
+=== After installation ===
+
+* ''Add systemd-boot to UEFI boot list.'' The (uneditable anyways) settings mapping drive UUIDs to HD* work fine.
+
+=== Optional ===
+
+* ''Update BIOS.'' According to Reddit, this helps with battery life.
+
+=== Troubleshooting ===
+
--- a/dell/xps/15-7590/README.wiki
+++ b/dell/xps/15-7590/README.wiki
@@ -22,7 +22,10 @@ These settings are needed both for booting the final install, and installer itse
 * ''Disable Intel hardware RAID and use AHCI instead.'' Intel doesn't seem to provide a working linux driver for this.  (If you just have SSD it's pointless and just slows things down needlessly anyways.)
 
 === Wifi ===
-Wifi does not work with kernels older than 5.1 (firmware not present) or newer (https://bbs.archlinux.org/viewtopic.php?id=247705)
+~~Wifi does not work with kernels older than 5.1 (firmware not present) or newer~~ (https://bbs.archlinux.org/viewtopic.php?id=247705)
+
+Update: The 48.ucode causes the Killer wifi card to crash. The iwlfwifi-cc-a0-46.ucode works perfectly. default.nix contains an overlay that removes the offending ucode from the linux-firmware bundle.
+
 ```
   # Use the systemd-boot EFI boot loader.
   boot.loader.systemd-boot.enable = true;
--- a/dell/xps/15-7590/README.wiki
+++ b/dell/xps/15-7590/README.wiki
@@ -37,6 +37,9 @@ Update: The 48.ucode causes the Killer wifi card to crash. The iwlfwifi-cc-a0-46
   boot.loader.efi.canTouchEfiVariables = true;
   boot.kernelPackages = pkgs.linuxPackages_5_1;
 ```
+Disable the `canTouchEfiVariables` after a boot or two to prevent NVRAM wearout.
+
+
 === After installation ===
 
 * ''Add systemd-boot to UEFI boot list.'' The (uneditable anyways) settings mapping drive UUIDs to HD* work fine.
--- a/dell/xps/15-7590/README.wiki
+++ b/dell/xps/15-7590/README.wiki
@@ -17,7 +17,7 @@ Not much tweaking of NixOS itself was needed. But we currently cannot automate t
 
 These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
 
-* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as changing any other simple setting.
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as changing any other simple setting.
 
 * ''Disable Intel hardware RAID and use AHCI instead.'' Intel doesn't seem to provide a working linux driver for this.  (If you just have SSD it's pointless and just slows things down needlessly anyways.)
 
--- a/dell/xps/15-7590/default.nix
+++ b/dell/xps/15-7590/default.nix
@@ -11,4 +11,24 @@
   boot.kernelPackages = pkgs.linuxPackages_5_1;
   boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
 
+  # The 48.ucode causes the Killer wifi card to crash.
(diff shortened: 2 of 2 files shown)


dell/xps/15-7590/README.wiki
```diff
@@ -0,0 +1,49 @@
...
+
+* CPU: Intel(R) Core(TM) i9-9980HK
+* RAM: 32 GB
+* HDD: 1 TiB SSD
+* Screen: 15" 4k (3840✕2160)
+* Input: Touchscreen and trackpad.
+
+== Firmware Configuration ==
+
+Not much tweaking of NixOS itself was needed. But we currently cannot automate the firmware setup, so this must be done by hand.
```
Mic92: Is it not supported by https://fwupd.org/ ?

dell/xps/15-7590/README.wiki
```diff
@@ -0,0 +1,49 @@
...
+
+=== Before installation ===
+
+These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as changing any other simple setting.
```
Mic92:
```suggestion
* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as changing any other simple setting.
```

dell/xps/15-7590/README.wiki
```diff
@@ -0,0 +1,49 @@
...
+
+```
+  # Use the systemd-boot EFI boot loader.
+  boot.loader.systemd-boot.enable = true;
+  boot.loader.grub = {
+    device = "nodev";
+    efiSupport = true;
+    efiInstallAsRemovable = true;
+  };
+  boot.loader.efi.canTouchEfiVariables = true;
```
Mic92: You want to disable this after installation or else you will wear out your NVRAM as it writes to it on every update.

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,34 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+  ];
+
+  # TODO: boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
```
Mic92: I don't think we should hard-code to systemd-boot here. Some people might prefer grub for different reasons.

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,34 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
```
Mic92: Do they still provide non-ssd options for this model? Otherwise we could also import the ssd profile.

Mic92:
Installation seems to be pretty much the same on all dell products (disable secure boot, disable raid mode). Maybe we could unify this somehow.

Camsbury:
@tomberek let me know if you have time to close this out soon. I am coming into one of these laptops soon for work and would love to have something official in place to work from.

dell/xps/15-7590/README.wiki
tomberek: Not familiar..

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,35 @@
...
+    ../../../common/pc/ssd
+  ];
+
+  # TODO: boot loader
+  #boot.loader.systemd-boot.enable = lib.mkDefault true;
+  boot.kernelPackages = pkgs.linuxPackages_5_1;
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
```
Mic92:
```suggestion
```

dell/xps/15-7590/README.wiki
```diff
@@ -0,0 +1,52 @@
...
+Update: The 48.ucode causes the Killer wifi card to crash. The iwlfwifi-cc-a0-46.ucode works perfectly. default.nix contains an overlay that removes the offending ucode from the linux-firmware bundle.
+
+```
+  # Use the systemd-boot EFI boot loader.
+  boot.loader.systemd-boot.enable = true;
+  boot.loader.grub = {
+    device = "nodev";
+    efiSupport = true;
+    efiInstallAsRemovable = true;
+  };
```
Mic92: This is either grub or systemd-boot right?

dell/xps/15-7590/README.wiki
```diff
@@ -0,0 +1,52 @@
...
+These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as changing any other simple setting.
+
+* ''Disable Intel hardware RAID and use AHCI instead.'' Intel doesn't seem to provide a working linux driver for this.  (If you just have SSD it's pointless and just slows things down needlessly anyways.)
+
+=== Wifi ===
+~~Wifi does not work with kernels older than 5.1 (firmware not present) or newer~~ (https://bbs.archlinux.org/viewtopic.php?id=247705)
+
+Update: The 48.ucode causes the Killer wifi card to crash. The iwlfwifi-cc-a0-46.ucode works perfectly. default.nix contains an overlay that removes the offending ucode from the linux-firmware bundle.
```
Mic92: What do you mean by `Killer` wifi card?

```suggestion
Update: The 48.ucode causes the wifi card to crash. The iwlfwifi-cc-a0-46.ucode works perfectly. default.nix contains an overlay that removes the offending ucode from the linux-firmware bundle.
```

Camsbury: That is the brand name.

Camsbury: "Killer" is the brand name.

dell/xps/15-7590/README.wiki
matthewbauer: fwupd is used to update firmware. It can't do anything like the below with custom configurations.

9999years:
Hey folks! Also using NixOS on the 15-7590. Here's some notes from my config:

- `boot.earlyVconsoleSetup` and `i18n.consoleFont = "latarcyrheb-sun32"` to make the boot logs a little less unreadable.
- `hardware.bumblebee.enable` gets the NVIDIA drivers working for me (but I *had* a lot of problems with them; not sure how it got resolved but there's nothing in my config other than that line).
- The EFI partition is really small (like, 200mb or something?) and got filled up when I upgraded the kernel to get WiFi drivers working. Might be worth putting a commented-out `boot.loader.grub.configurationLimit = 10` in the config to save someone else the pain of getting out of that situation.
- `services.xserver.dpi = 175` seems to be a decent value, but I haven't bothered to do the actual calculation.

Mic92:
> Hey folks! Also using NixOS on the 15-7590. Here's some notes from my config:
> 
>     * `boot.earlyVconsoleSetup` and `i18n.consoleFont = "latarcyrheb-sun32"` to make the boot logs a little less unreadable.
> 
>     * `hardware.bumblebee.enable` gets the NVIDIA drivers working for me (but I _had_ a lot of problems with them; not sure how it got resolved but there's nothing in my config other than that line).
> 
>     * The EFI partition is really small (like, 200mb or something?) and got filled up when I upgraded the kernel to get WiFi drivers working. Might be worth putting a commented-out `boot.loader.grub.configurationLimit = 10` in the config to save someone else the pain of getting out of that situation.
> 
>     * `services.xserver.dpi = 175` seems to be a decent value, but I haven't bothered to do the actual calculation.

You can also take this PR and add your changes if you want.

jonringer:
@Mic92 do you mind providing an example PR?

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+  ];
+
+  # Set to true for just the first run, then disable it.
+  # boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
+
+  # Earlier font-size setup
+  console.earlySetup = true;
```
Mic92: I think I will later factor this out into an high DPI module.

9999years: `console.earlySetup` is only valid in NixOS 20.03 onward; in 19.09 it's named `console.earlySetup`. Can we make it work for both versions, at least until NixOS 20.03 is stable?

Something like this should do it, I think:
```
  # ...
} // (if lib.versionAtLeast (lib.versions.majorMinor lib.version) "20.03"
      then { console.earlySetup = true; }
      else { boot.earlyVconsoleSetup = true; })
```

Mic92: @9999years yes. Just make a PR.

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,39 @@
...
+
+  # Prevent small EFI partiion from filling up
+  boot.loader.grub.configurationLimit = 10;
+
+  # The 48.ucode causes the Killer wifi card to crash.
+  # The iwlfwifi-cc-a0-46.ucode works perfectly
+  nixpkgs.pkgs = import <nixpkgs> {
+    config.allowUnfree = true;
```
rycee: This seems like a bad idea. It could cause somebody to unknowingly install unfree software.

dell/xps/15-7590/default.nix
```diff
@@ -0,0 +1,39 @@
...
+
+  # Earlier font-size setup
+  console.earlySetup = true;
+
+  # Prevent small EFI partiion from filling up
+  boot.loader.grub.configurationLimit = 10;
+
+  # The 48.ucode causes the Killer wifi card to crash.
+  # The iwlfwifi-cc-a0-46.ucode works perfectly
+  nixpkgs.pkgs = import <nixpkgs> {
```
rycee: Wouldn't this prevent the user from specifying their own `nixpkgs.config` options or doing cross platform builds of the configuration? I guess it may also interfere if you are deploying this configuration using nixops.

dell/xps/15-7590/default.nix
Mic92: We also have an option better suited for that called `hardware.enableRedistributableFirmware`.

dell/xps/15-7590/default.nix
Mic92: Please see: https://github.com/NixOS/nixos-hardware/pull/143

dell/xps/15-7590/default.nix
Mic92: Please see: https://github.com/NixOS/nixos-hardware/pull/143

PR #115, opened 2019-07-19
Replace ad hoc throttling fix with throttled

michaelpj:
`throttled` is a service designed to fix the same CPU throttling bug,
but it's an actively maintainted upstream project that we can use.

I'm unsure if we should make this change now. `throttled` is only in since [May](https://github.com/NixOS/nixpkgs/pull/62101), so this will only work for people on `nixos-unstable`. I don't know what our policy should be wrt to that.


Mic92:
Since it is a new service, we could backport throttled as well.

dell/xps/13-9370/default.nix
```diff
@@ -5,12 +5,13 @@
...
-    ../../../common/pc/laptop/cpu-throttling-bug.nix
   ];
 
   # Force S3 sleep mode. See README.wiki for details.
   boot.kernelParams = [ "mem_sleep_default=deep" ];
 
   # touchpad goes over i2c
   boot.blacklistedKernelModules = [ "psmouse" ];
+
+  services.throttled.enable = lib.mkDefault true;
```
Mic92: Should we not rather modify the profile?

Mic92: No strong opinion here. A single option line could be also good enough.

michaelpj: The module made sense because there were several things to do. Having a whole module which just enables a single option seems like overkill. Better to just enable it.

michaelpj:
Ah yes, I forgot about this. Updated - I changed some new users of `cpu-throttling-bug.nix` as well.

Mic92:
This one was just merged and also has the module: https://github.com/NixOS/nixos-hardware/pull/125/files

michaelpj:
Updated to catch that one too!

lenovo/thinkpad/t490/default.nix
```diff
@@ -26,5 +25,7 @@
 
     LABEL="pulseaudio_end"
     '';
+    
+    throttled.enable = lib.mkDefault true;
```
Mic92:
```suggestion
    services.throttled.enable = lib.mkDefault true;
```

michaelpj: It's inside an enclosing `sevices = { ... }`.

PR #116, opened 2019-07-23
enable fstrim on SSDs

Mic92:
This will run TRIM once a week using a systemd timer. Running TRIM regularly
will improves the performance and increases the SSDs lifespan. Since it is based
on the utillinux no additional package is required.

### DIFF
--- a/common/pc/ssd/default.nix
+++ b/common/pc/ssd/default.nix
@@ -4,4 +4,6 @@
   boot.kernel.sysctl = {
     "vm.swappiness" = lib.mkDefault 1;
   };
+
+  services.fstrim.enable = lib.mkDefault true;
 }


zarelit:
Does this work with encrypted drives? IIRC TRIM needs allow-discards to be enabled for it to work with LUKS devices

Mic92:
If it should be enabled for cryptsetup is up to the user in my opinion. It might have some security implications.

yegortimoshenko:
This does cancel out any potential for deniable encryption, but this certainly is not something that can be applied to a `/` LUKS partition. As-is, `fstrim` in this PR will be a non-op on LUKS partitions. To fix that, user will have to open their partition with `cryptsetup --allow-discards --persistent open /dev/sdaX root` once.

I guess we could alternatively set `rd.luks.options=discard` in `boot.kernelParams`, but that would result in enabling TRIM for all LUKS partitions, including external ones AFAIK, which are much more likely to benefit from deniable encryption (in particular, I think [DeLUKS](https://github.com/kriswebdev/cryptsetup-deluks) might be affected).

PR #126, opened 2019-10-16
Fix readme markdown syntax

adamse:


jonringer:
@adamse thanks for opening a PR :)

PR #128, opened 2019-10-30
Add Dell XPS-7390 Profile

Alecktos:
Initial profile for Dell XPS-7390, late 2019.

### DIFF
--- a/dell/xps/13-7390/default.nix
+++ b/dell/xps/13-7390/default.nix
@@ -4,6 +4,7 @@
   imports = [
     ../../../common/cpu/intel
     ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
   ];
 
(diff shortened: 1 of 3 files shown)


dell/xps/13-7390/default.nix
```diff
@@ -0,0 +1,10 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
```
Mic92: I think we can also enable the SSD module for trim support.

```suggestion
    ../../../common/pc/laptop
    ../../../common/pc/laptop/ssd
```

Mic92: As it comes with NVME by default.

PR #131, opened 2019-11-11
intel: add intel-media-driver

### LINKED ISSUE #99: Enable intel-media-driver?
cc: @yegortimoshenko 

Today I leaned that there is another driver from Intel that works with libva for accelerated video.

IF you add `intel-media-driver` to opengl packages, you can do this
`LIBVA_DRIVER_NAME=iHD vainfo` and see that a different driver is being used.

I don't really understand the pro/cons or any differences between using i965 and iHD, but I wanted to mention it since it doesn't seem widely documented/known.

I documented this on the related wiki page: https://nixos.wiki/wiki/Accelerated_Video_Playback

asymmetric:
Provides hardware accelerated video playback from Broadwell onwards.

Older boards (e.g. Sandy Bridge) are not supported. I'm not sure if they should specifically be excluded.

More info:

* https://github.com/intel/media-driver#supported-platforms
* https://wiki.archlinux.org/index.php/Hardware_video_acceleration#Intel

Closes #99.

### DIFF
--- a/common/cpu/intel/default.nix
+++ b/common/cpu/intel/default.nix
@@ -10,5 +10,6 @@
     vaapiIntel
     vaapiVdpau
     libvdpau-va-gl
+    intel-media-driver
   ];
 }


colemickens:
Sorry I missed your comment on the other thread, I don't know why I never sent it... this looks good to me too. Thanks!

jonringer:
I would like to verify that this doesn't break sandy bridge, but I have no way of doing so :(

wmertens:
I vote to merge and hope for the best. Other option is to inject /proc/cpuinfo into the configuration somehow

wmertens:
Should this also not enable the HuC firmware, whatever that is? https://github.com/intel/media-driver/blob/master/README.md#known-issues-and-limitations

Mic92:
Sandy bridge testers becomes quite rare these days.

JohnAZoidberg:
I've got sandy bridge but my chipset doesn't support the integrated video card :/

asymmetric:
@wmertens regarding HuC, there's some info on the [Arch wiki](https://wiki.archlinux.org/index.php/Intel_graphics#Enable_GuC_/_HuC_firmware_loading).

And a pretty nice chart of supported features on each architecture, courtesy of the [Gentoo wiki](https://wiki.gentoo.org/wiki/Intel#Feature_support).

asymmetric:
And it seems like HuC is only supported from Skywell onwards, which means we should make it conditional. Broadwells are still pretty common, I reckon.

PR #134, opened 2019-12-23
dell: add XPS 9570

matthewbauer:
This is my current laptop that I now have in a good enough place that I'd like to share. Most of it is taken from the Arch Wiki here:

https://wiki.archlinux.org/index.php/Dell_XPS_15_9570

Note that for reasonable performance, you really do need to use the NVIDIA proprietary drivers. I won't enable it by default, but will leave the snippet I'm using in.

### DIFF
--- a/dell/xps/15-9570/default.nix
+++ b/dell/xps/15-9570/default.nix
@@ -1,0 +1,38 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+  ];
+
+  # TODO: boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
+
+  nix.buildCores = lib.mkDefault 12;
+
+  # Recommended in Arch wiki, avoids shutdowns
+  boot.kernelParams = [
+    "mem_sleep_default=deep"
+    "nouveau.blacklist=0"
+    "acpi_osi=!"
+    "acpi_osi=\"Windows 2015\""
+    "acpi_backlight=vendor"
+  ];
+
+  # To just use Intel integrated graphics with Intel's open source driver
+  # hardware.nvidiaOptimus.disable = true;
+
+  # To setup proprietary NVIDIA Optimus properly.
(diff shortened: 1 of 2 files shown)


dell/xps/15-9570/default.nix
```diff
@@ -0,0 +1,38 @@
...
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+  ];
+
+  # TODO: boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
```
Mic92: Do not enable this by default. It should be only used during installation.
Enabling it by default will wear out the NVRAM since it would write new entries on every `nixos-rebuild switch`.

dell/xps/15-9570/default.nix
```diff
@@ -0,0 +1,38 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+  ];
+
+  # TODO: boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
```
Mic92: Is there anything inherently that would require systemd-boot over grub? If not than we should leave this up to the user to decide.

dell/xps/15-9570/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
+
+  nix.buildCores = lib.mkDefault 12;
```
Mic92: Maybe we should make this the default in NixOS:

```
nix.buildCores = lib.mkDefault 0;
```

This sets cores equal to the number of hardware threads in a system.

dell/xps/15-9570/default.nix
```diff
@@ -0,0 +1,38 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
```
Mic92: If all models come with at least an ssd, we could also enable:

```
../../common/pc/laptop/ssd
```

Mic92:
Please also link your profile in the README.

Mic92:
ping?

Mic92:
Closed due to lack of responsiveness. If everyone else wants to add this feel free to make a new pr.

PR #137, opened 2019-12-31
Add Lenovo ThinkPad X280 profile

mishudark:


Mic92:
Please also add a link in https://github.com/NixOS/nixos-hardware/blob/master/README.md

lenovo/thinkpad/x280/default.nix
```diff
@@ -0,0 +1,9 @@
+{
+  imports = [
+    ../.
+    ../../../common/cpu/intel
+    ../../../../common/pc/laptop/acpi_call.nix
```
Mic92: This one might make also sense:


```
../../common/pc/laptop/ssd
```

It seems like they sell SSDs for all devices.

mishudark: you are right, fixed

mishudark:
Link added to Readme.md

PR #138, opened 2020-01-02
Add Lenovo ThinkPad L13 profile

rummik:


Mic92:
Looks like we need to evaluate our profiles in serial to avoid OOM.

Mic92:
Please add your profile in the README as well.

rummik:
@Mic92 Updated

PR #142, opened 2020-02-07
Add Dell XPS 15-7590 suggested options

9999years:
These suggested options for X users are taken from my configuration, but aren't quite universal enough to be enabled by default.

### DIFF
--- a/dell/xps/15-7590/default.nix
+++ b/dell/xps/15-7590/default.nix
@@ -10,6 +10,13 @@
   # Set to true for just the first run, then disable it.
   # boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
 
+  # Load GPU drivers.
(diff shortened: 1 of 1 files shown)


dell/xps/15-7590/default.nix
```diff
@@ -10,6 +10,13 @@
   # Set to true for just the first run, then disable it.
   # boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
 
+  # Load GPU drivers.
+  # hardware.bumblebee.enable = lib.mkDefault true;
+
+  # High DPI for X users. 175 "looks reasonable" but I didn't do the actual DPI
+  # calculation.
+  # services.xserver.dpi = lib.mkDefault 175;
```
Mic92: Do you know how would one calculate that?

PR #143, opened 2020-02-09
xps/15-7590: don't enable unfree firmware by default

Mic92:
Also don't set nixpkgs.pkgs. Instead just add an overlay.


rycee:
Sorry for the delay. I think this is a good compromise! 👍

PR #147, opened 2020-02-19
travis: parallelize evaluation and limit RAM usage

Mic92:


Mic92:
Mhm. The error is quite hard to find now. Maybe I write a test runner that hides all the verbose output.

Mic92:
Let's see if serial execution gives better logs...

Mic92:
Also with serial execution the log is unreadable.

PR #148, opened 2020-03-03
dell/xps/13-7390: fix screen corruption issues

aaronjanse:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1855608

With this patch, the XPS 13 (7390)'s screen no longer randomly glitches (very noticeable, sometimes tough to undo and sometimes obstructing text). I've been using this patch for the past couple months on my XPS 13 (7390) and it has fixed the issue entirely.

On the other hand, this patch leads to the the Linux kernel being rebuild manually, so even though I think this code is good to document, I don't know if it should be merged.

### DIFF
--- a/dell/xps/13-7390/default.nix
+++ b/dell/xps/13-7390/default.nix
@@ -9,5 +9,34 @@
 
   boot.kernelParams = [ "mem_sleep_default=deep" ];
 
+  # fix screen corruption issues
+  boot.kernelPackages = lib.mkDefault
+    (let
+      linux_pkg =
+        { stdenv, buildPackages, fetchurl, perl, buildLinux, modDirVersionArg ? null, ... } @ args:
+        with stdenv.lib;
+        buildLinux (args // rec {
+          version = "5.3.16";
+
+          modDirVersion = if (modDirVersionArg == null) then concatStringsSep "." (take 3 (splitVersion "${version}.0")) else modDirVersionArg;
+
+          kernelPatches = [
+            {
+              name = "fix-display";
+              patch = pkgs.fetchpatch {
+                url = "https://bugs.freedesktop.org/attachment.cgi?id=144765";
+                sha256 = "sha256-Fc6V5UwZsU6K3ZhToQdbQdyxCFWd6kOxU6ACZKyaVZo=";
+              };
+            }
+          ];
+
+          src = fetchurl {
+            url = "mirror://kernel/linux/kernel/v5.x/linux-${version}.tar.xz";
+            sha256 = "19asdv08rzp33f0zxa2swsfnbhy4zwg06agj7sdnfy4wfkrfwx49";
+          };
+        } // (args.argsOverride or {}));
+        linux = pkgs.callPackage linux_pkg {};
+    in pkgs.recurseIntoAttrs (pkgs.linuxPackagesFor linux));
+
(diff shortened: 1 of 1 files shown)


Mic92:
Is this patch upstream or was reviewed? We could also integrate this patch directly upstream if it does not regress on other hardware.

andir:
On 01:22 03.03.20, Jörg Thalheim wrote:
> Is this patch upstream or was reviewed? We could also integrate this
> patch directly upstream if it does not regress on other hardware.

I searched through my mail archives but wasn't able to find that mail.

According to the bug entry it wasn't upstreamed yet due to missing
feedback: https://bugs.freedesktop.org/show_bug.cgi?id=110511#c32

The conversation continued over at
https://gitlab.freedesktop.org/drm/intel/issues/272 but wasn't resolved
from what I can say.

I also have a 9380 that is supposed to have flickering but I can't
really reproduce the effects seen there on my hardware. I occasionally
have a very short flickering of the screen. It feels like just a single
black frame.

If that fixes it for you then it might be worth responding to the linked
GitLab issue. If they aren't receptible of that we can just try to
submit that to the upstream kernel via their ML?

aaronjanse:
> If that fixes it for you then it might be worth responding to the linked GitLab issue. If they aren't receptible of that we can just try to submit that to the upstream kernel via their ML?

It looks like I didn't have much to add, so I left a comment pointing a fellow 7390 user to the patch I'm using.

> If they aren't receptible of that we can just try to submit that to the upstream kernel via their ML?

You mean the Linux kernel? That sounds awesome, and I'm definitely interested (even though I'd be submitting someone else's patch (under their name ofc)). However, I have zero experience submitting patches to the Linux kernel.

Do you have recommendations for how to move forward? I'm also open to putting this patch in Nixpkgs then waiting for freedesktop to figure things out (but it seems like that would take a while).

> if it does not regress on other hardware
I don't have much extra hardware to test on :-/

Mic92:
This issue might have been fixed by https://gitlab.freedesktop.org/drm/intel/issues/272#note_433464
in linux kernel version newer than 5.6rcX

aaronjanse:
I've updated to the latest 5.6 kernel in Nixpkgs. I can confirm that the screen issues are better than before, but I still get an occasional screen-full of random colors, something that never happened with the kernel patch above

kalbasit:
Can it be guarded with a falsy option?

aaronjanse:
What does it mean for something to be guarded with a falsy option?

kalbasit:
> What does it mean for something to be guarded with a falsy option?

```
options.nixos-hardware.fix-xps-13-issue.enable = mkEnableOption "fix... ";

config = mkIf (config.nixos-hardware.fix-xps-13-issue.enable) {
...
};
```

Something along these lines.

aaronjanse:
Oh, yeah, that would make sense. I think the primary goal is to make sure that XPS 7390 users at least know a patch exists. That option and/or this PR serve as documentation for the fix.

Do you think I should add the above option to the PR or simply closing this PR, allowing users to copy-paste the fix themselves when they find this PR via a search engine?

kalbasit:
I think you should add that option, and documented in a/the README somewhere and merge it down. it’s better for people to look/find documentation rather than deal with patches.

PR #149, opened 2020-03-11
lenovo/thinkpad/t495: init from e495.

danderson:
The E495 and T495 are based on the same core Ryzen platform, so the
E495 tweaks apply similarly to the T495.

Tested on my own T495, works well.


Mic92:
Please link to the profile in the README table

danderson:
@Mic92 added to README, and rebased. Thanks for the review, and sorry I missed the instructions!

jonringer:
the linked issue also mentions adding:
```
hardware.enableRedistributableFirmware = true;
```
is that needed as well?

danderson:
Empirically no: it was off on my laptop before I started using nixos-hardware, and the machine booted to X11 and worked just fine. It means we miss out on newer Intel wifi/bluetooth firmware and AMD microcode, but it looks like the burned-in firmware and microcode that ships with the T495 is "good enough".

I don't know what our policy is for that kind of thing? I think it's a good idea to `enableRedistributableFirmware=true`, but it's not strictly required and it might annoy people who don't want non-free blobs in their NixOS.

PR #152, opened 2020-03-23
Added Librem15v3

ngiger:
* Documented Librem laptop 15v3.
* Added a hint in the README.md on howto dual boot PureOS and NixSO
* Added a hint in the README.md on Automatically lock the desktop when removing the librem key


README.md
```diff
@@ -73,7 +73,7 @@ See code for all available configurations.
 | PC Engines APU                    | `<nixos-hardware/pcengines/apu>`                   |
 | [Raspberry Pi 2][]                | `<nixos-hardware/raspberry-pi/2>`                  |
 | [Samsung Series 9 NP900X3C][]     | `<nixos-hardware/samsung/np900x3c>`                |
-| [Purism Librem 13v3][]            | `<nixos-hardware/purism/librem/13v3>`              |
+| [Purism Librem 13v3/15v3][]       | `<nixos-hardware/purism/librem/13v3>`              |
```
Mic92: I would still add a second file even if we just make it a symlink for now.
The reason is that it allows us to introduce two files in future without breaking compatibility.

ngiger: No problem. I was considering this approach, too. Pushed a commit to resolve this issue.

README.md
```diff
@@ -74,6 +74,7 @@ See code for all available configurations.
 | [Raspberry Pi 2][]                | `<nixos-hardware/raspberry-pi/2>`                  |
 | [Samsung Series 9 NP900X3C][]     | `<nixos-hardware/samsung/np900x3c>`                |
 | [Purism Librem 13v3][]            | `<nixos-hardware/purism/librem/13v3>`              |
+| [Purism Librem 15v3][]            | `<nixos-hardware/purism/librem/15v3>`              |
```
Mic92: This one also needs a reference below the table as we do for the other profiles to become a valid hyperlink.

PR #154, opened 2020-04-21
Change TLP charge thresholds

j0hax:
Set TLP charging thresholds to values recommended by the TLP developer on [ThinkPad-Forum.de](https://thinkpad-forum.de/threads/184510-elementary-OS-Thinkpad-X230-einrichten?p=1865345&viewfull=1#post1865345).

### DIFF
--- a/lenovo/thinkpad/x230/default.nix
+++ b/lenovo/thinkpad/x230/default.nix
@@ -16,4 +16,10 @@
   services.xserver.deviceSection = lib.mkDefault ''
     Option "TearFree" "true"
   '';
+
+  services.tlp.extraConfig = lib.mkDefault ''
+    START_CHARGE_THRESH_BAT0=67
+    STOP_CHARGE_THRESH_BAT0=100
+  '';
+
 }


Mic92:
The thread does not quite look like TLP developer but rather users. Actually in the wiki article it says that one has to recalibrate the battery to reach the full capacity again. So I am not quite sure how they came up with the thresholds here.

j0hax:
The [ArchWiki](https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_X230#TLP) claims there is a hardware bug and also references the thread. Maybe the original developer, @linrunner, has further information?

Mic92:
It survived the archlinux wiki. Let's keep it.

linrunner:
I would advise against activating charge thresholds ootb in a Linux installation. Better to point users to the TLP FAQ so they can make their choice: https://linrunner.de/en/tlp/docs/tlp-faq.html#battery

67/100 are my personal settings for the X220 because of this: https://linrunner.de/en/tlp/docs/tlp-faq.html#erratic-battery

Mic92:
@linrunner so you would not recommend this setting on x220 by default? Note that this configuration is applied to x220 only.

linrunner:
As i said above. TLP *enables* users to use charge thresholds but won't *enforce* them. You have to consider that they come at the cost of a more or less reduced battery runtime.

PR #157, opened 2020-05-03
Add a shared HiDPI Configuration, use it.

pmyjavec:
I've noticed other laptops with HiDPI displays have this configuration set  and my laptop didn't, so I thought it might be worth adding a common configuration for basic HiDPI display settings which seem quite universal.

### DIFF
--- a/common/pc/display/hidpi.nix
+++ b/common/pc/display/hidpi.nix
@@ -1,13 +1,17 @@
-# Shared configuration which most (if not all) HiDPI setups should have.
-{ config, lib, pkgs, ... }:
+{ pkgs, lib, ... }: {
+  console.font = lib.mkDefault
+    "${pkgs.terminus_font}/share/consolefonts/ter-u28n.psf.gz";
 
-{
+  # Needed when doing cryptsetup
+  console.earlySetup = lib.mkDefault true;
+  boot.loader.systemd-boot.consoleMode = lib.mkDefault "max";
+
   services.xserver.dpi = lib.mkDefault 196;
   fonts.fontconfig.dpi = lib.mkDefault 196;
+  environment.variables = lib.mkDefault {
+    GDK_SCALE = "2";
+    GDK_DPI_SCALE = "0.5";
+    _JAVA_OPTIONS = "-Dsun.java2d.uiScale=2";
 
-  # Fix sizes of GTK/GNOME ui elements
-  environment.variables = {
-    GDK_SCALE = lib.mkDefault "2";
-    GDK_DPI_SCALE= lib.mkDefault "0.5";
   };
 }
--- a/common/pc/display/hidpi.nix
+++ b/common/pc/display/hidpi.nix
@@ -1,0 +1,13 @@
+# Shared configuration which most (if not all) HiDPI setups should have.
+{ config, lib, pkgs, ... }:
+
+{
+  services.xserver.dpi = lib.mkDefault 196;
+  fonts.fontconfig.dpi = lib.mkDefault 196;
+
+  # Fix sizes of GTK/GNOME ui elements
+  environment.variables = {
+    GDK_SCALE = lib.mkDefault "2";
+    GDK_DPI_SCALE= lib.mkDefault "0.5";
+  };
+}
--- a/lenovo/thinkpad/x1-extreme/gen2/default.nix
+++ b/lenovo/thinkpad/x1-extreme/gen2/default.nix
@@ -5,6 +5,7 @@ with lib;
 {
   imports = [
     ../.
+    ../../../../common/pc/display/hidpi.nix
   ];
 
   # Fixes an issue with incorrect battery reporting. See


Mic92:
Agreed!

common/pc/display/hidpi.nix
```diff
@@ -0,0 +1,13 @@
...
+{ config, lib, pkgs, ... }:
+
+{
+  services.xserver.dpi = lib.mkDefault 196;
+  fonts.fontconfig.dpi = lib.mkDefault 196;
+
+  # Fix sizes of GTK/GNOME ui elements
+  environment.variables = {
+    GDK_SCALE = lib.mkDefault "2";
+    GDK_DPI_SCALE= lib.mkDefault "0.5";
```
Mic92, line 11: How about this?

```
{ pkgs, lib, ... }: {
  console.font = lib.mkDefault
    "${pkgs.terminus_font}/share/consolefonts/ter-u28n.psf.gz";
  console.earlySetup = lib.mkDefault true;
  boot.loader.systemd-boot.consoleMode = lib.mkDefault "max";

  services.xserver.dpi = lib.mkDefault 196;
  fonts.fontconfig.dpi = lib.mkDefault 196;
  environment.variables = lib.mkDefault {
    GDK_SCALE = "2";
    GDK_DPI_SCALE = "0.5";
    _JAVA_OPTIONS = "-Dsun.java2d.uiScale=2";
  };
}
```

Mic92: This adds the linux consol, java and systemd-boot.

wmertens:
How does this behave when you plug in a FHD display? Xrandr/Wayland will do the right thing?

common/pc/display/hidpi.nix
pmyjavec: @Mic92 sweet, I'll append that.

utdemir:
Sorry, this is a bit off topic, but is there a way to check if the display I have is HiDPI?

common/pc/display/hidpi.nix
pmyjavec: @Mic92, I made the changes that you suggested, not had time to test it yet, but the tests are passing.

Thank you for the feedback!

pmyjavec:
@utdemir I guess you could check the available resolutions as presented by `xrandr`? It's not a bad suggestion, a super crude way of doing this might be:

`xrandr | grep -q 3840x216 && echo "HiDPI found"`

I'm sure there are much

pmyjavec:
> Sorry, this is a bit off topic, but is there a way to check if the display I have is HiDPI? If there is, it might be worth putting it as a comment somewhere.



> How does this behave when you plug in a FHD display? Xrandr/Wayland will do the right thing?

@wmertens , sorry I don't use wayland!

Mic92:
cc @colemickens Can you have a look regarding wayland. I also only tested them on Xorg.

Mic92:
I wonder if we should move this module to nixpkgs? We could include it in nixos-generators if we detect a large screen on installation. This way also desktops are covered.

pmyjavec:
@Mic92 I think that would greatly improve the user experience when first using NixOS! Especially as retina screens become more prominent.

matthewbauer:
> I wonder if we should move this module to nixpkgs? We could include it in nixos-generators if we detect a large screen on installation. This way also desktops are covered.

HiDPI isn’t really a property of the machine, but the display. I think it’s probably okay to put it in hardware-configuration.nix for laptops which have a fixed display but Xorg and fontconfig should be able to detect the DPI at runtime.

Mic92:
> > I wonder if we should move this module to nixpkgs? We could include it in nixos-generators if we detect a large screen on installation. This way also desktops are covered.
> 
> HiDPI isn’t really a property of the machine, but the display. I think it’s probably okay to put it in hardware-configuration.nix for laptops which have a fixed display but Xorg and fontconfig should be able to detect the DPI at runtime.

They don't adjust at runtime and I don't know why. @lheckemann has added hidpi detection to our configuration generator: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-generate-config.pl#L501

Even if we don't enable it by default I still see some value if it would be just one import away in NixOS.

Mic92:
I opened a pull request: https://github.com/NixOS/nixpkgs/pull/87730

pmyjavec:
Did we still want this change? Is it just awaiting further testing?

Mic92:
I made a new PR here: https://github.com/NixOS/nixpkgs/pull/88669

bbigras:
Now that the PR is merged. What does this PR needs?

I can help test on wayland (sway).

common/pc/display/hidpi.nix
```diff
@@ -0,0 +1,17 @@
...
+  # Needed when doing cryptsetup
+  console.earlySetup = lib.mkDefault true;
+  boot.loader.systemd-boot.consoleMode = lib.mkDefault "max";
+
+  services.xserver.dpi = lib.mkDefault 196;
+  fonts.fontconfig.dpi = lib.mkDefault 196;
+  environment.variables = lib.mkDefault {
+    GDK_SCALE = "2";
+    GDK_DPI_SCALE = "0.5";
+    _JAVA_OPTIONS = "-Dsun.java2d.uiScale=2";
```
Mic92, line 14: I think those environment variables might brake wayland. We cannot just check if `x11` is enabled since wayland will also likely enable x11 for xwayland.

common/pc/display/hidpi.nix
```diff
@@ -0,0 +1,17 @@
+{ pkgs, lib, ... }: {
+  console.font = lib.mkDefault
+    "${pkgs.terminus_font}/share/consolefonts/ter-u28n.psf.gz";
+
+  # Needed when doing cryptsetup
+  console.earlySetup = lib.mkDefault true;
+  boot.loader.systemd-boot.consoleMode = lib.mkDefault "max";
```
Mic92, line 7:
```suggestion
  hardware.video.hidpi.enable = true;
```

common/pc/display/hidpi.nix
```diff
@@ -0,0 +1,17 @@
...
+
+  services.xserver.dpi = lib.mkDefault 196;
+  fonts.fontconfig.dpi = lib.mkDefault 196;
```
Mic92, line 10: This is what I currently:

```suggestion
  services.xserver.displayManager.setupCommands = ''
    ${pkgs.xlibs.xrandr}/bin/xrandr --dpi eDP-1
  '';
```

It simply sets the dpi of my 4k screen as the dpi of x11. Not sure how to make the `eDP-1` part more general.

Mic92:
Actually all improvements could also go to the hidpi module in nixpkgs...

Mic92:
nixos-generate-config already detects screen sizes and enables the module automatically hence we don't need to include it for every profile.

pmyjavec:
@Mic92,

Not sure if you read my original comment, but none of this worked for me out of the box, it took me maybe a day or two to actual workout what I needed.

Maybe it's better now?

Mic92:
@pmyjavec What I am saying if you want to improve `HiDPI` than please contribute to the NixOS module that is now part of nixpkgs. I also use some of these options and even documented them in the wiki: https://nixos.wiki/wiki/Xorg However they seem not work for every setup yet.

pmyjavec:
@Mic92 ok nice, thanks for clarifying.

PR #158, opened 2020-05-04
New dell/latitude/3480

ehmry:
The touchpad is painful to use unless the "psmouse" kernel module
is disabled.


Mic92:
Do you need further testing?

ehmry:
Yes I do, I thought I fixed it but the touchpad stops working intermittently.

ehmry:
Well this module doesn't fix all the issues with the touchpad but its an improvement. Ready to merge.

PR #164, opened 2020-05-16
lenovo/thinkpad/x1/7th-gen: init

phunehehe:
Not sure how this fits with the "opinionated" bit in CONTRIBUTING.md, but I see similar stuff in lenovo/thinkpad/t490 so maybe it's OK?

This only works with unstable for now, until SOF is in stable (https://github.com/NixOS/nixpkgs/pull/87988).

### DIFF
--- a/lenovo/thinkpad/x1/7th-gen/audio.nix
+++ b/lenovo/thinkpad/x1/7th-gen/audio.nix
@@ -1,0 +1,15 @@
+{ lib, pkgs, ... }:
+{
+  # This can be removed when the default kernel is at least version 5.6
+  # https://github.com/NixOS/nixpkgs/pull/86168
+  boot.kernelPackages = lib.mkIf
+    (lib.versionOlder pkgs.linux.version "5.6")
+    (lib.mkDefault pkgs.linuxPackages_5_6);
+
+  # This can be removed when PulseAudio is at least version 14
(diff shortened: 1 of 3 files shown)


lenovo/thinkpad/x1/7th-gen/audio.nix
```diff
@@ -0,0 +1,13 @@
+{ lib, pkgs, ... }:
+{
+  # This can be removed when the default kernel is at least version 5.5
+  # https://github.com/NixOS/nixpkgs/pull/86168
+  boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_5_6;
```
Mic92, line 5: How about

```suggestion
  boot.kernelPackages = mkIf (stdenv.lib.versionOlder kernel.version "5.6") lib.mkDefault pkgs.linuxPackages_latest;
```

This way if our default kernel is not 5.6 and the user has not changed the default kernel, we use the latest kernel, which is 5.6 at the time.

Mic92: Should be

Mic92:
Please add your profile to the table so it is picked up by CI: https://github.com/NixOS/nixos-hardware#how-to-contribute-a-new-device-profile

lenovo/thinkpad/x1/7th-gen/audio.nix
phunehehe: Thanks! I adopted that with one change: use 5.6 instead of latest, since that's the version that has been tested.

PR #169, opened 2020-05-20
pc/laptop: remove redundant cpuFreqGovernor setting

erikarvstedt:
This setting has been part of NixOS since version 16.09.

It's defined in [services/hardware/tlp.nix](https://github.com/NixOS/nixpkgs/blob/82b5f87fcc710a99c47c5ffe441589807a8202af/nixos/modules/services/hardware/tlp.nix#L59).

### DIFF
--- a/common/pc/laptop/default.nix
+++ b/common/pc/laptop/default.nix
@@ -3,11 +3,5 @@
 {
   imports = [ ../. ];
 
-  # TODO: fix in NixOS/nixpkgs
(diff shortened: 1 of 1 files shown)


wmertens:
I don't see it in that file, only setting it to null?

erikarvstedt:
Yes, `cpuFreqGovernor` is set to `null` when `tlp` is enabled.

PR #172, opened 2020-06-21
README.md: update description of the project

Mic92:


rnhmjoj:
While you are at it: I've just found a typo: `Purism Librem 15v3`, should be a 13, not 15. That's why you see brackets instead of the right link.

PR #174, opened 2020-06-21
add xps 13 (9343)

bbigras:
I have no idea if `throttled` and `thermald` are needed since `common/pc/laptop` has `tlp`. I copied them from another profile.

Also I wonder if `boot.extraModulePackages = [ config.boot.kernelPackages.broadcom_sta ];` should be in the profile since `hardware-configuration.nix` already has it. I wonder if it could help people figure out that they need to make a custom iso with that driver to be able to install with the wifi. Maybe it could be noted in a README.

> With kernel 4.6.5 and tlp, the idle power usage can reach ~2.3 W with the kernel parameter pcie_aspm=force enabled. 

https://wiki.archlinux.org/index.php/Dell_XPS_13_(9343)#Powersaving
I also wonder if this should be included or only noted in a README. If included I'm not sure how the user could disable it.

### DIFF
--- a/dell/xps/13-9343/default.nix
+++ b/dell/xps/13-9343/default.nix
@@ -1,0 +1,12 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+  ];
+
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
+}
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ See code for all available configurations.
 | Dell Latitude 3480                | `<nixos-hardware/dell/latitude/3480>`              |
 | [Dell XPS E7240][]                | `<nixos-hardware/dell/e7240>`                      |
 | [Dell XPS 13 7390][]              | `<nixos-hardware/dell/xps/13-7390>`                |
+| [Dell XPS 13 9343][]              | `<nixos-hardware/dell/xps/13-9343>`                |
 | [Dell XPS 13 9360][]              | `<nixos-hardware/dell/xps/13-9360>`                |
 | [Dell XPS 13 9370][]              | `<nixos-hardware/dell/xps/13-9370>`                |
 | [Dell XPS 13 9380][]              | `<nixos-hardware/dell/xps/13-9380>`                |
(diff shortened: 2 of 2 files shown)


bbigras:
friendly ping @Mic92 @yegortimoshenko

dell/xps/13-9343/default.nix
```diff
@@ -0,0 +1,14 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services.throttled.enable = lib.mkDefault true;
```
Mic92, line 10: You can checkout this https://www.reddit.com/r/thinkpad/comments/870u0a/t480s_linux_throttling_bug/ and see if it applies to you.

dell/xps/13-9343/default.nix
```diff
@@ -0,0 +1,14 @@
...
+
+  # This will save you money and possibly your life!
+  services.thermald.enable = true;
```
Mic92, line 13: This seems like a good default for laptops in general. @yegortimoshenko should we enable it?

dell/xps/13-9343/default.nix
bbigras: Oh I didn't realize `throttled` was to fix some problem. I probably don't need it.

But it's weird. If I test without it (with `s-tui with `stress`) my CPU cap at about 2420 Mhz. The clock doesn't seem to go down after like 10 secs like the person had in the linked issue.

If I test with `throttled`, my cpu cap at about 2350 Mhz.

It's also weird that my max with turbo is 2.70 GHz. I wonder if the temp doesn't allow it to go higher.

```
| Processor Base Frequency | 2.20 GHz |
| Max Turbo Frequency      | 2.70 GHz |
```
https://ark.intel.com/content/www/us/en/ark/products/85212/intel-core-i5-5200u-processor-3m-cache-up-to-2-70-ghz.html

Should I remove `throttled` and try to find out on my own why I don't hit 2.70 GHz?

Mic92: Yes. Just remove it in that case.

bbigras: Thanks

bbigras:
Should I use `lib.mkDefault` with `services.thermald.enable`?

Mic92:
> Should I use `lib.mkDefault` with `services.thermald.enable`?

Yes.

bbigras:
Thanks. I made the change.

bbigras:
Thanks!! :smile:

PR #176, opened 2020-07-06
add flake support

Mic92:


flake.nix
```diff
@@ -0,0 +1,63 @@
...
+      lenovo-thinkpad-t480s = import ./lenovo/thinkpad/t480s;
+      lenovo-thinkpad-t490 = import ./lenovo/thinkpad/t490;
+      lenovo-thinkpad-t495 = import ./lenovo/thinkpad/t495;
+      lenovo-thinkpad-x140e = import ./lenovo/thinkpad/x140e;
+      lenovo-thinkpad-x220 = import ./lenovo/thinkpad/x220;
+      lenovo-thinkpad-x230 = import ./lenovo/thinkpad/x230;
+      lenovo-thinkpad-x250 = import ./lenovo/thinkpad/x250;
+      lenovo-thinkpad-x260 = import ./lenovo/thinkpad/x260;
+      lenovo-thinkpad-x270 = import ./lenovo/thinkpad/x270;
+      lenovo-thinkpad-x280 = import ./lenovo/thinkpad/x280;
```
yegortimoshenko, line 47: I wonder what would be preferable, `lenovo-thinkpad-x280` or `lenovo.thinkpad.x280`.

Mic92: I am not sure if we can use `lenovo.thinkpad.x280` i.e. if flakes support nesting under `nixosModules`. The current implementation seems like the safest choice to me.

Mic92: Nesting is not supported according to: https://github.com/NixOS/rfcs/pull/49#discussion_r450275707

PR #186, opened 2020-08-24
Add support for System76 hardware, specifically the Darter Pro 6

khumba:
This integrates the System76 hardware support at https://github.com/stites/system76-nixos into this repo.  @stites did the bulk of the work in that repo, and I added support for the Darter Pro.

This PR *only* includes the kernel modules for now, along with some NixOS options for controlling which modules are enabled.  These kernel modules are needed for multimedia keys and keyboard backlights to work on System76's laptops.  System76 provides also some daemons and tools for hardware management, and some of these are included in the other repo, but I haven't had the time/need/interest use them yet.

I have the darp6 and all hardware works, with the two `system76-acpi` and `system76-io` modules.  I can't personally test the `system76` module as my laptop doesn't need it, but I'm including it with minimal changes from @stites's repo on the assumption that it still works fine (though I'm bumping the version from 1.0.6 to 1.0.9, and dropping the initramfs-tools files from all of the kernel module packages, since they appear unused).

@stites: The license on your repo is BSD-3, and the license on this repo is CC0.  Are you okay with including your code here?  I see that your readme says that you intend to move things over, but we should get an explicit ACK from you on the license change before merging this.

Thanks!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -56,6 +56,8 @@
       pcengines-apu = import ./pcengines/apu;
       raspberry-pi-2 = import ./raspberry-pi/2;
       samsung-np900x3c = import ./samsung/np900x3c;
+      system76 = import ./system76;
+      system76-darp6 = import ./system76/darp6;
       purism-librem-13v3 = import ./purism/librem/13v3;
       purism-librem-15v3 = import ./purism/librem/15v3;
       supermicro-a1sri-2758f = import ./supermicro/a1sri-2758f;
--- a/system76/darp6/default.nix
+++ b/system76/darp6/default.nix
@@ -2,8 +2,20 @@
 #
 # https://system76.com/laptops/darter
 
-{ config, ... }:
+{ config, lib, options, pkgs, ... }:
 let
+  cfg = config.hardware.system76.darp6;
+
+  # Allow silencing the warning about these options if either is defined.
+  soundSettingsDefined =
+    options.hardware.system76.darp6.soundVendorId.isDefined ||
+    options.hardware.system76.darp6.soundSubsystemId.isDefined;
+
+  # We neeed both options non-null to be able to apply the headset fixup though.
+  soundSettingsAvailable =
+    soundSettingsDefined &&
+    (cfg.soundVendorId != null && cfg.soundSubsystemId != null);
+
   # darp6 needs system76-acpi-dkms, not system76-dkms:
   #
   # [1] https://github.com/pop-os/system76-dkms/issues/39
@@ -19,9 +31,73 @@ in
     ../../common/pc/laptop/ssd
   ];
 
-  boot.extraModulePackages = packages;
+  options.hardware.system76.darp6 = {
+    soundVendorId = lib.mkOption {
+      type = with lib.types; nullOr (strMatching "0x[0-9a-f]{8}");
+      description = ''
+        The vendor ID of the sound card PCI device, for applying the headset fixup.
+        This should be set to the value of the following file on your Darter Pro:
+            /sys/class/sound/hwC0D0/vendor_id
+        If this option has the default null value, then the headset fixup will
+        not be applied.
+      '';
+    };
+    soundSubsystemId = lib.mkOption {
+      type = with lib.types; nullOr (strMatching "0x[0-9a-f]{8}");
+      description = ''
+        The subsystem ID of the sound card PCI device, for applying the headset fixup.
+        This should be set to the value of the following file on your Darter Pro:
+            /sys/class/sound/hwC0D0/subsystem_id
+        If this option has the default null value, then the headset fixup will
+        not be applied.
+      '';
+    };
+  };
 
-  # system76_acpi automatically loads on darp6, but system76_io does not.
-  # Explicitly list all modules to be loaded, for consistency.
-  boot.kernelModules = map (drv: drv.moduleName) packages;
+  config = lib.mkMerge [
+    {
+      boot.extraModulePackages = packages;
+
+      # system76_acpi automatically loads on darp6, but system76_io does not.
+      # Explicitly list all modules to be loaded, for consistency.
+      boot.kernelModules = map (drv: drv.moduleName) packages;
+
+      warnings = lib.optional (!soundSettingsDefined) ''
+        For full Darter Pro support, set the options:
+        - hardware.system76.darp76.soundVendorId
+        - hardware.system76.darp76.soundSubsystemId
+        You can copy these values directly from:
+        - /sys/class/sound/hwC0D0/vendor_id
+        - /sys/class/sound/hwC0D0/subsystem_id
+        The headset audio fixup will not be applied without these values.
+        Set these options to null to silence this warning.
+      '';
+    }
+
+    # Apply the headset fixup patch from system76-driver, if the necessary
+    # options have been provided.
+    #
+    # See occurrences of "darp" in:
+    # https://github.com/pop-os/system76-driver/blob/master/system76driver/actions.py
+    # https://github.com/pop-os/system76-driver/blob/master/system76driver/products.py
+    (lib.mkIf soundSettingsAvailable {
+      boot.extraModprobeConfig = ''
+        options snd-hda-intel model=headset-mode-no-hp-mic patch=system76-audio-patch
+      '';
+
+      hardware.firmware = [
+        (pkgs.writeTextFile {
+          name = "system76-audio-patch";
+          destination = "/lib/firmware/system76-audio-patch";
+          text = ''
+            [codec]
+            ${cfg.soundVendorId} ${cfg.soundSubsystemId} 0
+
+            [pincfg]
+            0x1a 0x01a1913c
+          '';
+        })
+      ];
+    })
+  ];
 }
--- a/system76/darp6/default.nix
+++ b/system76/darp6/default.nix
@@ -1,0 +1,27 @@
+# Hardware profile for the Darter Pro 6 laptop by System76.
+#
+# https://system76.com/laptops/darter
+
+{ config, ... }:
+let
+  # darp6 needs system76-acpi-dkms, not system76-dkms:
+  #
+  # [1] https://github.com/pop-os/system76-dkms/issues/39
+  # jackpot51> system76-acpi-dkms is the correct driver to use on the darp6
+  #
+  # system76-io-dkms also appears to be loaded on darp6 with Pop!_OS, and
+  # system76-dkms does not, and in fact refuses to load.
+  packages = with config.boot.kernelPackages; [ system76-acpi system76-io ];
+in
+{
+  imports = [
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+  ];
+
+  boot.extraModulePackages = packages;
+
+  # system76_acpi automatically loads on darp6, but system76_io does not.
+  # Explicitly list all modules to be loaded, for consistency.
+  boot.kernelModules = map (drv: drv.moduleName) packages;
+}
--- a/system76/darp6/default.nix
+++ b/system76/darp6/default.nix
@@ -1,6 +1,14 @@
 # Hardware profile for the Darter Pro 6 laptop by System76.
 #
 # https://system76.com/laptops/darter
+#
+# Regarding kernel modules, darp6 needs system76-acpi-dkms, not system76-dkms:
+#
+# [1] https://github.com/pop-os/system76-dkms/issues/39
+# jackpot51> system76-acpi-dkms is the correct driver to use on the darp6
+#
+# system76-io-dkms also appears to be loaded on darp6 with Pop!_OS, and
+# system76-dkms does not, and in fact refuses to load.
 
 { config, lib, options, pkgs, ... }:
 let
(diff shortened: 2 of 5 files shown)


system76/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  cfg = config.hardware.system76;
+
+  useDkms = cfg.drivers.system76.enable;
+  useAcpiDkms = cfg.drivers.system76-acpi.enable;
+  useIoDkms = cfg.drivers.system76-io.enable;
+
+  kernelPackages = let inherit (config.boot.kernelPackages) callPackage; in {
+    system76 = (callPackage pkgs/system76-dkms {});
+    system76-acpi = (callPackage pkgs/system76-acpi-dkms {});
+    system76-io = (callPackage pkgs/system76-io-dkms {});
```
Mic92, line 15: Could you add those packages to nixpkgs itself? We don't have the infrastructure in this repository to test builds and this way it would also be cached by hydra.
If you link the PR here, I will review it.

system76/pkgs/system76-io-dkms/default.nix
```diff
@@ -0,0 +1,40 @@
...
+  hardeningDisable = [ "pic" ];
+  dontStrip = true;
+  dontPatchELF = true;
+
+  kernel = kernel.dev;
+  nativeBuildInputs = kernel.moduleBuildDependencies;
+
+  preBuild = ''
+    sed -e "s@/lib/modules/\$(.*)@${kernel.dev}/lib/modules/${kernel.modDirVersion}@" -i Makefile
+  '';
```
Mic92, line 26: This should do:

```
buildFlags = [
    "KERNELDIR=${kernel.dev}/lib/modules/${kernel.modDirVersion}/build"
];
```

system76/pkgs/system76-dkms/default.nix
```diff
@@ -0,0 +1,44 @@
...
+  hardeningDisable = [ "pic" ];
+  dontStrip = true;
+  dontPatchELF = true;
+
+  kernel = kernel.dev;
+  nativeBuildInputs = kernel.moduleBuildDependencies;
+
+  preBuild = ''
+    sed -e "s@/lib/modules/\$(.*)@${kernel.dev}/lib/modules/${kernel.modDirVersion}@" -i Makefile
+  '';
```
Mic92, line 26:
```
  buildFlags = [
    "KERNELDIR=${kernel.dev}/lib/modules/${kernel.modDirVersion}/build"
  ];
```

system76/pkgs/system76-acpi-dkms/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  hardeningDisable = [ "pic" ];
+  dontStrip = true;
+  dontPatchELF = true;
+
+  kernel = kernel.dev;
+  nativeBuildInputs = kernel.moduleBuildDependencies;
+
+  preBuild = ''
+    sed -e "s@/lib/modules/\$(.*)@${kernel.dev}/lib/modules/${kernel.modDirVersion}@" -i Makefile
+  '';
```
Mic92, line 26:
```suggestion
  buildFlags = [
    "KERNELDIR=${kernel.dev}/lib/modules/${kernel.modDirVersion}/build"
  ];
```

system76/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  imports = [ ../common/pc ];
+
+  options.hardware.system76 = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable support for System76 hardware.
+      '';
+    };
```
Mic92, line 28: This option seems redundant. Either you include the profile or not.

flake.nix
```diff
@@ -55,6 +55,8 @@
       pcengines-apu = import ./pcengines/apu;
       raspberry-pi-2 = import ./raspberry-pi/2;
       samsung-np900x3c = import ./samsung/np900x3c;
+      system76-darp6 = import ./system76/darp6;
+      system76-generic = import ./system76/generic;
```
Mic92, line 59: We have this convention in nixos-hardware:

```suggestion
      system76 = import ./system76;
      system76-darp6 = import ./system76/darp6;
```

system76/darp6/default.nix
```diff
@@ -0,0 +1,26 @@
...
+{ lib, ... }:
+{
+  imports = [
+    ../.
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+  ];
+
+  hardware.system76 = {
+    enable = lib.mkDefault true;
```
Mic92, line 14: Can be removed.

system76/darp6/default.nix
```diff
@@ -0,0 +1,26 @@
...
+
+    # darp6 needs system76-acpi-dkms, not system76-dkms:
+    #
+    # [1] https://github.com/pop-os/system76-dkms/issues/39
+    # jackpot51> system76-acpi-dkms is the correct driver to use on the darp6
+    #
+    # system76-io-dkms also appears to be loaded on darp6 with Pop!_OS, and
+    # system76-dkms does not, and in fact refuses to load.
+    drivers.system76-acpi.enable = lib.mkDefault true;
+    drivers.system76-io.enable = lib.mkDefault true;
```
Mic92, line 24: This module option seems to be true for all modules. Just remove the option and include the driver by default.

system76/default.nix
```diff
@@ -0,0 +1,72 @@
...
+
+    drivers.system76-io.enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to install the system76-io-dkms kernel module.
+      '';
+    };
+  };
```
Mic92, line 53: Please remove this option if it is true for all modules.

system76/default.nix
khumba: Good point, the packages belong over there.  PR created: https://github.com/NixOS/nixpkgs/pull/96514

system76/pkgs/system76-io-dkms/default.nix
khumba: Done (applied in the Nixpkgs PR).  I had to use `KERNEL_DIR` as that's what the Makefile uses.

flake.nix
khumba: I see.  I'm going to simplify this down to just one `system76/default.nix` module, with no options and just including all of the modules.  From what I can tell it doesn't cause problems to have all three modules enabled, and the drivers themselves are small (~160KiB total).

khumba:
Okay, I've simplified this right down and updated it to use the modules being added to Nixpkgs.

Mic92:
CI fails because the channel that contains the new dkms module is not yet in.
I think we can also backport the drivers to 20.09.

Mic92:
Ping me on the next nixpkgs bump that have the kernel modules: https://status.nixos.org/

shlevy:
@Mic92 Modules backported as part of https://github.com/NixOS/nixpkgs/commit/1991ea85b15ac3814d90b47365f1f383f6efd04d

shlevy:
@khumba Sorry, I didn't know about this PR. I just added a system76 module in NixOS, not sure how much should be here vs there https://github.com/NixOS/nixpkgs/commit/4b1850bad3c8cba93ffb12041da1a530cbe03c6d

Mic92:
@shlevy in future we actually move away from adding module options for each module and have profiles that are imported. This is what the nixos hardware rfc is about.

shlevy:
@Mic92 Which RFC? I see one about merging this repo with nixpkgs still in discussion, does that also include guidance about creating modules?

Sorry, I don't quite understand "move away from adding module options for each module". Is the idea that instead of `services.foo.enable = true` we'd have `imports [ services/foo.nix ]`?

khumba:
Hello,

@shlevy, thanks for packaging up the firmware daemon, that was next on my list to do and I'm glad it's already done.

To address your question from [here](https://github.com/NixOS/nixpkgs/pull/96514#issuecomment-697032485) about which modules are needed, after reading more code, I can answer this a bit better now than when I first put these PRs up.  System76 ships a [system76-driver](https://github.com/pop-os/system76-driver) utility with Pop!_OS.  That package has a [hard dependency](https://github.com/pop-os/system76-driver/blob/f3328c0dcb5f9b8d290f7c7b88e3d70bd0dfc857/debian/control) on the three kernel modules, so I assume that they are always fine to have them present and attempt loading.  As I mentioned, this works fine for darp6, even though the `system76` model isn't needed and won't load.

In addition to pulling in the modules, system76-driver has two other main functions.  It provides a (Debian post-install) script that runs `system76-driver-cli` to detect the hardware model and install config files as needed (e.g. hwdb.d, modprobe.d), and it provides a daemon that also provides (other) hardware tweaks as necessary.  It would be good to package up this daemon alongside system76-firmware for any hardware that needs this (darp6 doesn't and again it doesn't hurt to have it running).  The former fixup files are more annoying though, because depending on your [model](https://github.com/pop-os/system76-driver/blob/f3328c0dcb5f9b8d290f7c7b88e3d70bd0dfc857/system76driver/products.py#L155) certain [files](https://github.com/pop-os/system76-driver/blob/f3328c0dcb5f9b8d290f7c7b88e3d70bd0dfc857/system76driver/actions.py#L812) need to be created, and e.g. in the case of darp6 linked, that involves pulling actual hardware IDs.  So I've gone and added some options to this PR again to produce these config files, because unfortunately we can't rely on system76-driver-cli to do so.

Of course this means that means that a "generic" system76 profile isn't going to provide great support for every model; we at least need to know what the model is to determine the proper fixups to apply, or whether to ask for more info/options.

> @shlevy in future we actually move away from adding module options for each module and have profiles that are imported. This is what the nixos hardware rfc is about. [...]

> [...] Sorry, I don't quite understand "move away from adding module options for each module". Is the idea that instead of `services.foo.enable = true` we'd have `imports = [ services/foo.nix ]`?

I think this is a bigger question than this PR, and I'm happy to leave an `enable` option off of the system76 modules here because it's not strictly required.

Just personally, I'm not sure imports are a suitable replacement for enable options.  It's more convenient for the simple use case, but there's no support for conditionals or control from outside modules.  I can't have a generic config that includes the darp6 module if some other option is set, and I can't `mkForce` it off if I need to.  Reducing unnecessary imports or doing include-what-you-use would be good though.  Anyway I'm sure there's already been discussion about this somewhere.

I'm wondering whether we should continue here, or if I should port the darp6-specific fixups to Nixpkgs.  I'm not opposed to having this in Nixpkgs since it's tied to the packages there.

Thanks folks.

Mic92:
> @Mic92 Which RFC? I see one about merging this repo with nixpkgs still in discussion, does that also include guidance about creating modules?
> 
> Sorry, I don't quite understand "move away from adding module options for each module". Is the idea that instead of `services.foo.enable = true` we'd have `imports = [ services/foo.nix ]`?

rfc 70: https://github.com/NixOS/rfcs/pull/70 Sorry I never managed to put the minutes into the rfc repo: https://pad.petabyte.dev/PAHaaUmxS_-6_GkT8pwFFg#
but in a nut-shell in the future we want to have profiles that work similar to nixos-hardware but are part of nixpkgs itself. Since they won't have module that are imported into the nixos modules by default they also won't slow down evaluation. I will continue working on this next month, right now my I am busy with a paper deadline.

shlevy:
@khumba Since Linux 5.5+ has system76_acpi in-tree, I set the NIxOS module to exclude that one on newer kernels. But yes, I'm running now on a system booted with system76 and system76_io added to kernelModules on my darp6, and indeed system76 fails to load but no issues came from it.

khumba:
> @khumba Since Linux 5.5+ has system76_acpi in-tree, I set the NIxOS module to exclude that one on newer kernels.

That's cool, I wasn't aware it had been upstreamed.

Also I just noticed that system76-acpi-1.0.2 was released about a month ago.  Maybe we should bump the derivation from 1.0.1.  Not sure if it would be worth overriding the version shipped with the kernel in this case (on master at least), after some manual testing?

> But yes, I'm running now on a system booted with system76 and system76_io added to kernelModules on my darp6, and indeed system76 fails to load but no issues came from it.

Great to hear.

shlevy:
> Also I just noticed that system76-acpi-1.0.2 was released about a month ago. Maybe we should bump the derivation from 1.0.1

Yeah, definitely should be good to bump on master. Depending on if it's a bugfix or not we can backport

> Not sure if it would be worth overriding the version shipped with the kernel in this case (on master at least), after some manual testing?

Typically in my experience once the module is upstreamed that becomes the main source of development/releases, with the previous repo only sticking around for older kernels. Indeed, the README on the repo says

> This provides the system76_acpi in-tree driver for systems missing it.

But one of the two code-touching PRs is not yet upstream on Linus's tree. I've reached out to them at https://github.com/pop-os/system76-acpi-dkms/pull/3#issuecomment-702654946 for guidance.

Mic92:
@shlevy can you test this module? Can we merge it?

shlevy:
Does someone need it? As I said above, we now have the modules in-tree in the kernel, unless there's siginficant additional value IME it's much more maintainable to not carry external modules.

khumba:
I'll probably keep running the darp6 audio patch personally, but audio still works fine for me if it's not applied.  I added it for completeness and can't say what the patch actually affects.  I am fine with not merging this; other System76 devices may come along with config settings that are more essential, and we can deal with them then.

Mic92:
It also imports other profiles like laptop settings etc, so it has a value.
I think we also should have module even if they don't do much because it shows
that those machines has been tested with NixOS.

PR #191, opened 2020-09-23
mnt/reform: init

ehmry:
Add support for the MNT Reform laptop with the initial i.MX8 CPU board - https://mntre.com/

I've used this configuration to bootstrap NixOS from the Debian SD card that comes with this laptop, but I am still working on an SD installer image.


mnt/reform2-nitrogen8m/uboot/default.nix
```diff
@@ -0,0 +1,16 @@
+{ buildUBoot, fetchgit }:
+
+buildUBoot rec {
+  pname = "uboot-reform2-imx8mq";
```
Mic92, line 4: Please move the packages to nixpkgs, we cannot test/build in nixos-hardware. The module itself can remain here. If you post the PR url, I will have a look.

Mic92: Depending on your motivation you can also add support here: https://github.com/Mic92/nixos-aarch64-images
This allows people with aarch64 to remaster our official aarch64 image with the custom u-boot.

Mic92: cc @samueldr

mnt/reform2-nitrogen8m/kernel.nix
```diff
@@ -0,0 +1,92 @@
...
+        extraConfig = ''
+          DRM_IMX_DCSS m
+        '';
+      }
+    ];
+  in lib.lists.unique (kernelPatches ++ reformPatches ++ [{
+    name = "MNT-Reform-imx8mq-config";
+    patch = null;
+    extraConfig = builtins.readFile ./kernel-config;
+  }]);
```
Mic92, line 80: @samueldr what should we do with the kernel?

samueldr: This is still an open question about *everything* kernel related and "those kind" of platforms, where patches or outright complete kernel trees are needed.

I think we need to "sit down" and discuss that with the relevant key people: can we provide kernel builds in a useful manner, without making them part of Nixpkgs outright?

I'm thinking about a strategy similar to Mobile NixOS' own for reducing churn in Hydra, instead of tracking Nixpkgs, track channel updates. It means that it lags behind slightly, but doesn't needlessly queue builds... (Though that last part is not really an issue for kernels, few changes trigger rebuilds.)

I guess Flakes would help in this current situation, but adding a strict dependency on flakes is not great either.

Though one of the main important things about that is that managing non-mainline kernels through Nixpkgs is bound to become untenable. I already want to move the Raspberry Pi foundation kernels outside whenever possible, in addition to the few misc. probably-unused kernels still present.

Ugh, sorry to drop all that in a PR review request :/.

mnt/reform2-nitrogen8m/uboot/default.nix
samueldr: As a complete reversal to my opinions about the kernel, U-Boot is "self-sustaining", since it's part of U-Boot mainline. Adding a build for a particular device is basically free.

We might want to stray away from including extraConfig. Sure, that means it'll have the default 3s boot delay, but that's how we build our U-Boots for now: stock as mainline does.

It's also quite cheap to build a custom U-Boot if needed by the end-user. That means it's not a huge deal if they need to customize it.

Except for: see my other comments about the patches.

samueldr: I'm not sure we want to introduce a non-mainline U-Boot into Nixpkgs. [I went as far as making myself honest](https://github.com/NixOS/nixpkgs/pull/98607), by removing the last non-mainline U-Boot from Nixpkgs.

A bit like for kernels, the main reasoning behind isn't about configuration or device-specificity, but about *maintainership*. Adding U-Boot defconfigs for a mainline board is trivial, and pretty much free.

samueldr:
Not pertinent to *reviewing* the PR per se, but more about being a good citizen in the "NixOS on ARM" stuff.

* * *

I see you're patching environment variables here:

https://github.com/NixOS/nixos-hardware/blob/37d1eff124a54e8e78261acd56d3d09ea9a337dd/mnt/reform2-nitrogen8m/u-boot/env_vars.patch#L9-L30

Am I to understand that this ends up making that ~2018 era U-Boot use the [Generic Distro Configuration Concept](https://gitlab.denx.de/u-boot/u-boot/-/blob/master/doc/README.distro)?

Well, [I figure it must](https://github.com/NixOS/nixos-hardware/pull/191/files#diff-7e60ba346d9bcada1a8e6ef1b780f8daR13-R17). Too bad about the lack of graphics in U-Boot :/.

mnt/reform2-nitrogen8m/u-boot/default.nix
```diff
@@ -0,0 +1,17 @@
+{ buildUBoot, fetchgit }:
```
samueldr, line 1: You seem to have accidentally doubled-up on U-Boots, with a `uboot` and a `u-boot` directory.

 * [ ] Delete the unused `u-boot` directory.

mnt/reform2-nitrogen8m/kernel.nix
samueldr: Oh, additionally `linux_5_7` is bound to disappear. It's already marked EOL at kernel.org.

mnt/reform2-nitrogen8m/kernel-config
```diff
@@ -0,0 +1,2180 @@
+6LOWPAN n
```
samueldr, line 1: Is this original work?

If it isn't, where can we get the updated recommended options to enable?

mnt/reform2-nitrogen8m/default.nix
```diff
@@ -0,0 +1,54 @@
...
+{
+  boot = {
+    initrd = {
+      availableKernelModules = [ "nvme" "usbhid" ];
+      kernelModules = [ "nwl-dsi" "imx-dcss" ];
+    };
+    extraModprobeConfig = "options imx-dcss dcss_use_hdmi=0";
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_reformNitrogen8m_latest;
+    kernelParams =
+      [ "console=ttymxc0,115200" "console=tty1" "cma=512M" "pci=nomsi" ];
```
samueldr, line 12: I wonder if, just as in [#97883](https://github.com/NixOS/nixpkgs/pull/97883), we shouldn't set `cma=` on the command-line.

 * https://github.com/NixOS/nixpkgs/pull/97883/commits/3a13b44e862c0a6b86de975aa1a0c9dea1e4adc9

The custom kernel config is already setting it. So I don't think it's needed here.

* * *

Though, as our default universal kernel config sets it as 32M, an hypothetical upcoming time where this computer uses mainline with defconfig would need to see this set in some ways.

I would tend to think that it would be better off as a default `dtbo` that's shipped with the proper configuration for this device, in a future update. I mean, I don't even know myself how to do it exactly, so I wouldn't ask someone to do the dirty work of figuring this out. You're just caught in the cross-fire between my neurons.

mnt/reform2-nitrogen8m/kernel-config
```diff
@@ -0,0 +1,2180 @@
...
+CHR_DEV_ST n
+CICADA_PHY n
+CIFS n
+CLEANCACHE n
+CLK_IMX8MM n
+CLK_IMX8MN n
+CLK_IMX8MP n
+CLK_QORIQ n
+CLOCK_THERMAL y
+CMA_SIZE_MBYTES 512
```
samueldr, line 189: (ref: this is used when `cma=` is not present on the command-line, and no cma nodes exist in the device tree.)

mnt/reform2-nitrogen8m/kernel.nix
Mic92: I actually would not those packages to hydra and the official binary cache but have github repositories in nix-community. We have a diverse community and only a few people with access to the nixos core infrastructure. In nix community there is a 1TB cachix cache, which would allow to upload custom kernel builds and the whole process is a lot less bureaucratic.

mnt/reform2-nitrogen8m/uboot/default.nix
ehmry: FWIW the U-Boot maintainer knows about these patches and wants this board fixed in U-Boot upstream.

mnt/reform2-nitrogen8m/kernel.nix
ehmry: > Ugh, sorry to drop all that in a PR review request :/.

Not a problem, the hardware hasn't shipped en masse, so we have time to do this right.

mnt/reform2-nitrogen8m/kernel-config
ehmry: This is semi-original, I took the config of the [stock MNT kernel](https://source.mntmn.com/MNT/reform-system-image/src/branch/reform2-redo/reform2-imx8mq/template-kernel/kernel-config), diffconfig'ed and trimed it a bit. Some stuff is disabled to reduce the compile time. MNT uses their patches against a 5.7 kernel, so I chose the same.

mnt/reform2-nitrogen8m/uboot/default.nix
ehmry: > Depending on your motivation you can also add support here: https://github.com/Mic92/nixos-aarch64-images This allows people with aarch64 to remaster our official aarch64 image with the custom u-boot.

Ok, I can do that when I have a working SD image.

mnt/reform2-nitrogen8m/kernel.nix
samueldr: There is a level of trust around "just using a cachix repo", that I highly dislike. That's the main reason I don't publish anything for the Pinebook Pro WIP stuff, I don't want to have people either not realise they rely on an implicit trust on me, or have them trust what I do.

If the Nix community builder is transparent enough for that, it'd help, we'd probably be able to handle that for all NixOS on ARM things. I really want to move all the things in a centralized location with builds. @Mic92 care to open an issue and start the discussion about that at the right place (and tagging me in)?

mnt/reform2-nitrogen8m/uboot/default.nix
Mic92: @ehmry you basically only need to know the offsets where the chip expects uboot to be: https://github.com/Mic92/nixos-aarch64-images/blob/main/images/rockchip.nix#L19

samueldr: @Mic92, be careful, that's not necessarily a given that an SoC will boot like Allwinners and Rockchips, by looking at a particular offset on the storage.

Though, here it looks like it may be, *when configured appropriately*.

 * https://community.nxp.com/t5/i-MX-Processors-Knowledge-Base/i-MX8-Boot-process-and-creating-a-bootable-image/ta-p/1101253

ehmry:
I've rebased the initial commit and pushed a second commit that adds and installer module. Not sure the best way to make it discoverable.

ehmry:
FWIW I've been building the SD image by adding a nixpkgs input to the flake and a package

```diff
+    packages.aarch64-linux.reform2-nitrogen8m-sd-installer = let
+      installer = nixpkgs.lib.nixosSystem {
+        system = "aarch64-linux";
+        modules = [ ./mnt/reform2-nitrogen8m/installer.nix ];
+      };
+    in installer.config.system.build.sdImage;
+
```

mnt/reform2-nitrogen8m/kernel.nix
Mic92: @samueldr I just realized that we don't have aarc64 builds in nix-community. Apart from that adding declarative jobs is supported https://github.com/nix-community/infra/blob/0374f082455b9cce0a369bdc53cc4fa2ca9c87a0/build01/hydra-declarative-projects.nix#L5 So the only other alternative I see is to have a jobset on the official hydra. There might be also aarch64 support on travis. I don't know if this is enough to build kernel, but if it is than this is actually the better solution because than everyone can customize their built.

ehmry:
I made a seprate repository here: https://github.com/nix-community/hardware-mnt-reform

This laptop needs a pinned nixpkgs to build its kernel, so I don't think its appropriate to keep here.

PR #194, opened 2020-10-07
Add support for the Lenovo Thinkpad L14 (AMD) laptop

blitz:
This PR adds support for the Lenovo Thinkpad L14 Gen 1 laptop with AMD Ryzen CPU (Renoir).

I'm not 100% sure where to put the module. If some one has a better idea than `l14-amd` I'm all ears. Should it be `l14/amd` instead?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,8 @@
       lenovo-thinkpad-e470 = import ./lenovo/thinkpad/e470;
       lenovo-thinkpad-e495 = import ./lenovo/thinkpad/e495;
       lenovo-thinkpad-l13 = import ./lenovo/thinkpad/l13;
+      lenovo-thinkpad-l14-intel = import ./lenovo/thinkpad/l14/intel;
+      lenovo-thinkpad-l14-amd = import ./lenovo/thinkpad/l14/amd;
       lenovo-thinkpad-p53 = import ./lenovo/thinkpad/p53;
       lenovo-thinkpad-t410 = import ./lenovo/thinkpad/t410;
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -30,6 +30,7 @@
       lenovo-ideapad-z510 = import ./lenovo/ideapad/z510;
       lenovo-thinkpad-e495 = import ./lenovo/thinkpad/e495;
       lenovo-thinkpad-l13 = import ./lenovo/thinkpad/l13;
+      lenovo-thinkpad-l14-amd = import ./lenovo/thinkpad/l14-amd;
```
Mic92, line 33: As you said we should move this to:

```suggestion
      lenovo-thinkpad-l14-amd = import ./lenovo/thinkpad/l14/amd;
```

lenovo/thinkpad/l14-amd/default.nix
```diff
@@ -0,0 +1,24 @@
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
+    ../.
+    ../../../common/cpu/amd
```
Mic92, line 6: Could their be a generic l14 module that gets extended by the more specific amd profile?

flake.nix
blitz: Done.

lenovo/thinkpad/l14-amd/default.nix
blitz: So I've created `l14/amd` and `l14/intel`. I've populated the `l14/intel` profile with the same content as the `l13` profile, because its basically the same laptop in a smaller chassis, _but_ I couldn't test it.

PR #199, opened 2020-10-25
t14s: fix minimum kernel version

dadada:
Adds special settings for AMD models. linuxPackages 5.8 is currently only available in 20.09, but is required for backlight and microphone to work. The graphics chipset should work with older kernels (reportedly also 5.4), but this is untested.

There is some manual work required to get suspend and the fingerprint sensor to work, so I left a comment, in the hopes that it may be found. It's probably infeasible to put a warning, since the firmware version and UEFI settings are hard to detect from nix.

### DIFF
--- a/lenovo/thinkpad/t14s/amd/gen1/default.nix
+++ b/lenovo/thinkpad/t14s/amd/gen1/default.nix
@@ -1,0 +1,8 @@
+
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
+    ../.
+  ];
+}
--- a/lenovo/thinkpad/t14s/default.nix
+++ b/lenovo/thinkpad/t14s/default.nix
@@ -7,6 +7,11 @@
     ../../../common/pc/laptop/acpi_call.nix
   ];
 
+  # For suspending to RAM to work, set Config -> Power -> Sleep State to "Linux" in EFI.
+  # See https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_X1_Carbon_(Gen_6)#Suspend_issues
+
+  # Fingerprint sensor requires a firmware-update to work.
+
   # Force use of the thinkpad_acpi driver for backlight control.
   # This allows the backlight save/load systemd service to work.
   boot.kernelParams = [ "acpi_backlight=native" ];
--- a/lenovo/thinkpad/t14s/default.nix
+++ b/lenovo/thinkpad/t14s/default.nix
@@ -3,7 +3,6 @@
 {
   imports = [
     ../.
-    ../../../common/cpu/amd
     ../../../common/pc/laptop/acpi_call.nix
   ];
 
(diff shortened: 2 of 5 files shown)


lenovo/thinkpad/t14s/default.nix
```diff
@@ -7,6 +7,11 @@
     ../../../common/pc/laptop/acpi_call.nix
```
Mic92, line 7: This profile already imports `amd`. Should that not be moved to the amd profile?

dadada: I'm not sure if I understand you correctly. Do you mean we should move line 6 to `lenovo/thinkpad/t14s/amd/default.nix`?

lenovo/thinkpad/t14s/amd/gen1/default.nix
```diff
@@ -0,0 +1,8 @@
+
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
```
Mic92, line 5: Maybe we should add this once, we have gen2 and it becomes necessary that we need different configuration?

Mic92: Is there already a second generation? What if Lenovo never releases one?

lenovo/thinkpad/t14s/default.nix
Mic92: Given you changes, I think we are on the same page.

lenovo/thinkpad/t14s/amd/gen1/default.nix
Mic92: What I mean to say, I would only introduce `lenovo/thinkpad/t14s/amd/default.nix` for now and add `gen1` later if newer models come with different needs.

dadada: I was thinking of the case where you have <nixos-hardware/lenovo/thinkpad/t14s/amd> `imports` and then if we have to do the split into `gen1` and `gen2`, the import path breaks, because the specific parts for `gen1` go into `./gen1`. Or is there a way to later notify the user that they maybe want to include the more specific file instead?

Mic92: I think we could add a warning in this case, by defining some options inside the `gen1` and `gen2` modules.

dadada: If `/etc/nixos/configuration.nix` contained `import = [ <nixos-hardware/lenovo/thinkpad/t14s/amd> ];`, then `./gen1` and `./gen2/` would not be evaluated and the warning will never be shown.

dadada: Also Lenovo lists the laptop as `Gen1` everywhere, so that this heavily implies that there may be a Gen2 at some point along with a new T13 and T15.

PR #200, opened 2020-10-26
t14s: removed amd import for general profile

sevenfourk:
t14s are coming in intel and amd, i guess there should split like in here:

https://github.com/NixOS/nixos-hardware/tree/master/lenovo/thinkpad/l14

### DIFF
--- a/lenovo/thinkpad/t14s/default.nix
+++ b/lenovo/thinkpad/t14s/default.nix
@@ -3,7 +3,6 @@
 {
   imports = [
     ../.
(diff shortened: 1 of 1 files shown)


sevenfourk:
disclosure: I don't own a hardware or run nixOS yet, but reading/researching around.

Mic92:
Thanks for the PR, we will also fix it here: https://github.com/NixOS/nixos-hardware/pull/199/files

sevenfourk:
@Mic92 got it, great!

PR #202, opened 2020-10-31
macbook: added option for swapped cmd and opt

sevenfourk:
not 100% sure it should be default, as technically cmd is a super key, but being accustomed to PC keyboard layout I think it might easily be default.

ref: https://wiki.archlinux.org/index.php/Apple_Keyboard#Switching_Cmd_and_Alt/AltGr

### DIFF
--- a/apple/default.nix
+++ b/apple/default.nix
@@ -2,6 +2,7 @@
 
 {
   boot.kernelParams = [
+    "hid_apple.swap_opt_cmd=1"
     "hid_apple.iso_layout=0"
   ];
 
--- a/apple/default.nix
+++ b/apple/default.nix
@@ -2,7 +2,6 @@
 
 {
   boot.kernelParams = [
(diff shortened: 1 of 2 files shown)


apple/default.nix
```diff
@@ -2,6 +2,7 @@
 
 {
   boot.kernelParams = [
+    "hid_apple.swap_opt_cmd=1"
```
Mic92, line 5: Given that it is not easy to disable those options again, should we rather have a `Readme.md` in `./apple` that just document this option?

```markdown
## Switching Cmd and Alt/AltGr

This will switch the left Alt and Cmd key as well as the right Alt/AltGr and Cmd key. 

boot.kernelParams = [
  "hid_apple.swap_opt_cmd=1"
];
```

sevenfourk: yeah, let's do that. updated.

PR #206, opened 2020-11-04
lenovo/thinkpad/t14: created symlink to t14s/

### LINKED ISSUE #205: Lenovo Thinkpad T14 AMD Gen 1
Do you think we need a separate profile or we can use this one:

https://github.com/NixOS/nixos-hardware/tree/master/lenovo/thinkpad/t14s/amd/gen1 ?

sevenfourk:
should close #205


lenovo/thinkpad/t14
```diff
@@ -0,0 +1 @@
+t14s/
```
Mic92, line 1: Can you also add this to README.md and flake.nix?

sevenfourk: added, can you please check.

PR #214, opened 2020-11-26
apple/macbook-pro/12-1: PM: set cpuFreqGovernor and power Up/Down commands

### LINKED ISSUE #210: apple/macbook-pro/12-1: wireless don't get reconnected after resume
not sure what to add here, but wanted to at least file an issue as a todo item.

hardware: 03:00.0 Network controller: Broadcom Inc. and subsidiaries BCM43602 802.11ac Wireless LAN SoC (rev 01)

### LINKED ISSUE #211: apple/macbook-pro/12-1: won't resume after hibernation
description:
``systemctl hibernate`` -- works fine, though when resuming it fails to get into the system, and the system reboots by itself.

kernel:
``Linux mac 5.4.78 #1-NixOS SMP Wed Nov 18 18:20:34 UTC 2020 x86_64 GNU/Linux``

using:

```nix
  systemd.sleep.extraConfig = ''
    HibernateMode=shutdown
  '';
```

ram/swap:

```console
❯ free -m
              total        used        free      shared  buff/cache   available
Mem:           7870        1675        4452         507        1742        5397
Swap:          8191           0        8191
```

using luks encryption:

```console
❯ lsblk
NAME          MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINT
sda             8:0    0   113G  0 disk
├─sda1          8:1    0   511M  0 part  /boot
└─sda2          8:2    0 112.5G  0 part
  └─crypted   254:0    0 112.5G  0 crypt
    ├─vg-swap 254:1    0     8G  0 lvm   [SWAP]
    └─vg-root 254:2    0 104.5G  0 lvm   /
```

sevenfourk:
fixes #211.
fixes #210.

Added default CPU governor "conservative" to enable gradually increasing/decreasing CPU freq, rather than using "powersave", which would keep CPU freq at 0.8Ghz.

 Added power Up/Down PM commands which will unload and load brcmfmac on state change. This is not elegant way, to say the least.

### DIFF
--- a/apple/macbook-pro/12-1/README.md
+++ b/apple/macbook-pro/12-1/README.md
@@ -1,0 +1,14 @@
+# MacBook Pro 12,1
+
+## Wireless won't get reconnected after resume/hibernate
+
+As an example similar code could be used to restart wireless interface:
+
+```nix
+powerManagement.powerUpCommands = ''
+  ${pkgs.systemd}/bin/systemctl restart wpa_supplicant.service
+'';
+};
+```
+
+You can apply this to your network management software of choice.
--- a/apple/macbook-pro/12-1/README.md
+++ b/apple/macbook-pro/12-1/README.md
@@ -2,7 +2,8 @@
 
 ## Wireless won't get reconnected after resume/hibernate
 
-As an example similar code could be used to restart wireless interface:
+The wifi driver is unloaded before suspend/hibernate to workaround driver issues.
+This means it might be required to restart your wifi deamon i.e. wpa_supplicant:
 
 ```nix
 powerManagement.powerUpCommands = ''
--- a/apple/macbook-pro/12-1/default.nix
+++ b/apple/macbook-pro/12-1/default.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ lib, pkgs, ... }:
 
 {
   imports = [
@@ -6,6 +6,17 @@
     ../../../common/pc/laptop/ssd
     <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix>
   ];
+
+  powerManagement = {
+    # enable gradually increasing/decreasing CPU frequency, rather than using
+    # "powersave", which would keep CPU frequency at 0.8GHz.
+    cpuFreqGovernor = lib.mkDefault "conservative";
+
+    # brcmfmac being loaded during hibernation would not let a successful resume
+    # https://bugzilla.kernel.org/show_bug.cgi?id=101681#c116
+    powerUpCommands = lib.mkBefore "${pkgs.kmod}/bin/modprobe brcmfmac";
+    powerDownCommands = lib.mkBefore "${pkgs.kmod}/bin/rmmod brcmfmac";
+  };
 
   # USB subsystem wakes up MBP right after suspend unless we disable it.
   services.udev.extraRules = lib.mkDefault ''


apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,20 @@
     <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix>
   ];
 
+  powerManagement = {
+    cpuFreqGovernor = lib.mkDefault "conservative";
```
Mic92, line 11:
```suggestion
    # enable gradually increasing/decreasing CPU freq, rather than using "powersave", which would keep CPU freq at 0.8Ghz.
    cpuFreqGovernor = lib.mkDefault "conservative";
```

apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,20 @@
...
+
+    powerUpCommands = ''
+      ${pkgs.kmod}/bin/modprobe brcmfmac
+      ${pkgs.coreutils}/bin/sleep 5
+      ${pkgs.systemd}/bin/systemctl restart wpa_supplicant.service
```
Mic92, line 16: We need something better here: There are two different wpa supplicant modules that have different service names. Than there is network manager with its own wpa_supplicant and iwd, which independent from network manager. What does actually happen if the wifi interface disappears/appears? Does wpa supplicant not pick it up again?

apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,21 @@
     <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix>
   ];
 
+  powerManagement = {
+    # enable gradually increasing/decreasing CPU freq, rather than using "powersave", which would keep CPU freq at 0.8Ghz.
```
Mic92, line 11: Can you explain how this behavior is better than the current default?

apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,20 @@
     <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix>
```
sevenfourk, line 7: wanted to discuss also this module, I've tried unloading and brcmfmac and loading b43, but it did not work for me. i suspect that it lingers here from older commits.

Mic92: cc @teh who added this.

apple/macbook-pro/12-1/default.nix
sevenfourk: for me when using default governor "powersave" did not scale CPU freq at all, and I had extremely low performance, almost unusable in gui applications.

apple/macbook-pro/12-1/default.nix
Mic92: All this module does is `hardware.enableRedistributableFirmware = true;`. Is there any meaningful message in dmesg when doing the modprobe?

Mic92: This is where b43 lives: https://github.com/torvalds/linux/tree/master/drivers/net/wireless/broadcom/b43

apple/macbook-pro/12-1/default.nix
Mic92: Thanks for explanation.

apple/macbook-pro/12-1/default.nix
sevenfourk: ah, indeed. so it's not using b43.

sevenfourk:
```console
[  175.994580] cfg80211: Loading compiled-in X.509 certificates for regulatory database
[  175.994858] cfg80211: Loaded X.509 cert 'sforshee: 00b28ddf47aef9cea7'
[  175.994895] platform regulatory.0: Direct firmware load for regulatory.db failed with error -2
[  175.994897] cfg80211: failed to load regulatory.db
[  176.055245] Broadcom 43xx driver loaded [ Features: PNL ]
```

but since this driver is not used, i think it does not matter much.

apple/macbook-pro/12-1/default.nix
sevenfourk: found what happens: when unloading a module, wpa_supplicant simply stops.

```cosole
Nov 27 13:57:31 mac sudo[10347]: sevenfourk : TTY=pts/2 ; PWD=/home/sevenfourk/Sources/GitHub/nixos-hardware ; USER=root ; COMMAND=/run/current-system/sw/bin/m
odprobe -r brcmfmac
Nov 27 13:57:31 mac sudo[10347]: pam_unix(sudo:session): session opened for user root by (uid=0)
Nov 27 13:57:31 mac kernel: usbcore: deregistering interface driver brcmfmac
Nov 27 13:57:31 mac kernel: ieee80211 phy1: brcmf_fil_cmd_data: bus is down. we have nothing to do.
Nov 27 13:57:31 mac kernel: ieee80211 phy1: brcmf_cfg80211_get_tx_power: error (-5)
Nov 27 13:57:31 mac dhcpcd[1524]: wlp3s0: fe80::52c7:bfff:fe0d:2c92 is unreachable
Nov 27 13:57:31 mac wpa_supplicant[10086]: wlp3s0: CTRL-EVENT-DISCONNECTED bssid=50:c7:bf:0d:2c:91 reason=1 locally_generated=1
Nov 27 13:57:31 mac wpa_supplicant[10086]: dbus: wpa_dbus_property_changed: no property SessionLength in object /fi/w1/wpa_supplicant1/Interfaces/0
Nov 27 13:57:31 mac dhcpcd[1524]: arp_read: wlp3s0: Network is down
Nov 27 13:57:31 mac dhcpcd[1524]: wlp3s0: carrier lost
Nov 27 13:57:31 mac systemd[1]: Stopping WPA Supplicant...
```

i always used the simplest method: only wpa_supplicant, do you think i should better use network manager or so?

apple/macbook-pro/12-1/default.nix
sevenfourk: do you think we might put load/unload of module here: <nixpkgs/nixos/modules/hardware/network/broadcom-43xx.nix> ?

apple/macbook-pro/12-1/default.nix
Mic92: I use iwd because it's interface is nicer than wpa_supplicant. It works both with and without network manager. Maybe we should mention in the README that one needs to restart wpa_supplicant somehow if this is the case?

apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,14 @@
...
   ];
 
+  powerManagement = {
+    # enable gradually increasing/decreasing CPU freq, rather than using "powersave", which would keep CPU freq at 0.8Ghz.
+    cpuFreqGovernor = lib.mkDefault "conservative";
+
+    powerUpCommands = ''${pkgs.kmod}/bin/modprobe brcmfmac'';
```
Mic92, line 14: Is this both required for suspend and hibernate?

```      
Commands executed when the machine powers down. That is, they're executed both when the system shuts down and when it goes to suspend or hibernation.
````

Comment (sevenfourk) [reply apple/macbook-pro/12-1/default.nix:14]: for me it's required. if not using this config i would have to manually restart wpa_supplicant.

File: apple/macbook-pro/12-1/default.nix (hunk 2 above)
Comment (sevenfourk) [reply apple/macbook-pro/12-1/default.nix]: yes, perhaps documenting it would be the best idea, we can't restart the service, someone might not even use it.

File: apple/macbook-pro/12-1/default.nix (hunk 5 above)
Comment (Mic92) [reply apple/macbook-pro/12-1/default.nix:17]: I mean does `brcmfmac` also breaks suspend or just hibernate?

File: apple/macbook-pro/12-1/README.md [hunk 6]
```diff
@@ -0,0 +1,12 @@
+# MacBook Pro 12,1
+
+## Wireless won't get reconnected after resume/hibernate
+
+As an example similar code could be used to restart wireless interface:
+
+```nix
+powerManagement.powerUpCommands = ''
```
Mic92, line 8: @Infinisil How can we make this string appended after our own module value? We are currently doing `rmmod`. I think there was a way to do priorities.

apple/macbook-pro/12-1/default.nix
sevenfourk: ah, yeah, it does not break resume after suspend, so only required for hibernate.

apple/macbook-pro/12-1/default.nix
sevenfourk: added a README.md

apple/macbook-pro/12-1/default.nix
Mic92: Ok. So it should be only unloaded for `hypernate-sleep.target` and "hybrid-sleep.target"? https://github.com/NixOS/nixpkgs/blob/950bb1452032e8e839ddb2af0dc85350c73025a6/nixos/modules/config/power-management.nix#L92

That would require two services, but systemd's script field should keep the boiler code needed for this low.

apple/macbook-pro/12-1/README.md
Infinisil: Not sure how `rmmod` relates, but if you mean how to order different `powerUpCommand` assignments, you can use `mkBefore` or `mkAfter` for that, which make sure the value is ordered before or after the default priority.

Mic92: Thanks! This was the helper functions I was looking for.

apple/macbook-pro/12-1/default.nix
sevenfourk: yeah, only for:

* hibernate-sleep.target
* hybrid-sleep.target

you mean, we should override this block:

```nix
    systemd.services.post-resume =
      { description = "Post-Resume Actions";
        after = [ "suspend.target" "hibernate.target" "hybrid-sleep.target" ];
        script =
          ''
            /run/current-system/systemd/bin/systemctl try-restart post-resume.target
            ${cfg.resumeCommands}
            ${cfg.powerUpCommands}
          '';
        serviceConfig.Type = "oneshot";
      };
```

with only two targets?

generally, i'm not that fluent in nix, if you would provide the best solution here i would really appreciate it.

Mic92: No just create new systemd services:

```
    systemd.services.post-resume-brcmfmac = { 
        description = "Post-Resume Actions";
        after = [ "hibernate.target" "hybrid-sleep.target" ];
        script = ''
          # here goes your code
        '';
        serviceConfig.Type = "oneshot";
      };
```

sevenfourk: added.

sevenfourk: recalled another case: sometimes on lid close (sleep/suspend) brcmfmac could crash itself, so i guess the better approach would be to still use it in powerUp/Down commands of PM. unload/load driver.

if that's too hacky, let's leave it as it, and I might keep it in my local configuration for now.

apple/macbook-pro/12-1/default.nix
```diff
@@ -7,6 +7,29 @@
...
+    # "powersave", which would keep CPU frequency at 0.8GHz.
+    cpuFreqGovernor = lib.mkDefault "conservative";
+  };
+
+  # brcmfmac wireless module being loaded during hibernation would not let a
+  # successful resume: https://bugzilla.kernel.org/show_bug.cgi?id=101681#c116
+  systemd.services.pre-sleep-brcmfmac = {
+    description = "Pre-Sleep Actions";
+    wantedBy = [ "sleep.target" ];
+    before = [ "sleep.target" ];
```
Mic92, line 21: Reading the manpage: https://www.freedesktop.org/software/systemd/man/systemd.special.html#
it should be those targets. but please test this part:

```suggestion
    wantedBy = [ "hibernate.target" "hybrid-sleep.target" ];
    before = [ "hibernate.target" "hybrid-sleep.target" ];
```

apple/macbook-pro/12-1/default.nix
Mic92: Ok. If this is the case, than we keep the original.

apple/macbook-pro/12-1/README.md
Mic92: @sevenfourk If you revert to the previous `powerManagement.powerUpCommands/powerDownCommands`
but use `mkBefore`:

```
powerManagement.powerUpCommands = lib.mkBefore ''
  ${pkgs.kmod}/bin/modprobe brcmfmac
'';
```

Than users can just follow the readme to restart their network manager daemon of choice and the ordering will make sure that their services gets added after the modprobe.

apple/macbook-pro/12-1/default.nix
sevenfourk: added lib.mkBefore for both Up/Down commands.

apple/macbook-pro/12-1/README.md
sevenfourk: thanks guys, leaving ordering for README example as is. added lib.mkBefore for module manipulation.

apple/macbook-pro/12-1/README.md
```diff
@@ -0,0 +1,14 @@
+# MacBook Pro 12,1
+
+## Wireless won't get reconnected after resume/hibernate
+
+As an example similar code could be used to restart wireless interface:
```
Mic92, line 5:
```suggestion
The wifi driver is unloaded before suspend/hibernate to workaround driver issues.
This means it might be required to restart your wifi deamon i.e. wpa_supplicant:
```

Mic92:
Thanks!

PR #220, opened 2020-12-29
Copy NVidia 9500 NVidia settings for 9550

Ericson2314:
I've been using this (but hand-rolled, in github.com/Ericson2314/nixos-configuration/) for quite some time.

### DIFF
--- a/dell/xps/15-9550/nvidia/default.nix
+++ b/dell/xps/15-9550/nvidia/default.nix
@@ -1,0 +1,15 @@
+{ lib, pkgs, ... }:
+{
+  imports = [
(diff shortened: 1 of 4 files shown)


dell/xps/15-9550/nvidia/default.nix
```diff
@@ -0,0 +1,15 @@
+{ lib, pkgs, ... }:
```
Mic92, line 1: Please also reference this file in the README table as well as flake.nix.

Ericson2314: Done.

PR #221, opened 2021-01-03
Port of linux-surface/linux-surface Kernel patches to NixOS

mexisme:
This is an update and partial rewrite of the initial port of the linux-surface patches to NixOS, and replaces #170.
This specific PR intends to focus only on the Kernel patches, not any of the other fixes (e.g. ath10k firmware for Surface Go) or support binaries (e.g. `iptsd` or `dtx`).

- Original is https://github.com/jakeday/linux-surface
- Forked to https://github.com/linux-surface/linux-surface

It defaults to kernel 5.10.2, which includes initial support for the built-in WebCam.

The support matrix from the upstream repo is here:
- https://github.com/linux-surface/linux-surface/wiki/Supported-Devices-and-Features#feature-matrix


Mic92:
ci errors look unrelated.

microsoft/surface/kernel/default.nix
```diff
@@ -0,0 +1,471 @@
...
+        kernelPatches = [
+          kernelPatches.bridge_stp_helper
+          # kernelPatches.request_key_helper
+          # kernelPatches.export_kernel_fpu_functions
+        ]
+        ++ surface_kernelPatches.linux_5_5;
+      }
+    )));
+
+    linux_5_9_2 = (with pkgs; recurseIntoAttrs (linuxPackagesFor (
```
Mic92, line 450: Pinning kernel versions seems reasonable. Do you think you can maintain/test three kernel versions?

Mic92:
I try to fix CI in https://github.com/NixOS/nixos-hardware/pull/222

microsoft/surface/kernel/default.nix
mexisme: I'm not keen to support that many kernels, but left them in for the purpose of the draft — to get some feedback in case I'm going the wrong direction:
- 20.09 doesn't seem to support 5.10, yet.
- After 5.4 the ipts driver changed; I'm working on getting `iptsd` built from @hpfr's work.

mexisme: Actually, is it acceptable for me to raise multiple smaller PRs, instead?
I normally prefer that, to simplify review, therefore thinking that I should perhaps leave this PR for the kernel patching alone, and raise new PRs for the other work, for example the `ipstd` touch-screen daemon.

Mic92: I don't mind reviewing this or smaller PRs, right now the PR is not too big for me. I would suggest however to reduce the number of kernel to what you actually test. Keeping one kernel up-to-date is tedious enough usually.

mexisme: Thanks.
I'm already planning to drop the extra kernels.
The code strikes me as a little messy and hard to follow with them in.

There will def have to be a follow-up PR to deal with some problems I've discovered, and am working through, with the nonfree linux firmware from kernel.org;  it's arguably independent of this PR, though.

I've been gathering notes to add to a README.

microsoft/surface/default.nix
```diff
@@ -0,0 +1,11 @@
+{ config, lib, pkgs, ... }:
+{
+  imports = [
+    ./kernel
+    ./firmware
+    ./hardware_configuration.nix
+  ];
+
+  # boot.kernelPackages = surface_kernelPackages.linux_5_10_2;
+  # boot.kernelPatches = surface_kernelPatches.linux_???;
```
mexisme, line 10: Remove commented code

microsoft/surface/hardware_configuration.nix
```diff
@@ -0,0 +1,19 @@
...
+    options snd_ac97_codec power_save=1
+    options iwlwifi power_save=Y
+    options iwldvm force_cam=N
+    options ath10k_core skip_otp=Y
+  '';
+
+  boot.kernelParams = [ "mem_sleep_default=deep" ];
+
+  # TLP is known to cause problems on Surface unless correctly configured.
+  # See: https://github.com/linux-surface/linux-surface/blob/master/README.md
```
mexisme, line 15: Move to README?

microsoft/surface/kernel/default.nix
```diff
@@ -0,0 +1,6 @@
+{ config, lib, pkgs, ... }:
+
+{
+  # TODO: Create a function for selecting preferred kernel:
```
mexisme, line 4: Move to TODO.org

microsoft/surface/firmware/ipts.nix
```diff
@@ -0,0 +1,13 @@
+{ stdenv, pkgs }:
+let
+  repos = (pkgs.callPackage ../repos.nix {});
+in
+stdenv.mkDerivation {
+  name = "microsoft-surface-ipts-firmware";
```
StollD, line 6: Hi, linux-surface co-maintainer here. FYI, the ipts firmware files are only needed if you are using kernel 4.19 with the old IPTS driver. With kernel 5.4+ and the new driver these are not used anymore and dont need to be installed.

Mic92: In theory older kernels in NixOS but its probably a good idea to just settle on the newer ones for this hardware.

mexisme: > Hi, linux-surface co-maintainer here. FYI, the ipts firmware files are only needed if you are using kernel 4.19 with the old IPTS driver. With kernel 5.4+ and the new driver these are not used anymore and dont need to be installed.

Thanks for that -- I hadn't realised that had changed from the docs, I'm afraid.

I'll follow-up over there, except for asking whether that includes the Surface Go and SL3 variations?
As a few notes in either the `iptsd` or `intel-precise-touch` repos imply they're different?

mexisme: > In theory older kernels in NixOS but its probably a good idea to just settle on the newer ones for this hardware.

I agree.
But, I'm also tempted wrap it with an option switch rather than remove it entirely, though?

StollD: > I'll follow-up over there, except for asking whether that includes the Surface Go and SL3 variations?
> As a few notes in either the `iptsd` or `intel-precise-touch` repos imply they're different?

The Surface Go line and the AMD version of the Surface Laptop 3 don't use IPTS. The touchscreen on the Go's is supported upstream, while the one on the SL3-AMD doesn't work at all right now.

mexisme: > > I'll follow-up over there, except for asking whether that includes the Surface Go and SL3 variations?
> > As a few notes in either the `iptsd` or `intel-precise-touch` repos imply they're different?
> 
> The Surface Go line and the AMD version of the Surface Laptop 3 don't use IPTS. The touchscreen on the Go's is supported upstream, while the one on the SL3-AMD doesn't work at all right now.
> 

Oh, grrr! I completely misunderstood that part, then!
Thanks very much!

mexisme: > Hi, linux-surface co-maintainer here. FYI, the ipts firmware files are only needed if you are using kernel 4.19 with the old IPTS driver. With kernel 5.4+ and the new driver these are not used anymore and dont need to be installed.

I've removed the derivative for the `ipts` firmware, thanks.

mexisme:
FYI: PR #107346 appears to affect my building this in 20.09; builds fine in unstable.

mexisme:
I've refactored the commit history in my branch to try to help with review.
As it stands, this PR should build the 5.10.2 kernel correctly for the MS Surface range.

There is a `TODO.org` file documenting the other work needed (and incoming) e.g:
- Fix for ath10k firmware on Surface Go
- Build +install `iptsd` user-mode HID/input daemon for the newer IPTS firmware
- DTX and `surface-control` tools

If it's acceptable, I'd rather do those in follow-up PRs, but not sure whether that makes it OK to merge this into the main branch as is?

microsoft/surface/repos.nix
```diff
@@ -0,0 +1,20 @@
...
+    url="https://github.com/linux-surface/linux-surface.git";
+    rev="25ab2cf75e5eda5ab9739db1907300010c06dacf";
+    sha256="0h8624d7ix1p6ysw9bllmnnwnv164z8xkx56zj3vdczn91vmqcf9";
+  };
+
+  surface-ipts-firmware = fetchgit {
+    url="https://github.com/linux-surface/surface-ipts-firmware.git";
+    rev="bd5093318d2550d5d668241d0d34df4be3fc03ab";
+    sha256="0q7k52yk3jdliizl3cq1m3zbbqsx360wxybnbgkaq46rawzhw5bk";
+  };
```
Mic92, line 13: Do we still need this?

microsoft/surface/repos.nix
```diff
@@ -0,0 +1,20 @@
...
+
+  libwacom-surface = fetchgit {
+    url="https://github.com/linux-surface/libwacom-surface.git";
+    rev="12628e069298ac2e409501d423a1bf8567b5d1bb";
+    sha256="1lydyfx33crd0qvzdyqyp39d0d9fnsplch6429sagdsn6a139ap0";
+  };
```
Mic92, line 19: I would prefer to make this a package in nixpkgs later.

microsoft/surface/README.md
```diff
@@ -0,0 +1,75 @@
...
+- https://github.com/linux-surface/surface-aggregator-module/wiki/Testing-and-Installing
+
+## Firmware, Drivers and Support Tools
+
+### IPTS
+
+IPTS is used on most of the Surface range, except for Surface Go and Surface Laptop 3 (AMD version).
+
+The firmware files from the [surface-ipts-firmware repo](https://github.com/linux-surface/surface-ipts-firmware.git)
+are only needed on kernel 4.19.
```
Mic92, line 28: I would actually reduce complexity here and only support newer kernels.

microsoft/surface/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+On kernel 5.4 or newer you may also need the user-space `iptsd` daemon (below).
+
+### WiFi
+
+For the Surface Go, please see the "Issues" sections below.
+
+### IPTSD
+
+The `iptsd` daemon works with the `intel-precise-touch` driver to convert raw touch data into events
```
Mic92, line 38: Also this should go to nixpkgs.

microsoft/surface/README.md
```diff
@@ -0,0 +1,75 @@
+# Derivatives for Microsoft Surface notebooks
+
+These derivatives use the patches from the [linux-surface repo](https://github.com/linux-surface/linux-surface/tree/master/patches).
+
+## Kernel
+
+The kernel needs several patches to make it work correctly with some of the hardware on various
```
Mic92, line 7: We can keep the kernel here.

Mic92:
> If it's acceptable, I'd rather do those in follow-up PRs, but not sure whether that makes it OK to merge this into the main branch as is?

Fine with me.

microsoft/surface/repos.nix
mexisme: Actually, not any more.

microsoft/surface/README.md
mexisme: This comment still applies, but should be documented in the linux-surface repos, so no longer needed here.

microsoft/surface/README.md
mexisme: The intel precise touch driver is technically split into a kernelspace part (`intel-precise-touch`) and a user-space part (`iptsd`).

The kernelspace part collects the raw capacitive touch data from the Intel Management Engine, and provides an API for the user-space to connect to.
The user-space portion does some floating-point manipulation of the raw data and provides this to the HID / input subsystem.

I believe the IPTS system is not exclusively used on MS Surface, so could be moved elsewhere in the repo, but Linux can't use the hardware without both parts working together.

Could that mean both parts should stay in this repo?

Mic92: Having it in nixpkgs allows to test for build breakages and provide a binary cache. Having it here allows to update kernel and userspace together. Which one do you prefer?

mexisme: > Having it in nixpkgs allows to test for build breakages and provide a binary cache. Having it here allows to update kernel and userspace together. Which one do you prefer?

I feel like the kernel-space and user-space code should probably stay together at this stage, because neither will work without the other;  I'm checking whether they are tightly-coupled across releases, though.
I also feel that since the `iptsd` takes a fraction of the time that a new kernel takes to build, any time saved by binary caching might not matter right now.

It seems like the scripts in `//tests/` are working pretty well, too?

In future, I assume it would be OK to deprecate and move into `nixpkgs` if that makes more sense?

Mic92: Yes. But might cause troubles with stable releases (that don't have those packages yet) - so some backport would be needed.

mexisme: > I feel like the kernel-space and user-space code should probably stay together at this stage, because neither will work without the other; I'm checking whether they are tightly-coupled across releases, though.

I've confirmation from upstream that they are expected to be released together, and that while there is some compatibility between releases, the two sides are reasonably tightly-coupled at the API.

> Yes. But might cause troubles with stable releases (that don't have those packages yet) - so some backport would be needed.

Is this referring to my "deprecate and move into `nixpkgs`" question?
If so:  I know this could be less-smooth, and (personally) think if it happens, then the kernel patches would also have to move.

So my thinking goes like this:
I expect there'll be some negotiation to get these derivations accepted in `nixpkgs` as it looks like upstream can't support every single patch-release of a kernel, anyway. 
Matching that to whichever kernel is "current" in `nixpkgs` in a way that people are comfortable with might take some negotiation.

TBH, that's one of the advantages of `nixos-hardware` : it can get some use by people (e.g. me!) while that negotiation happens.

Does that sound reasonable?

Mic92: I don't think that we want to maintain more kernel configurations in nixpkgs. We need to test kernel modules for every one of them and it becomes tedious if they break for certain kernel configurations. So this kernel might remain in nixos-hardware for a while. Maybe I can also organize merges access for you so that kernel updates can be delivered timely.

Mic92:
Looks good to me if https://github.com/NixOS/nixos-hardware/pull/221#discussion_r558231445 https://github.com/NixOS/nixos-hardware/pull/221#discussion_r558231778 and https://github.com/NixOS/nixos-hardware/pull/221#discussion_r558232452 is addressed.

microsoft/surface/repos.nix
mexisme: I will extract this out.

microsoft/surface/README.md
mexisme: I dropped the explanation for older kernels, beyond explaining it used a no-longer-supported method.

PR #223, opened 2021-01-11
Update flake.nix

will-clarke:
I think there may have been a typo - it makes more sense to me to name macbook pro with `pro`. 🤷 😃


Mic92:
Thanks!

PR #224, opened 2021-01-16
Add MSI GS60 2QE

seqizz:


msi/gs60/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    initrd = {
+      availableKernelModules = [
+        "ahci"
+        "ehci_pci"
+        "rtsx_pci_sdmmc"
+        "sd_mod"
+        "usb_storage"
+        "xhci_pci"
+      ];
+    };
```
Mic92, line 27: Also this is correct, I think it is better to rely on information on nixos-generate hardware because it will be regenerated for new kernel versions. Otherwise this looks good to me:

```suggestion
```

PR #225, opened 2021-01-23
asus/zenbook/ux393: init

terinjokes:
Add initial configuration for the Asus Zenbook S UX393.


terinjokes:
I just asked for a refund because I realized I have the wrong revision.

PR #226, opened 2021-02-02
Add the Microsoft Surface configuration to flakes.nix

winterqt:


flake.nix
```diff
@@ -61,6 +61,7 @@
       lenovo-thinkpad-x1-6th-gen = import ./lenovo/thinkpad/x1/6th-gen;
       lenovo-thinkpad-x1-7th-gen = import ./lenovo/thinkpad/x1/7th-gen;
       lenovo-thinkpad-x1-extreme = import ./lenovo/thinkpad/x1-extreme;
+      microsoft-surface = import ./microsoft/surface;
```
Mic92, line 64: On related notes should surface-pro/3/default.nix import microsoft-surface?

winterqt: Since the original PR adding the new configuration didn’t remove the Pro 3 configuration, I kept it like it was.

That should work though.

Mic92: @mexisme should we import the more generic module also for this module?

winterqt: Per [the wiki](https://github.com/linux-surface/linux-surface/wiki/Supported-Devices-and-Features), SP3 is supported by the `linux-surface` patches, so it should be safe.

Mic92: Sorry for taking this PR hostage. If this is the case, we should just do it now because otherwise it will never happen.

winterqt:
Should I also make it so `microsoft/surface-pro/3` is just an alias to the generic configuration, so users not using flakes get switched over to the generic configuration without breaking their existing configuration?

flake.nix
mexisme: > @mexisme should we import the more generic module also for this module?

In general, I would say so; the intent of the linux-surface patches is to support all Surface devices.

But I've just woken, and can't fully remember the content of the Surface 3 specific derivation.

winterqt: > and can't fully remember the content of the Surface 3 specific derivation.

@mexisme It’s a single option that’s definitely made obsolete with the custom kernel, and doesn’t even include the custom kernel.
Based on this, I think it’s safe to completely remove.

flake.nix
```diff
@@ -61,7 +61,8 @@
       lenovo-thinkpad-x1-6th-gen = import ./lenovo/thinkpad/x1/6th-gen;
       lenovo-thinkpad-x1-7th-gen = import ./lenovo/thinkpad/x1/7th-gen;
       lenovo-thinkpad-x1-extreme = import ./lenovo/thinkpad/x1-extreme;
-      microsoft-surface-pro-3 = import ./microsoft/surface-pro/3;
```
Mic92, line 64: Looks good but can you also remove the file?

microsoft/surface-pro/3/default.nix
```diff
@@ -1,9 +1,4 @@
 { lib, ... }:
-
 {
-  # to use the type cover in the initrd
-  boot.kernelModules = [ "hid-microsoft" ];
-
-  # TODO: reverse compat
-  networking.wireless.enable = lib.mkDefault true;
+  imports = [ ../../surface ];
```
Mic92, line 3: I think this is better for backwards compatibility.

winterqt: That's actually what I was going to do, heh.

LGTM.

PR #227, opened 2021-02-03
microsoft-surface: add surface-control

winterqt:
This adds [surface-control](https://github.com/linux-surface/surface-control), a command line tool to control various aspects of the Microsoft Surface line of devices from the command line.

This PR depends on https://github.com/NixOS/nixpkgs/pull/111690, and is a draft until it is merged.

### DIFF
--- a/microsoft/surface/default.nix
+++ b/microsoft/surface/default.nix
@@ -1,5 +1,6 @@
 { config, lib, pkgs, ... }: {
-  imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
+  imports =
+    [ ./kernel ./hardware_configuration.nix ./firmware/surface-go/ath10k ];
 
   environment.systemPackages = with pkgs; [ surface-control ];
   users.groups.surface-control = { };
--- a/microsoft/surface/default.nix
+++ b/microsoft/surface/default.nix
@@ -1,8 +1,7 @@
-{ config, lib, pkgs, ... }:
-{
-  imports = [
-    ./kernel
-    ./firmware
-    ./hardware_configuration.nix
-  ];
+{ config, lib, pkgs, ... }: {
+  imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
+
+  environment.systemPackages = with pkgs; [ surface-control ];
+  users.groups.surface-control = {};
+  services.udev.packages = [ pkgs.surface-control ];
 }
--- a/microsoft/surface/default.nix
+++ b/microsoft/surface/default.nix
@@ -2,6 +2,6 @@
   imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
 
   environment.systemPackages = with pkgs; [ surface-control ];
-  users.groups.surface-control = {};
+  users.groups.surface-control = { };
   services.udev.packages = [ pkgs.surface-control ];
 }
--- a/microsoft/surface/README.md
+++ b/microsoft/surface/README.md
@@ -42,7 +42,11 @@ kernel-space driver into events for the HID / input sub-system.
 
 ### DTX, `surface-control`
 
-*TODO*
+#### surface-control
+
+For controlling the performance modes and other aspects of the device, the [`surface-control`](https://github.com/linux-surface/surface-control) tool is included.
+
+To be able to control the performance mode without using `sudo`, add your user to the `surface-control` group.
 
 # ToDo's Not Done
 


microsoft/surface/default.nix
```diff
@@ -1,8 +1,8 @@
...
-    ./kernel
-    ./firmware
-    ./hardware_configuration.nix
-  ];
+{ config, lib, pkgs, ... }: {
+  imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
+
+  environment.systemPackages = with pkgs; [ surface-control ];
+
+  services.udev.extraRules = ''
```
winterqt, line 6: There is almost certainly a better way to do this.
I considered adding it to the package itself, but it’s not in tree (it’s in the README of the package). What’s the best course of action to take? (leave it like it is, move to package, move to package and PR upstream to get it in tree?)

Mic92: You mean adding udev rules from the package or adding the package to nixpkgs?

winterqt: The latter is already being done, see the linked PR in this ones description.

I mean if I should either:
- Ask upstream to put the udev rule in a file instead of in the README, so it can be put in the package and updated by them as needed (which it probably won’t be often)
- Expose it in the package derivation, but still use a constant string like I do here

The first one is probably best, right?

danieldk: Upstreaming is almost always the best option. But if upstream does not want to add it, it can be added to the derivation as well, we have several derivations that include udev rules.

Mic92: Yes they just need to install udev rules into a standard location and than we can use the udev packages option in nixos.

winterqt: Upstream said they'll look into it.

winterqt: As of https://github.com/linux-surface/surface-control/commit/190f21fe32b4be430bb6da28edead3fb45ebdb68, there is now a udev rule embedded.

However, when adding it to the package and enabling it via `services.udev.packages`, I get:
```
/usr/bin/chmod is called in udev rules but is not executable or does not exist
```
which I expected and fixed in the initial version of this PR (with `${pkgs.coreutils}/bin/chmod` in the string), but not really sure how to go about this when it's in the package itself...

Once this is fixed, I'll add it to the Nixpkgs PR and we can get this merged.

@danieldk Any tips?

danieldk: > @danieldk Any tips?

`chmod`/`chown` can be specified by their full paths using e.g. `substituteInPlace`:

```
substituteInPlace 40-surface-control.rules \
   --replace "/usr/bin/chmod" "${coreutils}/bin/chmod" \
   --replace "/usr/bin/chown" "${coreutils}/bin/chown"
```

But I also wonder why they don't use `udev`'s built-in MODE and GROUP attributes:

```
KERNEL=="01:03:01:00:01", SUBSYSTEM=="surface_aggregator", GROUP="surface-control", MODE="0644"
```

Maybe that doesn't work because it is a sysfs file and not a device node? I don't know udev well enough to know the answer. If it does work, it may also be worthwhile to consider the use of the `uaccess` tag:

```
KERNEL=="01:03:01:00:01", SUBSYSTEM=="surface_aggregator", TAG+="uaccess"
```

But that is a security question (who do you want to give access).

microsoft/surface/default.nix
```diff
@@ -1,8 +1,6 @@
...
-{
-  imports = [
-    ./kernel
-    ./firmware
-    ./hardware_configuration.nix
-  ];
+{ config, lib, pkgs, ... }: {
+  imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
+
+  environment.systemPackages = with pkgs; [ surface-control ];
```
Mic92, line 4: I guess we need to wait until the channel moved.

winterqt: If you mean a stable release: no need, this configuration already requires the use of unstable.
If you mean unstable: yeah.

Mic92: No. I mean unstable.

winterqt: How long does that usually take?

Mic92: It should be already there. Let's restart the test.

winterqt:
Hey @Mic92, can we get this merged? Thanks! :)

Mic92:
Looks like we got a merge conflict now. Maybe afterwards?

winterqt:
Should be good now.

microsoft/surface/default.nix
```diff
@@ -1,8 +1,7 @@
-{ config, lib, pkgs, ... }:
-{
-  imports = [
-    ./kernel
-    ./hardware_configuration.nix
-    ./firmware/surface-go/ath10k
-  ];
+{ config, lib, pkgs, ... }: {
+  imports = [ ./kernel ./firmware ./hardware_configuration.nix ];
```
Mic92, line 2: `./firmware` does not exists.

PR #230, opened 2021-02-09
Surface Go:  use KillerNetworking firmware instead of official non-free copy

mexisme:
This overrides the firmware for the ath10k QCA6174 device, in the standard/upstream Linux Non-free Firmware, with the `board.bin` file from Killer Networking.
This firmware is required by the Surface Go, as per https://github.com/linux-surface/linux-surface/wiki/Surface-Go#wifi-firmware


mexisme:
Things to do:
- Try to d/l from support.killernetworking.com
  - Appears to have some code that gives an HTTP 406 when using fetchurl / nix-prefetch-url
- Implement option wrapping around whether to use the Surface Go

microsoft/surface/TODO.org
```diff
@@ -14,14 +14,26 @@
...
     - [X] Download github.com/kvalo/ath10k-firmware ?
     - [X] Download kernel.org linux-firmware?
-    - [ ] Download `board.bin` from Killer Networks?
-    - [ ] Install ath10k firmware fix for Surface Go
+    - [X] Download `board.bin` from Killer Networks?
+    - [X] Install ath10k firmware fix for Surface Go
       - `board-2.bin` needs to be removed / replaced with `board.bin`
+    - [ ] Add a config option for enabling the Surface Go Firmware
+    - [ ] Replace the vendorised `board.bin` with a `fetchurl`
+      - Getting HTTP 406 when using `fetchurl` or `nix-prefetch-url` that has to be diagnosed
```
Mic92, line 36: Where did you got this from in the first place?

microsoft/surface/repos.nix
```diff
@@ -1,8 +1,27 @@
...
+
+  ath10k-firmware = fetchgit {
+    url="https://github.com/kvalo/ath10k-firmware";
+    rev="84b47062aab31d67156e0a3ef593a6999a12864b";
+    sha256="0l8wfj8z4jbb31nzqkaxisby0n6061ix01c5di9bq66iby59j8py";
+  };
+
+    surface-go-ath10k-firmware = fetchurl {
+    url="https://support.killernetworking.com/K1535_Debian/board.bin";
+    # url="https://www.killernetworking.com/support/K1535_Debian/board.bin";
```
Mic92, line 23: Is see.

Mic92: This looks like it could break anytime the firmware is upgraded. Do they have versioned downloads?

microsoft/surface/firmware/default.nix
```diff
@@ -2,5 +2,7 @@
 {
   hardware.enableAllFirmware = true;
   hardware.firmware = [
+    # TODO: Wrap with an option:
+    (pkgs.callPackage ./ath10k.nix {})
```
Mic92, line 6: Maybe just re-use `config.hardware.enableAllFirmware`?

microsoft/surface/repos.nix
mexisme: > This looks like it could break anytime the firmware is upgraded. Do they have versioned downloads?

Not as far as I can tell.
And their Linux docs only say "Use the official firmware repo for Linux" — which is the one that has the non-working firmware.

Every other post of wiki entry I could find for other OSes (e.g. Arch, Debian) has the same advice for d/l directly from Killer Networking.

There appear to be some efforts to get this firmware included in the official firmware, but it seems to have stalled.

This draft PR has the `board.bin` embedded in it, for testing.
Once I'm happy with how the `.nix` code is looking, I plan to figure out how to get a stable version of the firmware BLOB.... which I think I've neglected to add to `TODO.org` (will fix that)

microsoft/surface/firmware/default.nix
mexisme: > Maybe just re-use config.hardware.enableAllFirmware?

Are you referring to the `TODO` about wrapping with an option?
If so:  because that essentially "breaks" the original QCA6174 firmware drop by deleting the `board.bin` and `board-2.bin` files, and replacing with this new one, it felt appropriate to let people explicitly choose this specific part, rather than force it on them.

Mic92: Fair point.

microsoft/surface/repos.nix
Mic92: Should we upload the firmware to archive.org instead? Unversioned links is a bit of a minefield.

mexisme: I hadn't thought of that.
Is there a risk of this dropping off, too?

This is what I had thought:

- I suggested that the GitHub linux-surface owners might like to put it in a repo of their own approx. a week ago; I've been quite busy, so haven't had time to follow that up.

- I tried to follow the email thread to request the upstream `ath10k` owner add this `board.bin` but it's complete radio silence, AFAICT.
  I'm still hoping that sometime soon, they either respond to the email request to add it, or shares info about how to create the aggregate `board-2.bin` firmware blob.

- Should I create a public repo of my own, and add it there? This feels a bit like the last option, though.

Mic92: All options sounds acceptable to me.

mexisme: > All options sounds acceptable to me.

Thanks.

The upstream team are talking about my suggestion to put the firmware in their own repo, but also whether they should combine some of their other firmware repo's into a single one.
Until that's decided/done I'm going to use the copy in my own repo — primarily to unstick this PR.

microsoft/surface/firmware/surface-go/ath10k/default.nix
```diff
@@ -0,0 +1,22 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.hardware.microsoft-surface.firmware.surface-go-ath10k;
+in
+{
+  options = {
+    hardware.microsoft-surface.firmware.surface-go-ath10k = {
+      replace = lib.mkEnableOption ''Use the "board.bin" firmware for ath10k-based WiFi on Surface Go.'';
```
mexisme, line 8: I renamed this from `enable` to `replace` for a clearer option-name.
This will hopefully avoid name-clashes with future alternative fixes for this problem.

e.g. there's an experimental kernel patch being discussed, which will hopefully support a way to non-destructively use the Surface Go firmware blob:
- https://github.com/linux-surface/kernel/commit/22ef83836c4aa89e9eb98de9b47ed24b6c2a1d45

microsoft/surface/repos.nix
```diff
@@ -1,8 +1,26 @@
...
+    sha256="0l8wfj8z4jbb31nzqkaxisby0n6061ix01c5di9bq66iby59j8py";
+  };
+
+  surface-go-ath10k-firmware_backup = fetchgit {
+    url="git@github.com:mexisme/linux-surface_ath10k-firmware.git";
+    rev="74e5409e699383d6ca2bc4da4a8433d16f3850b1";
+    sha256="169vgvxpgad9anmchs22fj5qm6ahzjfdnwhd8pc280q705vx6pjk";
   };
```
Mic92, line 25: Have you tested this code? I don't think you can actually pull from ssh-based git's in the nix sandbox without additional configuration:

```suggestion
  surface-go-ath10k-firmware_backup = fetchFromGitHub {
    # based on https://support.killernetworking.com/K1535_Debian/board.bin
    owner = "mexisme";
    repo = "linux-surface_ath10k-firmware";
    rev="74e5409e699383d6ca2bc4da4a8433d16f3850b1";
    sha256="<TODO>";
  };
```

microsoft/surface/repos.nix
```diff
@@ -1,8 +1,26 @@
-{ lib, pkgs, fetchgit }:
+{ lib, pkgs, fetchgit, fetchurl }:
 {
   linux-surface = fetchgit {
```
Mic92, line 3: also switch to fetchFromGithub here since it is faster especially for large repositories.

microsoft/surface/repos.nix
```diff
@@ -1,8 +1,26 @@
...
     url="https://github.com/linux-surface/linux-surface.git";
-    rev="25ab2cf75e5eda5ab9739db1907300010c06dacf";
-    sha256="0h8624d7ix1p6ysw9bllmnnwnv164z8xkx56zj3vdczn91vmqcf9";
+    rev="2b6d2db5bfc7b678c37badc80e0a223bf89b5ef7";
+    sha256="03qlwxijznmk043zcpwl1yc56zhrlwdaw8k1cymb8zrxfpgw0r14";
+  };
+
+  ath10k-firmware = fetchgit {
```
Mic92, line 9: and here.

microsoft/surface/firmware/surface-go/ath10k/default.nix
Mic92: ok.

microsoft/surface/repos.nix
mexisme: Argh!  Yes, I knew that, and it's a silly error on my part.
I wonder if it allowed me to because I have the ssh-agent auto-loaded.

microsoft/surface/repos.nix
mexisme: OK, I wasn't aware of that improvement.

microsoft/surface/repos.nix
mexisme: FYI:  I've updated this PR to use `fetchFromGitHub` where possible.

mexisme:
Retesting the build + install, now.
It'll take a wee while, and I'll drop a comment in here when it completes.

mexisme:
> Retesting the build + install, now.
> It'll take a wee while, and I'll drop a comment in here when it completes.

I left the rebuild going overnight (getting too late) and can confirm it's worked successfully.

microsoft/surface/repos.nix
```diff
@@ -1,8 +1,29 @@
-{ lib, pkgs, fetchgit }:
+{ lib, pkgs, fetchFromGitHub, fetchurl }:
 {
-  linux-surface = fetchgit {
-    url="https://github.com/linux-surface/linux-surface.git";
+  linux-surface = fetchFromGitHub {
+    owner="linux-surface";
+    repo="linux-surface";
     rev="f8fab978a480a4ed57e9ebb6928683b2e443c1c5";
     sha256="0zwybprwjckpapxm6gxzh6hwdd1w91g5sjxn6z52zlvvjpkmw959";
```
Mic92, line 7: Please check what if the has is correct here. It should change if you go from `fetchgit` to `fetchFromGitHub`. You can check by changing one character in the hash and nix still producing the same hash.

mexisme: I just rechecked it:

```
% nix-prefetch-github linux-surface linux-surface --rev f8fab978a480a4ed57e9ebb6928683b2e443c1c5
{
    "owner": "linux-surface",
    "repo": "linux-surface",
    "rev": "f8fab978a480a4ed57e9ebb6928683b2e443c1c5",
    "sha256": "0zwybprwjckpapxm6gxzh6hwdd1w91g5sjxn6z52zlvvjpkmw959",
    "fetchSubmodules": true
}
```

Mic92: Yes, but the difference is if you use `fetchSubmodules` than it is actually using `fetchgit`.

Mic92: But in this case it seems fine.

mexisme: > Yes, but the difference is if you use `fetchSubmodules` than it is actually using `fetchgit`.

Which I'm guessing might be part of why it worked?
AFAICT, the default is `false` for `fetchForGitHub`?

> But in this case it seems fine.
I have tested rebuilding+installing this twice on a Surface Go — as that's kind-of my whole reason for caring about this!

PR #231, opened 2021-02-12
mpb15-1: init

hlolli:
This is my first PR to this repo, so I'm not sure what and what not to include. I also want to add that I'm still working on getting suspend and TouchBar working properly, I will follow up with those changes later, as well as a fix for mbpfan which needs a patch for the bus-id. I'll also add that later.

But important fact here I'm not including and I need an opinion about how to approach that. Which is that it's almost without exception, a requirement to use the archlinux's kernel patches https://github.com/aunali1/linux-mbp-arch, which I applied successfully like this. Note that I ignore 1 patch because it was merged upstream to the kernel (I expect most of these to be merged into the kernel, but I'm not 100% sure).

```nix
kernelPatches = (let
      src = (pkgs.fetchFromGitHub {
        owner = "aunali1";
        repo = "linux-mbp-arch";
        rev = "60cef373c14ba6a7b35d0af67d04dce7eb604f2e";
        sha256 = "11a75jiak14b2nyx283c39gry1xi57klsi6mjzljxliaih2agqqc";
       });
      items = with builtins;
      (filter (i:
        (match ".*\.patch$" i != null) &&
	# already merged https://github.com/torvalds/linux/commit/81a86e1bd8e7060ebba1718b284d54f1238e9bf9
        (match ".*0003-iwlwifi.*" i == null))
        (attrNames (readDir src)));
      patches = map (i: {name = i; patch = (pkgs.writeTextFile {name = i; text = builtins.readFile "${src}/${i}";});}) items;
      in (patches)
    );
    
```

One option would be to just document it. Or add it to nixpkgs, in which case I'd worry if these patches would cause problem for non mbp users.

Also the test is failing because of `boot.loader.systemd-boot.enable = false;` 

```
The option `boot.loader.systemd-boot.enable' has conflicting definition values:
- In `<unknown-file>': true
- In `/home/hlolli/nixos-hardware/apple/macbook-pro/15-1': false
```

Would it be ok to remove https://github.com/NixOS/nixos-hardware/blob/9c952961f1f1a643b8b8e5d4efab6717afec1bbe/tests/build-profile.nix#L5 (I think it's from there?).

### DIFF
--- a/apple/kernel-modules/bridge-drv.nix
+++ b/apple/kernel-modules/bridge-drv.nix
@@ -1,0 +1,40 @@
+{ pkgs, kernel }:
+
+pkgs.stdenv.mkDerivation rec {
+  name = "mbp-bridge";
+  version = "b43fcc069da73e051072fde24af4014c9c487286";
+
+  src = pkgs.fetchFromGitHub {
+    owner = "MCMrARM";
+    repo = "mbp2018-bridge-drv";
+    rev = version;
+    sha256 = "0ac2l51ybfrvg8m36x67rsvgjqs1vwp7c89ssvbjkrcq3y4qdb53";
+  };
+
+  sourceRoot = "source";
+  hardeningDisable = [ "pic" "format" ];
+  nativeBuildInputs = kernel.moduleBuildDependencies;
+
+  makeFlags = [
+    "KERNELRELEASE=${kernel.modDirVersion}"
+    "KERNEL_DIR=${kernel.dev}/lib/modules/${kernel.modDirVersion}/build"
+    "INSTALL_MOD_PATH=${placeholder "out"}"
+  ];
+
+  postUnpack = ''
+    chmod -R +rw source
+  '';
+
+  buildPhase = ''
+    echo 'obj-m += bce.o
+bce-objs := pci.o mailbox.o queue.o queue_dma.o vhci/vhci.o vhci/queue.o vhci/transfer.o audio/audio.o audio/protocol.o audio/protocol_bce.o audio/pcm.o' > Makefile
+    make -C ${kernel.dev}/lib/modules/${kernel.modDirVersion}/build \
+      M=$(pwd) modules
+  '';
+
+  installPhase = ''
+   mkdir -p $out/lib/modules/${kernel.modDirVersion}
+   cp *.ko $out/lib/modules/${kernel.modDirVersion}
+  '';
+
+}
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ See code for all available configurations.
 | Apple MacBook Air 6,X             | `<nixos-hardware/apple/macbook-air/6>`             |
 | [Apple MacBook Pro 10,1][]        | `<nixos-hardware/apple/macbook-pro/10-1>`          |
 | Apple MacBook Pro 12,1            | `<nixos-hardware/apple/macbook-pro/12-1>`          |
+| Apple MacBook Pro 15,1            | `<nixos-hardware/apple/macbook-pro/15-1>`          |
 | Asus TUF FX504GD                  | `<nixos-hardware/asus/fx504gd>`                    |
 | BeagleBoard PocketBeagle          | `<nixos-hardware/beagleboard/pocketbeagle>`        |
 | [Dell G3 3779][]                  | `<nixos-hardware/dell/g3/3779>`                    |
--- a/apple/README.md
+++ b/apple/README.md
@@ -1,6 +1,61 @@
+## Macbooks with T2 chips (15 and later)
+
+# builtin wifi
+
+There's currently no existing firmware for Macbooks containing the new T2 chips.
+When installing NixOs you'll need to rely on 3rd party hardware (ex. usb-c -> ethernet unit).
+
+To get the builtin wifi running, make sure you store the firmware specific to your macbook
+as is described here: https://gist.github.com/TRPB/437f663b545d23cc8a2073253c774be3#wifi
+
+If you already formatted your osx partition (like I did), you can probably still access
+the wifi with trial and error using the firmware uploaded here https://packages.aunali1.com/apple/wifi-fw/18G2022/C-4364__s-B2
+
+To apply the firmware to your NixOs, I did the following which can be modified matching your machine and your firmware
+
+```nix
+{ stdenv,lib, fetchurl }:
+
+let clm_blob = fetchurl {
+      url = "https://packages.aunali1.com/apple/wifi-fw/18G2022/C-4364__s-B2/maui.clmb";
+      sha256 = "0c626zskm6yx3dxpvhx0q6j1hsvqwxdwm6nif3kkfahyfx69glny";
+    };
+    apple_txt = fetchurl {
+      url = "https://packages.aunali1.com/apple/wifi-fw/18G2022/C-4364__s-B2/P-midway-ID_M-HRPN_V-m__m-7.7.txt";
+      sha256 = "01nxmcds26aanvwmpjxwh0kk9i3l7v3gh25qrmy58cmw4mwxhsa5";
+    };
+
+in stdenv.mkDerivation {
+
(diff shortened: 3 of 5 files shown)


apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
+{ pkgs, kernel }:
+
+pkgs.stdenv.mkDerivation rec {
+  name = "mbp-bridge";
```
Mic92, line 4: I think this is a good candidate for nixpkgs.

apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
...
+    "INSTALL_MOD_PATH=${placeholder "out"}"
+  ];
+
+  postUnpack = ''
+    chmod -R +rw source
+  '';
+
+  buildPhase = ''
+    echo 'obj-m += bce.o
+bce-objs := pci.o mailbox.o queue.o queue_dma.o vhci/vhci.o vhci/queue.o vhci/transfer.o audio/audio.o audio/protocol.o audio/protocol_bce.o audio/pcm.o' > Makefile
```
Mic92, line 30: Why is this step necessary?

Mic92:
> This is my first PR to this repo, so I'm not sure what and what not to include. I also want to add that I'm still working on getting suspend and TouchBar working properly, I will follow up with those changes later, as well as a fix for mbpfan which needs a patch for the bus-id. I'll also add that later.
> 
> But important fact here I'm not including and I need an opinion about how to approach that. Which is that it's almost without exception, a requirement to use the archlinux's kernel patches [aunali1/linux-mbp-arch](https://github.com/aunali1/linux-mbp-arch), which I applied successfully like this. Note that I ignore 1 patch because it was merged upstream to the kernel (I expect most of these to be merged into the kernel, but I'm not 100% sure).
> 
> ```nix
> kernelPatches = (let
>       src = (pkgs.fetchFromGitHub {
>         owner = "aunali1";
>         repo = "linux-mbp-arch";
>         rev = "60cef373c14ba6a7b35d0af67d04dce7eb604f2e";
>         sha256 = "11a75jiak14b2nyx283c39gry1xi57klsi6mjzljxliaih2agqqc";
>        });
>       items = with builtins;
>       (filter (i:
>         (match ".*\.patch$" i != null) &&
> 	# already merged https://github.com/torvalds/linux/commit/81a86e1bd8e7060ebba1718b284d54f1238e9bf9
>         (match ".*0003-iwlwifi.*" i == null))
>         (attrNames (readDir src)));
>       patches = map (i: {name = i; patch = (pkgs.writeTextFile {name = i; text = builtins.readFile "${src}/${i}";});}) items;
>       in (patches)
>     );
>     
> ```

How much patches does this fork come with?


> 
> One option would be to just document it. Or add it to nixpkgs, in which case I'd worry if these patches would cause problem for non mbp users.
> 
> Also the test is failing because of `boot.loader.systemd-boot.enable = false;`

This could become `mkDefault = true;` actually.

> 
> ```
> The option `boot.loader.systemd-boot.enable' has conflicting definition values:
> - In `<unknown-file>': true
> - In `/home/hlolli/nixos-hardware/apple/macbook-pro/15-1': false
> ```
> 
> Would it be ok to remove
> 
> https://github.com/NixOS/nixos-hardware/blob/9c952961f1f1a643b8b8e5d4efab6717afec1bbe/tests/build-profile.nix#L5
> (I think it's from there?).

apple/kernel-modules/bridge-drv.nix
hlolli: I had been building a lot of kernel modules, and realized that instead of patching the makefile, which often is using absolute kernel paths with uname in it etc, I'd just replace it with the parts actually needed. I'll see if I can get away without this step.

apple/kernel-modules/bridge-drv.nix
hlolli: ok, will more it

hlolli:
> Also see microsoft/surface/kernel/linux-5.10.2/default.nix

perfect.

I guess this will take a few more days at least. I have the TouchPad working and there are some fallacies in this PR regarding the touchpad. But defenitely a good feeling to see all the hardware starting to light up :P

apple/kernel-modules/bridge-drv.nix
Mic92, line 30: There is indeed an absolute path in the all target. But since you already using the right kernel makefile, you can just ignore it. The following works for me:

```suggestion
```

Mic92: see below.

apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
...
+    make -C ${kernel.dev}/lib/modules/${kernel.modDirVersion}/build \
+      M=$(pwd) modules
+  '';
+
+  installPhase = ''
+   mkdir -p $out/lib/modules/${kernel.modDirVersion}
+   cp *.ko $out/lib/modules/${kernel.modDirVersion}
```
Mic92, line 37:
```suggestion
make -C ${kernel.dev}/lib/modules/${kernel.modDirVersion}/build $makeFlags \
      M=$(pwd) modules_install
```

Mic92: This also insures it gets compressed.

apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  sourceRoot = "source";
+  hardeningDisable = [ "pic" "format" ];
+  nativeBuildInputs = kernel.moduleBuildDependencies;
+
+  makeFlags = [
+    "KERNELRELEASE=${kernel.modDirVersion}"
+    "KERNEL_DIR=${kernel.dev}/lib/modules/${kernel.modDirVersion}/build"
+    "INSTALL_MOD_PATH=${placeholder "out"}"
+  ];
```
Mic92, line 22: Not really used:
```suggestion
```

apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  postUnpack = ''
+    chmod -R +rw source
+  '';
```
Mic92, line 26: I could still build it after removing that. Is it a left-over?

```suggestion
```

apple/kernel-modules/bridge-drv.nix
```diff
@@ -0,0 +1,40 @@
...
+  version = "b43fcc069da73e051072fde24af4014c9c487286";
+
+  src = pkgs.fetchFromGitHub {
+    owner = "MCMrARM";
+    repo = "mbp2018-bridge-drv";
+    rev = version;
+    sha256 = "0ac2l51ybfrvg8m36x67rsvgjqs1vwp7c89ssvbjkrcq3y4qdb53";
+  };
+
+  sourceRoot = "source";
```
Mic92, line 14:
```suggestion
```

apple/kernel-modules/bridge-drv.nix
hlolli: yes, leftover, I'll remove. I was copy pasting the same template, and this was some junk code.

hlolli: yes, leftover, I'll remove. I was copy pasting the same template, and this was some junk code.

apple/kernel-modules/bridge-drv.nix
hlolli: nice thanks! I'll apply all of these changes to a nixpkgs PR.

Mic92:
Looks like the driver is not included and we have no real user of the hardware. So close this for now until we have someone else to take this over.

howarddo2208:
Hey there, I'm using a 15,1 and just dual boot installed nixOS through graphical kde iso image. How can I test this out? I'm still a nix noob

PR #241, opened 2021-03-23
Add profiles for AMD GPUs

j0hax:
Inspired by [https://github.com/NixOS/nixpkgs/issues/116619](https://github.com/NixOS/nixpkgs/issues/116619), I decided to create basic profiles for AMD GPUs according to the NixOS Wiki [article](https://nixos.wiki/wiki/AMD_GPU).

The profile for Sea Islands works fine on my machine (R9 290X), I expect the one for Southern Islands to work equally well.

### DIFF
--- a/common/gpu/amd/default.nix
+++ b/common/gpu/amd/default.nix
@@ -1,0 +1,17 @@
+{ config, lib, pkgs, ... }:
+
+{
+  boot.initrd.kernelModules = [ "amdgpu" ];
+  services.xserver.videoDrivers = [ "amdgpu" ];
+  
+  hardware.opengl.extraPackages = with pkgs; [
+    rocm-opencl-icd
+    rocm-opencl-runtime
+    amdvlk
+  ];
+
+  hardware.opengl = {
+    driSupport = lib.mkDefault true;
+    driSupport32Bit = lib.mkDefault true;
+  };
+} 
--- a/common/gpu/amd/sea-islands/default.nix
+++ b/common/gpu/amd/sea-islands/default.nix
@@ -1,4 +1,5 @@
 {
   imports = [ ../. ];
+  # Explicitly set amdgpu support in place of radeon
   boot.kernelParams = [ "radeon.cik_support=0" "amdgpu.cik_support=1" ];
 }
--- a/common/gpu/amd/sea-islands/default.nix
+++ b/common/gpu/amd/sea-islands/default.nix
@@ -1,0 +1,4 @@
+{
+  imports = [ ../. ];
+  boot.kernelParams = [ "radeon.cik_support=0" "amdgpu.cik_support=1" ];
+}
--- a/common/gpu/amd/southern-islands/default.nix
+++ b/common/gpu/amd/southern-islands/default.nix
@@ -1,4 +1,5 @@
 {
   imports = [ ../. ];
+  # Explicitly set amdgpu support in place of radeon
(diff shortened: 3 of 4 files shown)


common/gpu/amd/default.nix
```diff
@@ -0,0 +1,17 @@
+{ config, lib, pkgs, ... }:
+
+{
+  boot.initrd.kernelModules = [ "amdgpu" ];
```
Mic92, line 4: In nixpkgs we have `nixpkgs/nixos/modules/hardware/video/amdgpu.nix`.
I think we should extend it with it. We still could have a smaller module
here that enables this profile.
If you link the nixpkgs PR. I can have a look at it.

common/gpu/amd/sea-islands/default.nix
```diff
@@ -0,0 +1,4 @@
+{
+  imports = [ ../. ];
+  boot.kernelParams = [ "radeon.cik_support=0" "amdgpu.cik_support=1" ];
```
Mic92, line 3: Can you explain what this module enables and why we only have it for certain generations?

notgne2: Certain AMD cards without these flags will refuse to use the amdgpu driver and only run with the Radeon driver, despite the amdgpu driver working absolutely fine and often better, I've had these parameters and the `si` equivilents in my common configuration for years now.

The Arch wiki provides a little more insight on this too: https://wiki.archlinux.org/index.php/AMDGPU

Mic92: Ok. It would be great if this would be commented in a comment above the flag.

common/gpu/amd/default.nix
j0hax: Please clarify, you would like amdgpu set as a kernel module if it is set as an Xorg driver?

I could make a pull request like this for the file you mentioned:

```diff
  {
    config = mkIf (elem "amdgpu" config.services.xserver.videoDrivers) {
      boot.blacklistedKernelModules = [ "radeon" ];
+     boot.initrd.kernelModules = [ "amdgpu" ];
    };
  }

```

common/gpu/amd/sea-islands/default.nix
j0hax: I wonder if we should still explicitly set the `radeon.cik_support=0` parameter when the radeon driver can be blacklisted by `nixpkgs/nixos/modules/hardware/video/amdgpu.nix`?

common/gpu/amd/default.nix
Mic92: I mean it could populate opengl.extraPackages. I don't have an opinion on wether it should enable hardware.opengl.

common/gpu/amd/sea-islands/default.nix
Mic92: Makes sense.

common/gpu/amd/default.nix
j0hax: The `amdgpu.nix` module has been removed in https://github.com/NixOS/nixpkgs/pull/111551#issuecomment-802082766.

I think we should continue work here.

common/gpu/amd/sea-islands/default.nix
Mic92: Ok. From my side I only would add a comment here.

PR #247, opened 2021-04-05
Add mnt-reform2-nitrogen8m

ehmry:
An aarch64 laptop. If this is merged then I will create a flake that imports this module to build an installer.

https://www.crowdsupply.com/mnt/reform
https://mntre.com/

### DIFF
--- a/mnt/reform2-nitrogen8m/kernel/default.nix
+++ b/mnt/reform2-nitrogen8m/kernel/default.nix
@@ -1,0 +1,97 @@
+{ lib, buildLinux, fetchurl, fetchgit, linux_5_7, kernelPatches, ... }@args:
+
+let
+  linux = linux_5_7;
+  systemImageSrc = fetchgit {
+    url = "https://source.mnt.re/reform/reform-system-image.git";
+    rev = "ef6bae4def9ef08ee388254abf4f1839d44c07a1";
+    sha256 = "03nnwjzm8a0bk821p6sfscd04x3jgj29l16ccdkg10xcv2g2z5s8";
+  } + "/reform2-imx8mq";
+in lib.overrideDerivation (buildLinux (args // {
+  inherit (linux) src version;
+
+  features = {
+    efiBootStub = false;
+    iwlwifi = false;
+  } // (args.features or { });
+
+  kernelPatches = let
+    patchDir = "${systemImageSrc}/template-kernel/patches";
+    reformPatches = map ({ name, extraConfig ? "" }: {
+      inherit name extraConfig;
+      patch = "${patchDir}/${name}.patch";
+    }) [
+      {
+        name = "0001-drm-bridge-Add-NWL-MIPI-DSI-host-controller-support";
+        extraConfig = ''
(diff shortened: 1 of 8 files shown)


mnt/reform2-nitrogen8m/kernel/linux-5.7.nix
```diff
@@ -0,0 +1,18 @@
+{ stdenv, buildPackages, fetchurl, perl, buildLinux, modDirVersionArg ? null, ... } @ args:
+
+with stdenv.lib;
```
SuperSandro2000, line 3:
```suggestion
with lib;
```

mnt/reform2-nitrogen8m/kernel/linux-5.7.nix
```diff
@@ -0,0 +1,18 @@
+{ stdenv, buildPackages, fetchurl, perl, buildLinux, modDirVersionArg ? null, ... } @ args:
```
SuperSandro2000, line 1:
```suggestion
{ lib, stdenv, buildPackages, fetchurl, perl, buildLinux, modDirVersionArg ? null, ... } @ args:
```

SuperSandro2000 (changes requested):

mnt/reform2-nitrogen8m/kernel/linux-5.7.nix
ehmry: Good catch, I copied this from and old Nixpkgs.

mnt/reform2-nitrogen8m/kernel/default.nix
```diff
@@ -0,0 +1,93 @@
...
+        name = "mnt3008-MNT-Reform-imx8mq-missing-makefile-change-for-imx-dc";
+        extraConfig = ''
+          DRM_IMX_DCSS m
+        '';
+      }
+    ];
+  in lib.lists.unique (kernelPatches ++ reformPatches ++ [{
+    name = "MNT-Reform-imx8mq-config";
+    patch = null;
+    extraConfig = builtins.readFile ./kernel-config;
```
Mic92, line 80: Is there a way to avoid adding so many big files to this repository like initial-sound.state and the kernel config?

ehmry: The initial-sound.state file could be taken from a git checkout, but the kernel config is something custom.

ehmry:
bump

Mic92:
Can we store the kernel config here? https://github.com/NixOS/nixos-hardware/releases/tag/not-a-release

mnt/reform2-nitrogen8m/kernel/default.nix
Mic92: Can we take kernel config and initial sound state from here? https://github.com/NixOS/nixos-hardware/releases/tag/mnt-reform2-nitrogen8m-v1

ehmry: Yes, that works, good idea.

PR #252, opened 2021-04-25
Add support for Enhanced Session on Hyper-V

bltavares:
This commit adds [Hyper-V Enhanced Session](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/enhanced-session-mode)
over RDP to get high-DPI video, using the [tips from](https://the.kalaclista.com/posts/2018/10/21/000000/) and [video from this issue](https://github.com/NixOS/nixpkgs/issues/57290).

### DIFF
--- a/microsoft/hyper-v/default.nix
+++ b/microsoft/hyper-v/default.nix
@@ -1,0 +1,104 @@
+{ config, pkgs, options, lib, ... }:
+
+with lib; {
+
+  options = {
+    hyperv.enhancedSession = mkOption
+      {
+        type = types.bool;
+        default = true;
+        description = ''
+          Enables enhanced hyperv session over RDP
+          Enable Guest Services on Hyper-V Management settings and
+          don't forget to enable access over VSock on a Admin powershell with:
+
+          % Set-Vm -VmName <VM> -EnhancedSessionTransportType HvSocket
+        '';
+      };
+
+    hyperv.video = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Enables Display Manager support on HyperV
+      '';
+    };
+
+    hyperv.defaultWindowManager = mkOption
+      {
+        type = types.separatedString " ";
+        default = "${config.services.xserver.displayManager.sessionData.wrapper}";
+        description = ''
+          If ~/startwm.sh exists as an executable script, it will take priority over this option.
+        '';
+      };
+  };
+
+  config = mkMerge ([
+    {
+      boot.kernelModules = [ "hv_sock" ];
+      virtualisation.hypervGuest.enable = true;
+    }
+
+    (mkIf config.hyperv.video {
+      services.xserver = {
+        modules = [ pkgs.xorg.xf86videofbdev ];
+        videoDrivers = [ "hyperv_fb" ];
+      };
+    })
+
+    (mkIf
+      config.hyperv.enhancedSession
+      {
+        services.xrdp = {
+          enable = true;
+          defaultWindowManager = config.hyperv.defaultWindowManager;
+          package = pkgs.xrdp.overrideAttrs (old: rec {
+            configureFlags = old.configureFlags ++ [ " --enable-vsock" ];
+
+            postInstall = old.postInstall + ''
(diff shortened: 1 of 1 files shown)


microsoft/hyper-v/default.nix
```diff
@@ -0,0 +1,104 @@
+{ config, pkgs, options, lib, ... }:
+
+with lib; {
+
+  options = {
+    hyperv.enhancedSession = mkOption
```
Mic92, line 6: I think these things are general enough to be put into nixpkgs itself.
We have `nixpkgs/nixos/modules/virtualisation/hyperv-guest.nix`.
This repo is more for hardware-specific quirks that can be expressed with minimal amount of nixos module options.
While your contribution is provides configuration for a hypervisor target than can be easier test by a wider audience.

Could you open a pull request towards nixpkgs? If you put a link here. I can have a look.

microsoft/hyper-v/default.nix
```diff
@@ -0,0 +1,104 @@
...
+    })
+
+    (mkIf
+      config.hyperv.enhancedSession
+      {
+        services.xrdp = {
+          enable = true;
+          defaultWindowManager = config.hyperv.defaultWindowManager;
+          package = pkgs.xrdp.overrideAttrs (old: rec {
+            configureFlags = old.configureFlags ++ [ " --enable-vsock" ];
```
Mic92, line 57: I think we can enable vsock by default in xrdp.

microsoft/hyper-v/default.nix
```diff
@@ -0,0 +1,104 @@
...
+
+            postInstall = old.postInstall + ''
+              ${pkgs.gnused}/bin/sed -i -e 's!use_vsock=false!use_vsock=true!g'                               $out/etc/xrdp/xrdp.ini
```
Mic92, line 60: Do you see any way to move this configuration outside of the module? Than we don't need a rebuild to change options.

microsoft/hyper-v/default.nix
```diff
@@ -0,0 +1,104 @@
...
+      default = true;
+      description = ''
+        Enables Display Manager support on HyperV
+      '';
+    };
+
+    hyperv.defaultWindowManager = mkOption
+      {
+        type = types.separatedString " ";
+        default = "${config.services.xserver.displayManager.sessionData.wrapper}";
```
Mic92, line 30: Maybe cc @jtojnar if you open the PR towards nixpkgs regarding this option.

microsoft/hyper-v/default.nix
bltavares: I'm new to nix, so I'm not sure. On other systems, we edit a file on `/etc/xrdp` to make it work, but I'm not sure how to make it on nix

microsoft/hyper-v/default.nix
bltavares: As there was a `hyper-v` folder on this project I thought it was the appropriate place. I could share latter on nixpkgs instead (but likely on a weekend).

microsoft/hyper-v/default.nix
bltavares: I didn't have much success in this setting tho - I had to create `~/startwm.sh` executable to make it work hehe

microsoft/hyper-v/default.nix
Mic92: Does xrdp maybe supports reading this configuration by supplying command line flags? Otherwise it might be possible to add configure flags in order to change sysconfdir (that's what it is called in autotools-based projects) to point to `/etc/xrdp` instead.

microsoft/hyper-v/default.nix
Mic92: I think we could still have a simple module that imports the right nixpkgs module. Just so that users can find easier find it, when browsing nixos-hardware.

Mic92:
Did open a PR in nixpkgs?

bltavares:
Hey there - I didn't have the time to keep looking at this further and did not open a PR on nixpkgs.

Mic92:
Please re-open if you want to continue to work on this.

PR #256, opened 2021-05-11
Add Raspberry Pi 4

domenkozar:


raspberry-pi/4/default.nix
```diff
@@ -0,0 +1,30 @@
+{ lib, pkgs, ...}:
+
+{
+  boot = {
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi4;
+    initrd.availableKernelModules = [ "usbhid" "usb_storage" "vc4" ];
+
+    # Needed for 4K displays
+    kernelParams = [ "cma=128M" ];
```
Mic92, line 9: Is this RAM reserved for graphic card?

raspberry-pi/4/default.nix
```diff
@@ -0,0 +1,30 @@
+{ lib, pkgs, ...}:
+
+{
+  boot = {
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi4;
+    initrd.availableKernelModules = [ "usbhid" "usb_storage" "vc4" ];
```
Mic92, line 6: Not sure about `vc4` but are the other modules not also included in `nixos-generate-config` output?

Mic92:
Please also reference this profile in README.md and `flake.nix`.

raspberry-pi/4/default.nix
```diff
@@ -0,0 +1,30 @@
...
+    # Needed for 4K displays
+    kernelParams = [ "cma=128M" ];
+
+    loader = {
+      raspberryPi = {
+        enable = true;
+        version = 4;
+        firmwareConfig = ''
+          dtparam=audio=on 
+          gpu_mem=192
```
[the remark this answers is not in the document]
domenkozar: I'd leave that to whoever wants to run it headless :)

raspberry-pi/4/default.nix
domenkozar: To be honest, I don't know, but searching through google most tutorials recommend it for a good default experience.

raspberry-pi/4/default.nix
domenkozar: I'd like for this file to be a substitute for `nixos-generate-config` so that there's one less step when installing NixOS.

cc @samueldr for `vc4`.

domenkozar: I've searched google and nixpkgs but couldn't find reference for that module, what hardware bit is it used for?

raspberry-pi/4/default.nix
```diff
@@ -0,0 +1,30 @@
...
+        '';
+      };
```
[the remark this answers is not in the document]
domenkozar: Alright, removed!

raspberry-pi/4/default.nix
domenkozar: RemoveD!

domenkozar:
@Mic92 @samueldr PTAL.

raspberry-pi/4/default.nix
samueldr: For the mainline kernel it's actually `pcie-brcmstb`/`pcie_brcmstb` (interchangeable with our tooling):

 - https://github.com/NixOS/nixpkgs/blob/65670311112f09f8e36b5ec714de0c222e61a09f/nixos/modules/profiles/all-hardware.nix#L81

Sorry I had the name wrong.

Though it's highly possible it't built as a built-in module for the vendor kernel.

raspberry-pi/4/default.nix
```diff
@@ -0,0 +1,18 @@
...
+    kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi4;
+    initrd.availableKernelModules = [ "usbhid" "usb_storage" "vc4" ];
+
+    loader = {
+      grub.enable = lib.mkDefault false;
+      generic-extlinux-compatible.enable = lib.mkDefault true;
+    };
+  };
+
+  nix.buildCores = 4;
```
Mic92, line 14: We no longer need this since by default it will use all cores already, see `man configuration.nix`:

```suggestion
```

domenkozar: How about we set `nix.maxJobs = 4;`?

Mic92: Also not needed. The current default for maxJobs in nixos is `auto`, which means it will also use 4 cores on raspberrypi

raspberry-pi/4/default.nix
domenkozar: Seems to be in already in the kernel.

domenkozar:
Thanks for the reviews @Mic92 & @samueldr, this is a lot shorter now :)

PR #259, opened 2021-05-11
friendlyarm/nanopc-t4: init

jakubgs:
As suggested by @Mic92 in https://github.com/NixOS/nixpkgs/pull/111034 I'm submitting [my fixes](https://github.com/jakubgs/nixos-nanopc-t4) for the [FriendlyARM NanoPC-T4](https://www.friendlyarm.com/index.php?route=product/product&product_id=225) device.

I'm not sure if forcing a specific kernel version is kosher, but on the other hand I don't know if my kernel config changes will work on any other version. I'm open to suggestions as to how to do this better.

Also unusure if I should call the folder `friendlyarm` or `friendlyelec`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ See code for all available configurations.
 | [Dell XPS 15 9560, nvidia only][] | `<nixos-hardware/dell/xps/15-9560/nvidia>`         |
 | [Dell XPS 15 9500][]              | `<nixos-hardware/dell/xps/15-9500>`                |
 | [Dell XPS 15 9500, nvidia][]      | `<nixos-hardware/dell/xps/15-9500/nvidia>`         |
+| FriendlyARM NanoPC-T4             | `<nixos-hardware/friendlyarm/nanopc-t4>`           |
 | [Google Pixelbook][]              | `<nixos-hardware/google/pixelbook>`                |
 | [Inverse Path USB armory][]       | `<nixos-hardware/inversepath/usbarmory>`           |
 | Lenovo IdeaPad Z510               | `<nixos-hardware/lenovo/ideapad/z510>`             |
@@ -113,7 +114,7 @@ See code for all available configurations.
 | Lenovo ThinkPad T440s             | `<nixos-hardware/lenovo/thinkpad/t440s>`           |
 | Lenovo ThinkPad T440p             | `<nixos-hardware/lenovo/thinkpad/t440p>`           |
 | Lenovo ThinkPad T450s             | `<nixos-hardware/lenovo/thinkpad/t450s>`           |
-| Lenovo ThinkPad T460              | `<nixos-hardware/lenovo/thinkpad/t460>`           |
+| Lenovo ThinkPad T460              | `<nixos-hardware/lenovo/thinkpad/t460>`            |
 | Lenovo ThinkPad T460s             | `<nixos-hardware/lenovo/thinkpad/t460s>`           |
 | Lenovo ThinkPad T470s             | `<nixos-hardware/lenovo/thinkpad/t470s>`           |
 | Lenovo ThinkPad T480s             | `<nixos-hardware/lenovo/thinkpad/t480s>`           |
--- a/friendlyarm/nanopc-t4/README.md
+++ b/friendlyarm/nanopc-t4/README.md
@@ -1,0 +1,91 @@
+# Description
+
+This document on how I configure [NixOS](https://nixos.org/) on [NanoPC-T4](https://wiki.friendlyarm.com/wiki/index.php/NanoPC-T4).
+
+# Installation
+
+To install NixOS on you can follow the [official instructions](https://nixos.wiki/wiki/NixOS_on_ARM/NanoPC-T4) and use the [pre-built images](https://github.com/tmountain/arch-nanopct4/tree/main/images/) from @tmountain. You can also build the U-Boot image yourself from `nixpkgs` based on changes added in [#111034](https://github.com/NixOS/nixpkgs/pull/111034).
+
+## NixOS on NVMe with ZFS
+
+It is possible to migrate the OS from the eMMC storage to the NVMe.
+
+In my case I migrated `/`, `/nix`, and `/home`, leaving `/boot` on the eMMC.
+
+Create the ZFS pool and three filesystems on the SSD:
+```sh
+zpool create -O xattr=sa -O acltype=posixacl -O mountpoint=none rpool /dev/nvme0n1
+for VOL in nix root home; do
+    zfs create -o mountpoint=legacy rpool/$VOL
+    mkdir /mnt/$VOL
+    mount.zfs rpool/$VOL /mnt/$VOL
+done
+```
+Then sync the original filesystem on eMMC to the new volumes:
+```sh
+rsync -rax /. /mnt/root
+rsync -rax /nix/. /mnt/nix
+rsync -rax /home/. /mnt/home
+rsync -rax /boot/. /
+```
+Afterwards create a configuration that looks like this:
+```nix
+{
+  # TODO: Make sure to update the eMMC device UUID!
+  fileSystems."/boot" = { device = "/dev/disk/by-uuid/1234-5678"; fsType = "ext4"; };
+  fileSystems."/" = { device = "rpool/root"; fsType = "zfs"; };
+  fileSystems."/nix" = { device = "rpool/nix"; fsType = "zfs"; };
+  fileSystems."/home" = { device = "rpool/home"; fsType = "zfs"; };
+}
+```
+And rebuild he system:
+```sh
+sudo nixos-rebuild boot
+```
+:warning: __IMPORTANT:__ After that it's necessary to run all the four `rsync` commands again to sync filesystems.
+
+Once everything is synced you can finally reboot.
+
+You should also clean up `/boot` afterwards.
+
+# UART Debug Console
+
+Device provides a Debug UART 4 Pin 2.54mm header connection, 3V level, 1500000bps.
+
+To connect to you will need a USB to UART converter/receiver that supports the speed of 1500000bps.
+
+The serial port parameters are [8-N-1](https://en.wikipedia.org/wiki/8-N-1).
+
+A reader using `CP2102` chip did not work but `FT232RL` works fine:
+
+![UART converter photo](https://github.com/NixOS/nixos-hardware/releases/download/not-a-release/FT232RL.jpg)
+
+You can use `minicom` or `picocom` to connect:
+```
+sudo minicom -b 1500000 -D /dev/ttyUSB0 
+sudo picocom -b 1500000 /dev/ttyUSB0
+```
+But you'll need to disable flow control with `Ctrl-A x`.
+
+Here is a good overview of UART USB-to-Serial adapters:
+
+* https://www.sjoerdlangkemper.nl/2019/03/20/usb-to-serial-uart/
+* https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT232R.pdf
+
+Pin layout where #4 is next to USB-C port:
+
+| Pin num.| #1  | #2 | #3 | #4 |
+|---------|-----|----|----|----|
+| Purpose | GND | V5 | TX | RX |
+
+Remember that the `TX` and `RX` ports should be swapped between UART adapter and the board.
+
+The V5 pin does not need to be connected if you're powering the board from another source.
+
+See the full board diagram for more details:
+
+![Board diagram](https://wiki.friendlyarm.com/wiki/images/b/bb/NanoPC-T4_1802_Drawing.png)
+
+You can access the recovery console by holding the __Recovery__ button and then pressing the __Power__ button.
+For this to work the device will have to be off, which requires holding the __Power__ button long enough.
+
--- a/friendlyarm/nanopc-t4/default.nix
+++ b/friendlyarm/nanopc-t4/default.nix
@@ -1,0 +1,26 @@
+{ lib, pkgs, ... }:
+
+{
+  boot.loader = {
+    grub.enable = lib.mkDefault false;
+    # Enables the generation of /boot/extlinux/extlinux.conf.
+    generic-extlinux-compatible.enable = lib.mkDefault true;
+  };
+
+  # UART debug console bitrates.
+  services.mingetty.serialSpeed = [ 1500000 115200 ];
+
+  # Enable additional firmware (such as Wi-Fi drivers).
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Fix for not detecting the M.2 NVMe SSD. Will cause recompilation.
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.10") (lib.mkDefault pkgs.linuxPackages_latest);
+  boot.kernelPatches = lib.mkDefault [{
+    name = "pcie-rockchip-config.patch";
+    patch = null;
+    extraConfig = ''
+      PHY_ROCKCHIP_PCIE y
+      PCIE_ROCKCHIP_HOST y
+    '';
+  }];
+}
--- a/flake.nix
+++ b/flake.nix
@@ -29,6 +29,7 @@
       dell-xps-15-9560-nvidia = import ./dell/xps/15-9560/nvidia;
       dell-xps-15-9500 = import ./dell/xps/15-9500;
       dell-xps-15-9500-nvidia = import ./dell/xps/15-9500/nvidia;
+      friendlyarm-nanopc-t4 = import ./friendlyarm/nanopc-t4;
       google-pixelbook = import ./google/pixelbook;
       inversepath-usbarmory = import ./inversepath/usbarmory;
       lenovo-ideapad-z510 = import ./lenovo/ideapad/z510;


jakubgs:
I also have some docs that I created for my own use, like the one for the UART console:
https://github.com/jakubgs/nixos-nanopc-t4/blob/master/UART

Or for using the NVMe for booting:
https://github.com/jakubgs/nixos-nanopc-t4#nixos-on-zfs

I can include them in this PR if anyone thinks it makes sense.

jakubgs:
I also have some docs that I created for my own use, like the one for the UART console:
https://github.com/jakubgs/nixos-nanopc-t4/blob/master/UART

Or for using the NVMe for booting:
https://github.com/jakubgs/nixos-nanopc-t4#nixos-on-zfs

I can include them in this PR if anyone thinks it makes sense.

friendlyarm/nanopc-t4/default.nix
```diff
@@ -0,0 +1,30 @@
...
+    extraConfig = ''
+      NVME_CORE y
+      BLK_DEV_NVME y
+      NVME_MULTIPATH y
+      PCIE_ROCKCHIP y
+      PCIE_ROCKCHIP_EP y
+      PCIE_ROCKCHIP_HOST y
+      PHY_ROCKCHIP_DP y
+      PHY_ROCKCHIP_PCIE y
+      PHY_ROCKCHIP_USB y
```
Mic92, line 27: Is this something we should just enable in our stock kernel by default?
Also I assume that some of these option can be build as module instead of a builtin module?

Mic92: cc @samueldr

README.md
```diff
@@ -94,6 +94,7 @@ See code for all available configurations.
 | [Dell XPS 15 9560, nvidia only][] | `<nixos-hardware/dell/xps/15-9560/nvidia>`         |
 | [Dell XPS 15 9500][]              | `<nixos-hardware/dell/xps/15-9500>`                |
 | [Dell XPS 15 9500, nvidia][]      | `<nixos-hardware/dell/xps/15-9500/nvidia>`         |
+| FriendlyARM NanoPC-T4             | `<nixos-hardware/friendlyarm/nanopc-t4>`           |
```
Mic92, line 97: Please also add an entry to flake.nix

friendlyarm/nanopc-t4/default.nix
```diff
@@ -0,0 +1,30 @@
...
+    grub.enable = lib.mkDefault false;
+    # Enables the generation of /boot/extlinux/extlinux.conf
+    generic-extlinux-compatible.enable = lib.mkDefault true;
+  };
+
+  # Enable additional firmware (such as Wi-Fi drivers).
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Fix for not detecting the M.2 NVMe SSD. Will cause recompilation.
+  boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_5_10;
```
Mic92, line 14: How about?

```suggestion
  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.10") lib.mkDefault pkgs.linuxPackages_latest;
```

Mic92: I don't think newer kernel would cause drivers to go away...

Mic92:
You can also put your writeups in a README.md in the nanopc subdirectory of this repository.

Mic92:
I guess something like this could be added for uart terminal?

```
  services.getty.serialSpeed = lib.mkDefault [ 1500000 115200 ];
```

friendlyarm/nanopc-t4/default.nix
jakubgs: Yes. The original list of modules I built into the kernel was much longer but I narrowed it down by trial and error. I bet I could get rid of a few more to know exactly which ones need to be built in. But I'd need to find the time to do that.

The annoying thing is that the UART console needs to beconnected so I can recover from a bad kernel.

friendlyarm/nanopc-t4/default.nix
jakubgs: Yeah, the config names MIGHT change, but it's unlikely. I'll do that.

README.md
jakubgs: Sure.

friendlyarm/nanopc-t4/default.nix
jakubgs: I will try to narrow it down, but this might take a few days, as one kernel compilation takes something like 4 hours each time.

jakubgs: I tested stock `pkgs.linuxPackages_5_11` but it fails to boot:
```
<<< NixOS Stage 1 >>>

loading module zfs...
loading module spl...
loading module dm_mod...
running udev...
Starting version 246
kbd_mode: KDSKBMODE: Inappropriate ioctl for device
Gstarting device mapper and LVM...
importing root ZFS pool "rpool"...............................................................
cannot import 'rpool': no such pool available
waiting for device /dev/disk/by-uuid/09a75307-8dda-4e21-9a91-709d0dc3338a to appear.......................
mounting rpool/root on /...
mount: mounting rpool/root on /mnt-root/ failed: No such file or directory
retrying...
...(omitted)...
mount: mounting rpool/root on /mnt-root/ failed: No such file or directory
retrying...
An error occurred in stage 1 of the boot process, which must mount the
root filesystem on `/mnt-root' and then start stage 2.  Press one
of the following keys:

  r) to reboot immediately
  *) to ignore the error and continue
```

jakubgs: As I said, almost 4 hours for kernel compilation:

image

jakubgs: Okay, I confirmed we don't need the `PHY_ROCKCHIP_DP` and `PHY_ROCKCHIP_USB` flags set to `y`.

jakubgs: Removing `PHY_ROCKCHIP_PCIE` seems to have broken boot:
```
[FAILED] Failed to start File Systeâ…f-dc5e-4b36-b356-e4782c20e5c7.
See 'systemctl status "systemd-fsck@devâ…\x2de4782c20e5c7.service"' for details.
[DEPEND] Dependency failed for /boot.
[DEPEND] Dependency failed for Local File Systems.
```
But in an unexpected way.

jakubgs: Dropping `PCIE_ROCKCHIP_EP` also worked...

Mic92: I think we can over-approximate kernel options and enable more than we need for this particular hardware.
The only thing we should be look into is using `m` for module instead of `y` for yes wherever possible.
Than we could have this in nixpkgs and people no longer need to compile their own kernel for this device.

friendlyarm/nanopc-t4/README.md
```diff
@@ -0,0 +1,91 @@
...
+|---------|-----|----|----|----|
+| Purpose | GND | V5 | TX | RX |
+
+Remember that the `TX` and `RX` ports should be swapped between UART adapter and the board.
+
+The V5 pin does not need to be connected if you're powering the board from another source.
+
+See the full board diagram for more details:
+
+![Board diagram](./images/NanoPC-T4_1802_Drawing.png)
```
Mic92, line 87: I uploaded images to a git tag in this repo. Can you link the images from there? 
https://github.com/NixOS/nixos-hardware/releases/download/not-a-release/FT232RL.jpg
https://github.com/NixOS/nixos-hardware/releases/download/not-a-release/NanoPC-T4_1802_Drawing.png

samueldr: Are those images properly licensed for use?

friendlyarm/nanopc-t4/default.nix
jakubgs: Well how about that. Looks the eMMC on my device is dying. I can't write to it. I just get endless streams of stuff like:
```
mmc2: running CQE recovery
```
and errors like:
```
blk_update_request: I/O error, dev mmcblk2, sector 34816 op 0x1:(WRITE) flags 0x8800 phys_seg 61 prio class 0
```
It seems like the eMMC is done for. Sad to see. Maybe I can somehow move the boot partition to a micro SD card...

jakubgs: Whelp, my NanoPC-T4 is borked because the eMMC is dead. I will have to somehow make it boot from the micro SD.

Maybe I'll have time to try that on the weekend. We'll see.

Mic92: Not sure about the nanopc but most arm chips will able to boot from multiple mediums.

friendlyarm/nanopc-t4/README.md
Mic92: Yes. Please check that as well. In doubt we should probably rather linking to them to avoid legal issues.

friendlyarm/nanopc-t4/default.nix
jakubgs: Yeah, it should work, as that's what the official docs do the installation: https://nixos.wiki/wiki/NixOS_on_ARM/NanoPC-T4

I will just have to do some kind of hybrid setup where the microSD is only a `/boot`, while the rest of the OS is on the NVMe.

jakubgs: Damn. The stock SD card image for NixOS doesn't detect the NVMe. So I have to compile my own with custom kernel options, but to do that I need an `aarch64` machine. But maybe it's possible to do this with a VM.

jakubgs: I know NixOS project has hydra servers for `aarch64` compilation. Would it be possible to build an SD card image like this:
https://hydra.nixos.org/job/nixos/release-20.09/nixos.sd_image.aarch64-linux
Just with these flags enabled?
```
NVME_CORE y
BLK_DEV_NVME y
NVME_MULTIPATH y
PCIE_ROCKCHIP y
```
As far as I know this is the bare minimum, but I didn't check any further since my eMMC failed.

Mic92: I can run a build if you give me the nixos configuration for that.

jakubgs: Thanks. I just need this added to it:
```nix
{
  boot.kernelPatches = [{
    name = "pcie-rockchip-config.patch";
    patch = null;
    extraConfig = ''
      NVME_CORE y
      NVME_MULTIPATH y
      BLK_DEV_NVME y
      PCIE_ROCKCHIP y
    '';
  }];

  services.mingetty.serialSpeed = [ 1500000 ];
}
```
So it detects the NVMe and UART works. I can manage the rest to make the SD card `/boot`.

Mic92: http://thalheim.io/nixos-sd-image-21.05pre-git-aarch64-linux.img.zst

Mic92: Build by `nixos-generate -f sd-aarch64-installer  -c foo.nix`

jakubgs: Thanks a bunch. I'll try it out today or tomorrow.

jakubgs: Looks like I was wrong and that's not the full list. I booted your image and I'm missing the NVMe.
```
[nixos@nixos:~]$ zcat /proc/config.gz | grep -E '(NVME_CORE|NVME_MULTIPATH|BLK_DEV_NVME|PCIE_ROCKCHIP)'
CONFIG_PCIE_ROCKCHIP=y
CONFIG_PCIE_ROCKCHIP_HOST=m
# CONFIG_PCIE_ROCKCHIP_EP is not set
CONFIG_NVME_CORE=y
CONFIG_BLK_DEV_NVME=y
CONFIG_NVME_MULTIPATH=y

[nixos@nixos:~]$ zcat /proc/config.gz | grep PHY_ROCKCHIP_PCIE
CONFIG_PHY_ROCKCHIP_PCIE=m
```
My best guess is that we're missing `PHY_ROCKCHIP_PCIE`. I think my eMMC issues cause some of my tests to not take effect, which mislead me as to what is necessary and what is not.

I will try to build a new image using your recovery image locally, we'll see how that works.

jakubgs: I have 3.4 GB on my SD card. We'll see if that enough to compile a custom kernel.

jakubgs:
```
cp: error writing '/nix/store/qcpddd0c0lb0dba34kwb5xr7x6b05hbw-linux-5.10.37/System.map': No space left on device
```
I guess that wasn't enough space...

Mic92: Should I rebuild with `PHY_ROCKCHIP_PCIE`?

jakubgs: I'm trying on my Raspberry Pi 4B with a bigger SD card. It's been compiling for 6 hours now and still going. Might work.

Thanks for asking tho.

jakubgs: I've re-compiled the image on my Rpi 4b with these flags:
```
NVME_CORE y
NVME_MULTIPATH y
BLK_DEV_NVME y
PCIE_ROCKCHIP y
PCIE_ROCKCHIP_EP y
PCIE_ROCKCHIP_HOST y
PHY_ROCKCHIP_PCIE y
```
And I can finally see the NVMe:
```
[0;nixos@nixos: ~nixos@nixos:~]$ sudo fdisk -l /dev/nvme0n1
Disk /dev/nvme0n1: 232.89 GiB, 250059350016 bytes, 488397168 sectors
Disk model: Samsung SSD 970 EVO Plus 250GB          
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 7BF4B90B-D5E8-2A4B-9D05-08A6D33AEF3E

Device             Start       End   Sectors   Size Type
/dev/nvme0n1p1      2048 488380415 488378368 232.9G Solaris /usr & Apple ZFS
/dev/nvme0n1p9 488380416 488396799     16384     8M Solaris reserved 1
```

jakubgs: I tried what was suggested in https://github.com/NixOS/nixpkgs/pull/111034#issuecomment-846134037 and used:
```
boot.initrd.kernelModules = [ "nvme-core" "nvme" "pcie-rockchip" "pcie-rockchip-host" "phy-rockchip-pcie" ];
```
And it doesn't boot. So it seems like something has to be baked into the kernel itself.

Mic92: I think `PCIE_ROCKCHIP_EP` needs to be enabled since it's not a kernel module but a compilation option. I checked the others and they are just kernel modules and already enabled? So is it maybe just this one option that is needed?

samueldr: (Before looking more into it) We might want to enable it outright for the NixOS kernels too

 - https://cateee.net/lkddb/web-lkddb/PCIE_ROCKCHIP_EP.html

> Say Y here if you want to support Rockchip PCIe controller in endpoint mode on Rockchip SoC. There is 1 internal PCIe port available to support GEN2 with 4 slots.

 - https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/pci/rockchip-pcie-ep.txt

* * *

I don't expect this config option to do anything.

```
~/tmp/linux/linux $ ag rockchip,rk3399-pcie-ep
drivers/pci/controller/pcie-rockchip-ep.c
560:    { .compatible = "rockchip,rk3399-pcie-ep"},

Documentation/devicetree/bindings/pci/rockchip-pcie-ep.txt
4:- compatible: Should contain "rockchip,rk3399-pcie-ep"
41:     compatible = "rockchip,rk3399-pcie-ep";
```

No device tree in mainline Linux expose the required property for this to be used.

Mic92: https://github.com/NixOS/nixpkgs/pull/124450

Mic92: @samueldr are missing device tree files maybe also the reason why simply adding the kernel modules is not enough? Maybe explicit modprobe commands are required.

samueldr: The `compatible` here means "look for a node with this identifier".

 - https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/drivers/pci/controller/pcie-rockchip-ep.c#L645-L651

I don't know why this driver is present in the kernel sources if no device tree seemingly uses this. I don't know what it does really. I don't know what "There is 1 internal PCIe port available to support GEN2 with 4 slots" really means here, the description lacks context.

jakubgs: I actually built a kernel with just these options without `PCIE_ROCKCHIP_EP` and it booted from the NVMe fine:
```
NVME_CORE y
PHY_ROCKCHIP_PCIE y
PCIE_ROCKCHIP y
PCIE_ROCKCHIP_HOST y
```
So it doesn't seem to be necessary.

samueldr: right, so at a glance these are all already part of the default kernel build

```
CONFIG_NVME_CORE=m
CONFIG_PHY_ROCKCHIP_PCIE=m
CONFIG_PCIE_ROCKCHIP=y
CONFIG_PCIE_ROCKCHIP_HOST=m
# CONFIG_PCIE_ROCKCHIP_EP is not set
```

So I guess `avialableKernelModules` would need to be:
```
[
  "nvme-core"
  "phy-rockchip-pcie"
 "pcie-rockchip-host"
]
```

Mic92: > right, so at a glance these are all already part of the default kernel build
> 
> ```
> CONFIG_NVME_CORE=m
> CONFIG_PHY_ROCKCHIP_PCIE=m
> CONFIG_PCIE_ROCKCHIP=y
> CONFIG_PCIE_ROCKCHIP_HOST=m
> # CONFIG_PCIE_ROCKCHIP_EP is not set
> ```
> 
> So I guess `availableKernelModules` would need to be:
> 
> ```
> [
>   "nvme-core"
>   "phy-rockchip-pcie"
>   "pcie-rockchip-host"
> ]
> ```

I think @jakubgs already tried adding those modules but it did not work. Could it be that the kernel does not load the available kernel modules correctly? Is maybe an explicit modprobe required?

samueldr: What *could* be happening is that there are *regulators* or other similar kernel modules needed.

The better way to look into that is to boot the generic linux image from SD entirely, verify NVMe is available, then dump the output from `lsmod`. That should list all modules the system is currently using.

Then, from that list, start adding/removing, starting intelligently like removing everything network-related.

That's how I've been handling the support for USB in stage-1 in the generic NixOS image. Some drivers are required to enable some hardware for another driver, but are not dependencies of that driver. Thy end up being linked by the device tree telling e.g. "to enable hardware X I need to control other hardware Y". Which sucks to track down.

jakubgs: In https://github.com/NixOS/nixos-hardware/pull/259#discussion_r640028223 I've identified 4 configs that are required.

After dropping `PCIE_ROCKCHIP_HOST` from the config and narrowing it down to just this and it stopped booting:
```nix
{
  boot.kernelPatches = [{
    name = "pcie-rockchip-config.patch";
    patch = null;
    extraConfig = ''
      NVME_CORE y
      PHY_ROCKCHIP_PCIE y
      PCIE_ROCKCHIP y
    '';
  }];
}
```
After that I've added `pcie-rockchip-host` to `initrd` modules:
```nix
{
  boot.initrd = {
    enable = true;
    kernelModules = [ "pcie-rockchip-host" ];
  };
}
```
But that didn't work, and failed to boot as well. So it appears that `PCIE_ROCKCHIP_HOST` needs to be baked in.

Unless I'm using the wrong module name, but I don't think I am:
https://cateee.net/lkddb/web-lkddb/PCIE_ROCKCHIP_HOST.html

jakubgs: Here's part of the log from an attempt to boot without `PCIE_ROCKCHIP_HOST`:
```
<<< NixOS Stage 1 >>>

loading module pcie-rockchip-host...
[    3.677351] rockchip-pcie f8000000.pcie: PCIe link training gen1 timeout!
loading module zfs...
loading module spl...
loading module dm_mod...
running udev...
Starting version 246
kbd_mode: KDSKBMODE: Inappropriate ioctl for device
Gstarting device mapper and LVM...
importing root ZFS pool "rpool"...............................................................
cannot import 'rpool': no such pool available
```

jakubgs: Based on this comment - https://github.com/rockchip-linux/kernel/issues/116#issuecomment-443336998 -I believe the timeout the error is referring to is this one:
https://github.com/torvalds/linux/blob/v5.11/drivers/pci/controller/pcie-rockchip-host.c#L332-L335

Which appears to be a hardcoded 500 miliseconds.

jakubgs: Just for comparison, with `PCIE_ROCKCHIP_HOST` baked in this is what kernel log shows:
```
[    2.685518] rockchip-pcie f8000000.pcie: host bridge /pcie@f8000000 ranges:
[    2.685566] rockchip-pcie f8000000.pcie:      MEM 0x00fa000000..0x00fbdfffff -> 0x00fa000000
[    2.685589] rockchip-pcie f8000000.pcie:       IO 0x00fbe00000..0x00fbefffff -> 0x00fbe00000
[    2.686294] rockchip-pcie f8000000.pcie: no vpcie12v regulator found
[    3.029951] rockchip-pcie f8000000.pcie: host bridge /pcie@f8000000 ranges:
[    3.029980] rockchip-pcie f8000000.pcie:      MEM 0x00fa000000..0x00fbdfffff -> 0x00fa000000
[    3.029995] rockchip-pcie f8000000.pcie:       IO 0x00fbe00000..0x00fbefffff -> 0x00fbe00000
[    3.030348] rockchip-pcie f8000000.pcie: no vpcie12v regulator found
[    3.069821] rockchip-pcie f8000000.pcie: PCI host bridge to bus 0000:00
```

samueldr: > Unless I'm using the wrong module name, but I don't think I am:

You are, and with NixOS's handling of modules names `pcie_rockchip_host` would have resolved to the same module.

This is odd. Sorry I don't have much more to say at this point.

I'll look "soon" (can't say when really) at my RK3399 device, and make it use the strictly mainline as-compiled-by-NixOS kernel and see if I can make it work.

Mic92: I guess for now we could merge the kernel configuration that you found into nixos hardware until we find a better fix.

jakubgs: Yeah, I'm doing one last build which should finish in 2 hours to test if `PHY_ROCKCHIP_PCIE` works by itself.
If not I will update this PR with just the two modules enabled, and of course update image links too.

jakubgs: Yuuuup, I've verified it doesn't work with `PHY_ROCKCHIP_PCIE` by itself. I'm going to update the PR.

friendlyarm/nanopc-t4/README.md
jakubgs: Done.

jakubgs:
Thanks for the fix.

Issue #260, opened 2021-05-12
recent nixos master fails to build due to tlp setting


wmertens:
See https://github.com/NixOS/nixpkgs/pull/97972#issuecomment-834774554

Not sure if this should be solved here or in nixpkgs.

Mic92:
Maybe would enable `config.services.tlp.enable` only if `services.power-profiles-daemon.enable ` is not enabled?

Mic92:
Right now this is only an issue if the user has enabled power-profiles-daemon them-self, right?

terlar:
That is partly true, or if they are using something that enables it, for example GNOME. Found in the linked issue:
https://github.com/NixOS/nixpkgs/commit/316928e8c15994053fb72252418d2b3197cb5dc1

I am one of the latter and not sure I would have figured it out without these two issues.

benley:
Pretty sure this is fixed by #263

PR #263, opened 2021-05-26
Tlp and power profiles daemon clash

bryanhonof:
Gnome 40 now uses [power-profile-daemon](https://gitlab.freedesktop.org/hadess/power-profiles-daemon ),which clashes with tlp.
This check will disable tlp whenever it finds that the
power-profiles-daemon is activated. This should also
fix the issue mentioned in #260.

I'm not 100% sure if this is "the way" to do it, but it fixes the problem
on my end. For reference, I'm using the Lenovo Thinkpad T495.

Also, this will be my first pull request to this project. So if I'm doing
anything wrong, please let me know :).

### DIFF
--- a/common/pc/laptop/default.nix
+++ b/common/pc/laptop/default.nix
@@ -6,9 +6,5 @@
   # Gnome 40 introduced a new way of managing power, without tlp.
   # However, these 2 services clash when enabled simultaneously.
   # https://github.com/NixOS/nixos-hardware/issues/260
-  services.tlp.enable = lib.mkDefault
-    (if config.services.power-profiles-daemon.enable == true then
-      false
-    else
-      true);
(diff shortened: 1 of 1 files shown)


common/pc/laptop/default.nix
```diff
@@ -3,5 +3,12 @@
...
 
-  services.tlp.enable = lib.mkDefault true;
+  # Gnome 40 introduced a new way of managing power, without tlp.
+  # However, these 2 services clash when enabled simultaneously.
+  # https://github.com/NixOS/nixos-hardware/issues/260
+  services.tlp.enable = lib.mkDefault
+    (if config.services.power-profiles-daemon.enable == true then
+      false
+    else
+      true);
```
Mic92, line 13:
```suggestion
  services.tlp.enable = lib.mkDefault !config.services.power-profiles-daemon.enable;
```

PR #272, opened 2021-06-03
raspberry-pi/4: Add poe-hat option

walkah:
This enables support for official [PoE Hat](https://www.raspberrypi.org/products/poe-hat/) for raspberry pi 4. Specifically, it adds controls for the included fans / cooling device. It adds a module option:

```
hardware.raspberry-pi."4".poe-hat.enable = true;
```

### DIFF
--- a/raspberry-pi/4/poe-hat.nix
+++ b/raspberry-pi/4/poe-hat.nix
@@ -1,0 +1,122 @@
+{ config, lib, pkgs, ... }:
+
+let 
+  cfg = config.hardware.raspberry-pi."4".poe-hat;
+in {
+  options.hardware = {
+    raspberry-pi."4".poe-hat = {
+      enable = lib.mkEnableOption ''
+        Enable support for the Raspberry Pi POE Hat.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    # Configure for modesetting in the device tree
+    hardware.deviceTree = {
+      overlays = [
+        # Equivalent to: https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
+        {
+          name = "rpi-poe-overlay";
+          dtsText = ''
+            /*
+            * Overlay for the Raspberry Pi POE HAT.
+            */
+            /dts-v1/;
+            /plugin/;
+
+            / {
+              compatible = "brcm,bcm2711";
+
+              fragment@0 {
+                target-path = "/";
+                __overlay__ {
+                  fan0: rpi-poe-fan@0 {
+                    compatible = "raspberrypi,rpi-poe-fan";
+                    firmware = <&firmware>;
+                    cooling-min-state = <0>;
+                    cooling-max-state = <4>;
+                    #cooling-cells = <2>;
+                    cooling-levels = <0 1 10 100 255>;
+                    status = "okay";
+                  };
+                };
+              };
+
+              fragment@1 {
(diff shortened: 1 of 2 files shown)


raspberry-pi/4/poe-hat.nix
```diff
@@ -0,0 +1,122 @@
...
+
+              __overrides__ {
+                poe_fan_temp0 =		<&trip0>,"temperature:0";
+                poe_fan_temp0_hyst =	<&trip0>,"hysteresis:0";
+                poe_fan_temp1 =		<&trip1>,"temperature:0";
+                poe_fan_temp1_hyst =	<&trip1>,"hysteresis:0";
+                poe_fan_temp2 =		<&trip2>,"temperature:0";
+                poe_fan_temp2_hyst =	<&trip2>,"hysteresis:0";
+                poe_fan_temp3 =		<&trip3>,"temperature:0";
+                poe_fan_temp3_hyst =	<&trip3>,"hysteresis:0";
```
samueldr, line 114: These are `dtparam` options; e.g. `dtparam=poe_fan_temp0=XXX,poe_fan_temp0_hyst=YYY` etc...

Ideally we'd make them options.

The way this maps is:

```dts
// param_name = <&label>,"property   :offset""
poe_fan_temp0 = <&trip0>,"temperature:0";

trip0: trip0 {
	temperature = <40000>;  <- "0th" item (offset) of property temperature, of node with label trip0
	hysteresis = <2000>;
	type = "active";
};
```

https://www.raspberrypi.org/documentation/configuration/device-tree.mdh

So here it's 4 pairs of params to control the four trip points.

Mic92: Is it safe to change the defaults? Do we want to expose it (with a warning)?

samueldr: No idea.

Only thing I know is that the documentation is here:

  - https://github.com/raspberrypi/linux/blob/bcb79983422d056ea85d2b4a515762c0d56690e5/arch/arm/boot/dts/overlays/README#L2487-L2505

The parameters are not hidden piece of information. But there's no in-depth details either.

walkah: Yeah, I don't know for sure, but doing some searching I've seen references to tuning the params (e.g. https://www.jeffgeerling.com/blog/2021/taking-control-pi-poe-hats-overly-aggressive-fan ) so it seems like a safe assumption. I'm using this code on a set of four rpi4s and am happy with it, but I can update to include the param options.

Mic92:
@samueldr do you want to see having these options first or is this PR good to merge?

raspberry-pi/4/poe-hat.nix
```diff
@@ -0,0 +1,122 @@
+{ config, lib, pkgs, ... }:
+
+let 
+  cfg = config.hardware.raspberry-pi."4".poe-hat;
+in {
+  options.hardware = {
+    raspberry-pi."4".poe-hat = {
+      enable = lib.mkEnableOption ''
+        Enable support for the Raspberry Pi POE Hat.
+      '';
```
tfc, line 10: `lib.mkEnableOption` prepends the string "Whether to enable " to the string that is provided as function argument. ([source](https://github.com/NixOS/nixpkgs/blob/master/lib/options.nix#L97))

This way we will get `"Weather to enable Enable support for the ..."`.

I suggest `"support for the Raspberry Pi POE Hat."`

Mic92:
@samueldr second friendly ping.

PR #273, opened 2021-06-06
Add onenetbook/4

t184256:
One Netbook 4 has a buggy stylus driver.
This submission compiles a patched one as an alternative kernel module and blacklists the mainline one.
No idea how idiomatic this is, suggestions are welcome.

### DIFF
--- a/onenetbook/4/default.nix
+++ b/onenetbook/4/default.nix
@@ -1,0 +1,21 @@
+{ pkgs, config, ... }:
+
+{
+  imports = [
+    ../../common/cpu/intel
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+  ];
+
+  # OneNetbook 4 has `GXTP7386:00 27C6:011A Stylus` exporting no buttons in 5.12
+  # and libinput does't consider it a tablet without them, but a touchscreen.
+  # This leads to real weird effects,
+  # starting from clicking as soon as the pen gets tracked.
+  # A kernel patch exists to resolve that, compiled as an out-of-tree module here
+  # to avoid recompiling the kernel for such a small change.
+  # `hid-multitouch-onenetbook4` is the fixed one, don't use `hid-multitouch`.
+  boot.blacklistedKernelModules = [ "hid-multitouch" ];
+  boot.extraModulePackages = [
+    (config.boot.kernelPackages.callPackage ./goodix-stylus-mastykin {})
+  ];
+}
--- a/onenetbook/4/default.nix
+++ b/onenetbook/4/default.nix
@@ -1,5 +1,9 @@
-{ pkgs, config, ... }:
+{ pkgs, config, lib, ... }:
 
+let
+  iio-sensor-proxy-supports-rotation =
+    lib.versionAtLeast pkgs.iio-sensor-proxy.version "3.0";
+in
 {
   imports = [
     ../../common/cpu/intel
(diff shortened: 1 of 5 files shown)


samueldr:
Closed by mistake?

t184256:
yep, will re-open

t184256:
also added whatevver fixes the accelerometer for me, also not sure how acceptable is that

onenetbook/4/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  nixpkgs.overlays = [
+    (self: super: {
+      iio-sensor-proxy = super.iio-sensor-proxy.overrideAttrs (oa: rec {
+        version = "3.0";
+        src = pkgs.fetchgit {
+          url = "https://gitlab.freedesktop.org/hadess/iio-sensor-proxy";
+          rev = version;
+          sha256 = "sha256-ibxA0xVsm+wVYrlXVOfaNypmmIHykGtoHZHJNXf461k=";
+        };
+      });
```
samueldr, line 40: I'm pretty sure we don't want to include such updates in the nixos-hardware repo. At the very least not without *also* providing the proper update in Nixpkgs. I opened the upgrade here https://github.com/NixOS/nixpkgs/pull/125989 as @r-ryantm won't pick it up as it is.

Though we *might* want to include this for current-stable (and current-unstable)... Though most likely we want to check with `lib.versionOlder super.iio-sensor-proxy.version "3.0"` and conditionally add the update. This way further upgrades are not clobbered by this.

Furthermore, we should use `fetchFromGitLab` here. As it uses similar semantics to `fetchFromGitHub`, which in turn prevents loading the whole git history.

samueldr (changes requested):

samueldr:
Heh, I was already writing about that before you asked as I spotted the package upgrade :).

Let's see what other maintainers think of the upgrade (assuming it'll be conditional to versionOlder).

onenetbook/4/default.nix
t184256: Ack, will look into fetchGitLab and conditioning this tomorrow.

t184256: I've slept over it and I agree that this repository should rather deal with configuration, not packaging. I think there's very little harm in conditioning the rotation on the software version + just suggesting how to update. PR updated (e138a81).

Plus, this isn't exactly wildly popular hardware, the amount of people inconvenienced with this is gonna be single digits at best, most likely zero. I think updating software through this repo still might make sense in some rare cases, but it should be reserved to smth like mitigating regressions from bad firmware updates, not something as puny as this case.

t184256:
Got any comments on patching a kernel module instead of patching the kernel part? It works, recompiles relatively quickly and looks similarly forward-compatible to me, but I might've got overly inventive about it.

samueldr:
Only comment is that it's a good example of doing exactly that, AFAICT. I'll try to remember if/whenever I need to "just patch a module" :).

t184256:
```
# Taken from https://www.spinics.net/lists/linux-input/msg72861.html,
# ID replaced to one installed in OneNetbook 4
```

I'm not aware where else can it be obtained from.

onenetbook/4/goodix-stylus-mastykin/default.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  postUnpack = ''
+    tar -C goodix-stylus-mastykin \
+      --strip-components=3 -xf ${kernel.src} --wildcards \
+      '*/drivers/hid/hid-ids.h' '*/drivers/hid/hid-multitouch.c'
+  '';
+
+  postPatch = ''
+    mv hid-multitouch.c hid-multitouch-onenetbook4.c
+    patch hid-multitouch-onenetbook4.c goodix-stylus-mastykin.patch
```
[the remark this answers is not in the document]
t184256:
```
# Taken from https://www.spinics.net/lists/linux-input/msg72861.html,
```

I'm not aware where else can it be obtained from.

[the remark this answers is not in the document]
samueldr: Some kernel mailing lists archive interfaces are terrible. (Most?)

Anyway, `marc.info` has an interesting feature, once you find the same mail... Which might be harder since it seems it's not indexed on big search engines :(.

The mail:

 - https://marc.info/?l=linux-input&m=161847127221531&

You should see a link with the name `["commit-b7ee3d7" (text/plain)]`

 - https://marc.info/?l=linux-input&m=161847127221531&q=p3

Plain text commit that can be `git am`'d or `fetchpatch`'d!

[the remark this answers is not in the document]
t184256: > Some kernel mailing lists archive interfaces are terrible. (Most?)

Can't argue with that, with the previous one I haven't even recognized that patches were not inline.

I've updated it to use fetchpatch (907f675) from the urls you've provided and no longer carry the patches.

Mic92:
I am closing this to hoping it retriggers a build.

onenetbook/4/default.nix
```diff
@@ -0,0 +1,52 @@
...
+
+  # OneNetbook 4 has an accelerometer,
+  hardware.sensor.iio.enable = lib.mkDefault iio-sensor-proxy-supports-rotation;
+  # said accelerometer needs rotation, rotation needs iio-sensor-proxy >= 3.0
+  services.udev.extraHwdb = lib.mkIf iio-sensor-proxy-supports-rotation ''
+    acpi:BOSC0200:BOSC0200:*
+     ACCEL_MOUNT_MATRIX=0, 1, 0; 0, 0, 1; 1, 0, 0
+  '';
+  # (this at least gets normal/left-up/right-up/bottom-up right)
+  # Until https://github.com/NixOS/nixpkgs/pull/125989 reaches you, you can use
```
Mic92, line 34: Since https://github.com/NixOS/nixpkgs/pull/125989 do you think we can remove this?

Mic92:
@samueldr good to go?

samueldr:
Yeah, the package upgrade isn't there anymore, I haven't looked in-depth at the other changes, and changes since, but a quick glance sees nothing wrong.

onenetbook/4/default.nix
t184256: While it's merged into master, 21.05 doesn't have it. I'd keep it if you ask me.

t184256:
rebased, conditionally patched the stylus patch for 5.12.12

PR #276, opened 2021-06-11
Remove intel drivers from nvidia-disable

thiagokokada:
Two reasons for this change:
- `intel` drivers are not updated from quite a long time (since ~2019), and `modesetting` is the preferred one for Intel iGPUs.
- Technically you may want to disable NVIDIA GPUs on laptops with AMD processors too.

We don't want to pin `modesetting` here because it is already the default since PR: https://github.com/NixOS/nixpkgs/pull/111551.


thiagokokada:
BTW, this also fixed composited desktop (e.g.: using `picom`) on my Dell Precision 5550 laptop (with `intel` drivers it was laggy and broken, with `modesetting` it is now working as it should).

Mic92:
Sounds reasonable.

PR #278, opened 2021-06-19
Dell Precision 5530: fix screen flicker

symphorien:
Fix flicker with a workaround found on the arch wiki. I have been testing it for a month.
Also remove one kernel parameter from my previous PR which is not required (and actually has nothing to do with nouveau). I took it from the issue I linked in comment without thinking very long...


symphorien:
test failure seems unrelated

PR #283, opened 2021-06-30
Add L13 Yoga

vherrmann:
This is more or less the combination of the config for L13 and X13 Yoga.


lenovo/thinkpad/l13-yoga/default.nix
```diff
@@ -0,0 +1,15 @@
...
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop/acpi_call.nix
+  ];
+
+  services.throttled.enable = lib.mkDefault true;
+
+  # automatic screen orientation
+  hardware.sensor.iio.enable = true;
+
+  services.xserver.wacom.enable = lib.mkDefault config.services.xserver.enable;
```
Mic92, line 14: Please move this to `../yoga.nix` and import this here and in `x13-yoga` instead.

lenovo/thinkpad/l13-yoga/default.nix
```diff
@@ -0,0 +1,15 @@
+{ nixos, lib, pkgs, config, stdenv, ... }:
+{
+  imports = [
+    ../.
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop/acpi_call.nix
```
Mic92, line 6: After https://github.com/NixOS/nixos-hardware/pull/285 you can just import l13 instead.
Actually I would move this profile to `l13/yoga.nix` instead.

vherrmann: Should I move x13-yoga/default.nix to x13/yoga.nix as well?

lenovo/thinkpad/l13-yoga/default.nix
vherrmann: I think I'll only move `hardware.sensor…`, since there is no guaranty all thinkpad-yogas will have a wacom touchscreen.

README.md
```diff
@@ -102,6 +102,7 @@ See code for all available configurations.
 | [Lenovo ThinkPad E470](lenovo/thinkpad/e470)                        | `<nixos-hardware/lenovo/thinkpad/e470>`            |
 | [Lenovo ThinkPad E495](lenovo/thinkpad/e495)                        | `<nixos-hardware/lenovo/thinkpad/e495>`            |
 | [Lenovo ThinkPad L13](lenovo/thinkpad/l13)                          | `<nixos-hardware/lenovo/thinkpad/l13>`             |
+| [Lenovo ThinkPad L13 Yoga](lenovo/thinkpad/l13-yoga)                | `<nixos-hardware/lenovo/thinkpad/l13-yoga>`                           |
```
Mic92, line 105:
```suggestion
| [Lenovo ThinkPad L13](lenovo/thinkpad/l13)                          | `<nixos-hardware/lenovo/thinkpad/l13>`             |
| [Lenovo ThinkPad L13 Yoga](lenovo/thinkpad/l13-yoga)                | `<nixos-hardware/lenovo/thinkpad/l13-yoga>`        |
```

lenovo/thinkpad/l13-yoga/default.nix
Mic92: I think we should actually always make a directory, so `l13/yoga/default.nix`.
`x13-yoga/default.nix` could be moved to `x13/yoga` but with a symlink `x13-yoga` -> `x13/yoga` for backwards compatibility.

lenovo/thinkpad/l13-yoga/default.nix
Mic92: ok. Let's keep it for now.

lenovo/thinkpad/l13-yoga/default.nix
vherrmann: ok, I don't think there's a need for backwardcompability, since nearly everyone will use the nixosmodules in flake.nix, or not?

Mic92: I think the majority of nixos users are still using nix stable.

vherrmann: Oh, sorry I forgot that it's possible to use nixos-hardware without flakes.

Mic92:
The symlink is absolute. Please do `ln -sf x13/yoga x13/yoga` in `lenovo/thinkpad` to get a relative symlink not containing your home directory.

PR #291, opened 2021-07-21
Dell XPS 17 9700

### LINKED ISSUE #290: Dell XPS 9700 (17) Support
Audio does not currently work, and would seem to require some specialized configuration.
See https://github.com/NixOS/nixpkgs/issues/130882

Would be nice to have some working config for this machien as well.

IvanMalison:
Not ready to go, but thought I'd put what I have up here just in case anyone wants to use it as a starting point.

Fixes #290

### DIFF
--- a/dell/xps/17-9700/README.wiki
+++ b/dell/xps/17-9700/README.wiki
@@ -54,7 +54,7 @@ the firmware setup, so this must be done by hand.
 These settings are needed both for booting the final install, and installer
 itself. Therefore, they must be done first.
 
-* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as
   changing any other simple setting.
 
 * ''Disable Intel hardware RAID and use AHCI instead.'' Intel doesn't seem to
--- a/dell/xps/17-9700/README.wiki
+++ b/dell/xps/17-9700/README.wiki
@@ -1,0 +1,62 @@
+= Dell XPS 17 9700 =
+
+== Tested Hardware ==
+CPU: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
+Screen: 17" FHD(1920x1200) InfinityEdge
+
+00:00.0 Host bridge: Intel Corporation 10th Gen Core Processor Host Bridge/DRAM Registers (rev 02)
+00:01.0 PCI bridge: Intel Corporation 6th-10th Gen Core Processor PCIe Controller (x16) (rev 02)
+00:02.0 VGA compatible controller: Intel Corporation CometLake-H GT2 [UHD Graphics] (rev 05)
+00:04.0 Signal processing controller: Intel Corporation Xeon E3-1200 v5/E3-1500 v5/6th Gen Core Processor Thermal Subsystem (rev 02)
+00:08.0 System peripheral: Intel Corporation Xeon E3-1200 v5/v6 / E3-1500 v5 / 6th/7th/8th Gen Core Processor Gaussian Mixture Model
+00:12.0 Signal processing controller: Intel Corporation Comet Lake PCH Thermal Controller
+00:13.0 Serial controller: Intel Corporation Device 06fc
+00:14.0 USB controller: Intel Corporation Comet Lake USB 3.1 xHCI Host Controller
+00:14.2 RAM memory: Intel Corporation Comet Lake PCH Shared SRAM
+00:14.3 Network controller: Intel Corporation Comet Lake PCH CNVi WiFi
+00:15.0 Serial bus controller [0c80]: Intel Corporation Comet Lake PCH Serial IO I2C Controller #0
+00:15.1 Serial bus controller [0c80]: Intel Corporation Comet Lake PCH Serial IO I2C Controller #1
+00:16.0 Communication controller: Intel Corporation Comet Lake HECI Controller
+00:1b.0 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #17 (rev f0)
+00:1c.0 PCI bridge: Intel Corporation Device 06b8 (rev f0)
+00:1c.4 PCI bridge: Intel Corporation Device 06bc (rev f0)
+00:1d.0 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #9 (rev f0)
+00:1f.0 ISA bridge: Intel Corporation Device 068e
+00:1f.3 Multimedia audio controller: Intel Corporation Comet Lake PCH cAVS
+00:1f.4 SMBus: Intel Corporation Comet Lake PCH SMBus Controller
+00:1f.5 Serial bus controller [0c80]: Intel Corporation Comet Lake PCH SPI Controller
+01:00.0 3D controller: NVIDIA Corporation TU117M [GeForce GTX 1650 Ti Mobile] (rev ff)
+02:00.0 Non-Volatile memory controller: SK hynix Device 1639
+03:00.0 Unassigned class [ff00]: Realtek Semiconductor Co., Ltd. RTS5260 PCI Express Card Reader (rev 01)
+04:00.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+05:00.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+05:01.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+05:02.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+05:04.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+06:00.0 System peripheral: Intel Corporation JHL7540 Thunderbolt 3 NHI [Titan Ridge 4C 2018] (rev 06)
+3a:00.0 USB controller: Intel Corporation JHL7540 Thunderbolt 3 USB Controller [Titan Ridge 4C 2018] (rev 06)
+6e:00.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+6f:00.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+6f:01.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+6f:02.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+6f:04.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge 4C 2018] (rev 06)
+70:00.0 System peripheral: Intel Corporation JHL7540 Thunderbolt 3 NHI [Titan Ridge 4C 2018] (rev 06)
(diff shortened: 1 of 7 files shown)


dell/xps/17-9700/default.nix
```diff
@@ -0,0 +1,18 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    # To just use Intel integrated graphics with Intel's open source driver
+    ../../../common/gpu/nvidia-disable
```
Mic92, line 8: This maybe should be commented out or a subprofile.

dell/xps/17-9700/default.nix
```diff
@@ -0,0 +1,18 @@
...
+  ];
+
+  # TODO: boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
+  boot.kernelParams = lib.mkDefault [ "acpi_rev_override" ];
```
workflow, line 14: Is that for the NVidia?

IvanMalison:
So thanks to @workflow, it seems that we have figured out how to get audio working for this laptop. Unfortunately, his fix seems to require a kernel at 5.11 or newer, but the current nixos lts is only at 5.10. However, we should ALSO be able to get it to work with by building https://github.com/maaarghk/soundwire-dkms/commits/master 

which I got working in nixpkgs here:  https://github.com/IvanMalison/nixpkgs/tree/soundwire

I guess the question I have is:

What should we put in this configuration? Is it a requirement that it works with the current nixos stable release?

For unstable users, its actually probably better to provide a version where we do not add the extra kernel module and the extra package, while of stable users, it will be a requirement to add it. Curious about how we should proceed here.

IvanMalison:
Interestingly, the default linux kernel is still 5.10 even on master and in unstable See (https://github.com/NixOS/nixpkgs/blob/28135ad08a5d3b40df1d68a686e8f65a106def3e/pkgs/top-level/all-packages.nix#L21487)

That means that no matter what, we probably need to add something like:
```
  boot.kernelPackages = pkgs.linuxPackages_latest;
```

Would it be better to use something a bit less recent like
```
  boot.kernelPackages = pkgs.linuxPackages_5_12;
```

Since we are going to be adding a kernel patch (really just modifying kernel flags), we actually need a kernel recompilation anyway:
```
  boot.kernelPatches = [{
    name = "enable-soundwire-drivers";
    patch = null;
    extraConfig = ''
      SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES y
      SND_SOC_INTEL_SOUNDWIRE_SOF_MACH m
      SND_SOC_RT1308 m
    '';
    ignoreConfigErrors = true;
  }];
```

IvanMalison:
It looks like there is some precedent for bumping the kernel version here, e.g.:
https://github.com/NixOS/nixos-hardware/blob/360b5a7bedf955275fa3fc89bcbe28a2a85bf62b/lenovo/thinkpad/e495/default.nix#L10

Mic92:
Yes. Just set a newer kernel if current lts version is too old.

dell/xps/17-9700/README.wiki
```diff
@@ -0,0 +1,62 @@
...
+
+Not much tweaking of NixOS itself was needed. But we currently cannot automate
+the firmware setup, so this must be done by hand.
+
+=== Before installation ===
+
+These settings are needed both for booting the final install, and installer
+itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as
```
Mic92, line 57:
```suggestion
* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as
```

dell/xps/17-9700/common.nix
```diff
@@ -0,0 +1,29 @@
+{ lib, pkgs, ... }:
+
+{
+  # Boot loader
+  boot.loader.systemd-boot.enable = lib.mkDefault true;
+  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
```
Mic92, line 6: Some people might prefer grub over systemd-boot.
Also enabling canTouchEfiVariables for every upgrade can destroy the efi flash.
Hence we should only enable it for installation.

```suggestion
```

dell/xps/17-9700/common.nix
```diff
@@ -0,0 +1,29 @@
...
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
+
+  boot.kernelPatches = [{
+    name = "enable-soundwire-drivers";
+    patch = null;
+    extraConfig = ''
+      SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES y
+      SND_SOC_INTEL_SOUNDWIRE_SOF_MACH m
+      SND_SOC_RT1308 m
```
Mic92, line 21: We can also enable those by default in nixpkgs....

dell/xps/17-9700/nvidia/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  # This runs only nvidia, great for games or heavy use of render applications
+
+  ##### disable intel, run nvidia only and as default
+  services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
+  hardware.nvidia.modesetting.enable = lib.mkDefault true;
+  hardware.nvidia.prime = {
+    enable = lib.mkDefault true;
+    nvidiaBusId = lib.mkDefault "PCI:1:0:0";
+    intelBusId = lib.mkDefault "PCI:0:2:0";
+  }
```
Mic92, line 20:
```suggestion
  };
```

dell/xps/17-9700/common.nix
IvanMalison: I would assume that they are not already there for a reason, but not really sure.

dell/xps/17-9700/common.nix
IvanMalison: Okay, there is already a precedent for setting these though, e.g. in 

https://github.com/nixos/nixos-hardware/blob/b31fa05b585f1d40c2f9feecb3d269f4034979b1/dell/xps/15-9560/xps-common.nix#L6

Furthermore, these are lib.mkDefault... if people prefer something else they could easily override, right?

dell/xps/17-9700/common.nix
Mic92: I don't think so. Many kernel options are simply not enabled by default. People enable them as they need them.
Since it's just an optional module, this should be fine.

IvanMalison: > We can also enable those by default in nixpkgs....

I'm a little lost as to where these are set in nixpkgs. Seems like the builder lives here: https://github.com/ivanmalison/nixpkgs/blob/c159b9d51e4bd7a3d29744ab1c6baa78c71bf93b/pkgs/os-specific/linux/kernel/generic.nix#L79

but its not clear that anything gets passed by default.

dell/xps/17-9700/common.nix
Mic92: Right but this option is still opinionated and we should rather rely on nixpkgs default. There is nothing in the hardware that mandates a specific bootloader. I am going to remove the bootloader from profile as well.

dell/xps/17-9700/common.nix
Mic92: They can be enabled here: https://github.com/IvanMalison/nixpkgs/blob/c159b9d51e4bd7a3d29744ab1c6baa78c71bf93b/pkgs/os-specific/linux/kernel/common-config.nix

Mic92: This change in nixpkgs is not mandatory to merge this PR so.

dell/xps/17-9700/common.nix
Mic92: See https://github.com/NixOS/nixos-hardware/pull/307

workflow (approved):
This is a big improvement over having no working defaults, anything still blocking the merge approval @Mic92 ? :)

Mic92:
Yes. See https://github.com/NixOS/nixos-hardware/pull/291/files#r693242215

peterstorm:
I tried this, and it works great with regards to audio! But is there a way to get the brightness control of the panel to work?

peterstorm:
> > I tried this, and it works great with regards to audio! But is there a way to get the brightness control of the panel to work?
> 
> 
> 
> Yes, but this really depends a lot more on your desktop environment. What are you using?

I see, thank you for replying! I'm using xmonad, but with the plasma KDE environment.

IvanMalison:
> I see, thank you for replying! I'm using xmonad, but with the plasma KDE environment.

I use xmonad as well. You can take a peek at what I do here:

https://github.com/IvanMalison/dotfiles/blob/af36fddc4164a8ea6f0d020a1a96dc6b80a3a66d/dotfiles/config/xmonad/xmonad.hs#L1057

Essentially that uses this script: https://github.com/IvanMalison/dotfiles/blob/af36fddc4164a8ea6f0d020a1a96dc6b80a3a66d/dotfiles/lib/bin/brightness.sh#L3

Which ultimately calls this:

https://github.com/IvanMalison/dotfiles/blob/af36fddc4164a8ea6f0d020a1a96dc6b80a3a66d/dotfiles/lib/bin/brightness_manager.py#L7

which is just a python script that basically writes to the file that manages brightness

IvanMalison:
@peterstorm I'm honestly kind of surprised that using the KDE environment doesn't make these keys work though. When I use stock KDE brightness keys are bound.

PErhaps when you use XMonad as WM the keypressses are intercepted though.

peterstorm:
> @peterstorm I'm honestly kind of surprised that using the KDE environment doesn't make these keys work though. When I use stock KDE brightness keys are bound.
> 
> 
> 
> PErhaps when you use XMonad as WM the keypressses are intercepted though.

You seem to be correct. Just booted up without xmonad (because I can't get it to work after switching to flakes and Home Manager, hehe) and brightness controls work! I see you also are a guru in xmonad flakes, any chance I can reach out to you somewhere for help there? 😅

IvanMalison:
> You seem to be correct. Just booted up without xmonad (because I can't get it to work after switching to flakes and Home Manager, hehe) and brightness controls work! I see you also are a guru in xmonad flakes, any chance I can reach out to you somewhere for help there? 😅

Sure https://matrix.to/#/#taffybar:matrix.org might be a good place.

Mic92:
Moving this to draft mode so it's obvious this is not ready for review.

dell/xps/17-9700/nvidia/default.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../../common/cpu/intel
```
mkg20001, line 5: common/gpu/nvidia could also be added here, this has a part of the options below already included

NelsonJeppesen:
I cant speak to SND_SOC_RT1308 kernel patch but SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES and SND_SOC_INTEL_SOUNDWIRE_SOF_MACH have been added to the default nixos kernel and no longer needed

Mic92:
@IvanMalison  are you still working on this? I might close the pull request otherwise to encourage someone else to pick it up.

IvanMalison:
Isn't the change that is needed just 1 line or so?

Mic92:
> Isn't the change that is needed just 1 line or so?

Yeah. But no one bothered to do this changed and now it has a merge conflict. Please reopen when you are ready with the changes.

workflow:
> I cant speak to SND_SOC_RT1308 kernel patch but SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES and SND_SOC_INTEL_SOUNDWIRE_SOF_MACH have been added to the default nixos kernel and no longer needed. Works for the XPS 9710 which has similar hardware AFAIK

Is that `nixos-unstable`? 5.16 kernel on latest `nixos-21.11` still seems to require it.

Mic92:
Potentially only on unstable just now: https://github.com/NixOS/nixpkgs/blob/dd697d414853a99d0d5f05ee23430923c6da1176/pkgs/os-specific/linux/kernel/common-config.nix#L308

workflow:
@Mic92 @IvanMalison I re-opened it [here](https://github.com/NixOS/nixos-hardware/pull/401), with the opinionated bootloader parts removed.
Also changed Nvidia to offload by default.

PR #301, opened 2021-08-14
rpi4: add module for audio settings

Valodim:
This adds an option to enable and configure audio on the Raspberry Pi 4.

### Usage

```nix
{
  imports = [
  ];

  hardware.raspberry-pi."4".audio.enable = true;
}
```

Enabling this setting does two things:
1. Apply a dts overlay that is equivalent to `dtparam=audio=on`. This was [authored](https://gist.github.com/samueldr/e83102caf431ccef8bc2e825f9273ca7) by @samueldr, who told me to test and make a PR and here we are :)
2. configure `module-udev-detect` with `tsched=0` in pulseaudio's `default.pa`. This is a commonly recommended fix ([1](https://steamcommunity.com/app/353380/discussions/6/1642042464753800526) [2](https://steamcommunity.com/app/353380/discussions/6/1642042464753800526) [3](https://wiki.archlinux.org/title/PulseAudio/Troubleshooting#Glitches,_skips_or_crackling)) for audio glitches that appear after a while of usage. I haven't had audio issues since adding this snippet.

I don't know what the policies are around fixes like the second one on this repo? I thought it'd be easiest to make this PR as a basis for discussion. :)

### DIFF
--- a/raspberry-pi/4/audio.nix
+++ b/raspberry-pi/4/audio.nix
@@ -1,0 +1,40 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.hardware.raspberry-pi."4".audio;
+in
+{
+  options.hardware = {
+    raspberry-pi."4".audio = {
+      enable = lib.mkEnableOption ''
+        configuration for audio
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    hardware.deviceTree = {
+      overlays = [
+        # Equivalent to dtparam=audio=on
+        {
+          name = "audio-on-overlay";
+          dtsText = ''
+            /dts-v1/;
+            /plugin/;
+            / {
+              compatible = "brcm,bcm2711";
+              fragment@0 {
+                target = <&audio>;
+
+                __overlay__ {
+                  status = "okay";
+                };
+              };
+            };
+          '';
+        }
+      ];
+    };
+  };
+}
+
--- a/raspberry-pi/4/default.nix
+++ b/raspberry-pi/4/default.nix
@@ -2,6 +2,7 @@
 
 {
   imports = [
+    ./audio.nix
     ./dwc2.nix
     ./modesetting.nix
     ./poe-hat.nix


raspberry-pi/4/audio.nix
```diff
@@ -0,0 +1,46 @@
...
+      enable = lib.mkEnableOption ''
+        Enable configuration for audio
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    hardware.deviceTree = {
+      overlays = [
+        # Equivalent to dtparam=audio=on
```
Mic92, line 18: Has the device tree any advantage over kernel parameter?

samueldr: *It works*.

That is:

 - It works in the NixOS generations lifecycle; booting a generation with the setting turned on makes it work, a generation with the setting turned off makes it work.
 - U-Boot clobbers the firmware-provided FDT (FDT is the Flattened Device Tree in memory) with the dtb file from the generation, so *pretty much all setup done by the firmware is lost*.

raspberry-pi/4/audio.nix
```diff
@@ -0,0 +1,46 @@
...
+          name = "audio-on-overlay";
+          dtsText = ''
+            /dts-v1/;
+            /plugin/;
+            / {
+              compatible = "brcm,bcm2711";
+              fragment@0 {
+                target = <&audio>;
+
+              __overlay__ {
```
samueldr, line 29:
```suggestion
                __overlay__ {
```

Indentation is off by one.

raspberry-pi/4/audio.nix
samueldr: OH, and I missed the most obvious part here: `dtparam=` is **not** a kernel parameter!

These are `config.txt` option that should be seen as turning bios options on/off on the Raspberry Pi. The proprietary vendor firmware consumes `dtparam` options and modify the built-in FDT.

raspberry-pi/4/audio.nix
```diff
@@ -0,0 +1,46 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.hardware.raspberry-pi."4".audio;
+in
+{
+  options.hardware = {
+    raspberry-pi."4".audio = {
+      enable = lib.mkEnableOption ''
+        Enable configuration for audio
```
samueldr, line 10:
```suggestion
        configuration for audio
```

*Whether to enable **Enable** configuration for audio* → *Whether to enable configuration for audio*.

`mkEnableOption` adds its own prefix to the description.

 - https://github.com/NixOS/nixpkgs/blob/069cd2bdc5303b9aafece12544138bac35160f7c/lib/options.nix#L98

samueldr:
Completely out of scope for this, but I wanted to get this out of my thoughts:

It would be nice to have a *trivial dtparam builder*.

This is because `dtparam=audio=on` means "Add a fragment for the target `<&audio>`, set `status = "okay";` in `__overlay__`". It should be relatively easy to make a trivial text template and play ~~madlibs~~ *complete the overlay* using onlt the target name. Though this would work **only** for `=on` values.

raspberry-pi/4/audio.nix
Valodim: [I blame you](https://gist.github.com/samueldr/e83102caf431ccef8bc2e825f9273ca7) :P

Valodim:
I removed the `tsched=0` part from this PR, for now. I've been using that for a year or so because it fixed glitches, but I'm now running my pi without it to test if that still applies, and so far I haven't encountered any issues. Maybe the problem was fixed upstream in pulseaudio.

I'll keep an eye on it, if the issue returns I'll put it in  a separate PR.

samueldr (approved):

PR #302, opened 2021-08-15
raspberry-pi/4: Include tsched=0 fix in audio module

Valodim:
Follow-up to #301 :)

Watched a video yesterday evening, got audio glitches after some 15 minutes. Applied the setting, audio issues gone.

### DIFF
--- a/raspberry-pi/4/audio.nix
+++ b/raspberry-pi/4/audio.nix
@@ -35,6 +35,12 @@ in
         }
       ];
     };
+
+    # set tsched=0 in pulseaudio config to avoid audio glitches
+    # see https://wiki.archlinux.org/title/PulseAudio/Troubleshooting#Glitches,_skips_or_crackling
+    hardware.pulseaudio.configFile = lib.mkDefault pkgs.runCommand "default.pa" {} ''
+      sed 's/module-udev-detect$/module-udev-detect tsched=0/' ${config.hardware.pulseaudio.package}/etc/pulse/default.pa > $out
+    '';
   };
 }
 


samueldr:
Not sure if it's an issue or not, but this would make it harder for end-users to *also* use `hardware.pulseaudio.configFile` since this doesn't compose. (See also the suggestion.)

 - https://search.nixos.org/options?channel=21.05&show=hardware.pulseaudio.configFile&from=0&size=50&sort=relevance&query=pulseaudio.configFile

Probably not an issue though.

raspberry-pi/4/audio.nix
```diff
@@ -35,6 +35,12 @@ in
         }
       ];
     };
+
+    # set tsched=0 in pulseaudio config to avoid audio glitches
+    # see https://wiki.archlinux.org/title/PulseAudio/Troubleshooting#Glitches,_skips_or_crackling
+    hardware.pulseaudio.configFile = pkgs.runCommand "default.pa" {} ''
```
samueldr, line 41:
```suggestion
    hardware.pulseaudio.configFile = lib.mkDefault pkgs.runCommand "default.pa" {} ''
```

While it won't compose with further changes to `hardware.pulseaudio.configFile` made by the user, at least it won't conflict.

Valodim:
Yeah, agreed that's not pretty. I can't think of a way to make this composable though, and `pulseaudio.configFile` is pretty much a "you're on your own if you use this" option already since it overrides the shipped defaults - hence the `sed` approach taken here.

samueldr (approved):
I haven't personally tested it, but it looks okay.

bnjmnt4n:
Not sure if I'm doing something wrong, but I got this error after bumping to the latest `nixos-hardware` and enabling the `audio` option for RPi4:

```
error: attempt to call something which is not a function but a set

       at /nix/store/w3yq2dwbaq8cn1j0ll9bm9wsy9w105lc-source/raspberry-pi/4/audio.nix:41:38:

           40|     # see https://wiki.archlinux.org/title/PulseAudio/Troubleshooting#Glitches,_skips_or_crackling
           41|     hardware.pulseaudio.configFile = lib.mkDefault pkgs.runCommand "default.pa" {} ''
             |                                      ^
           42|       sed 's/module-udev-detect$/module-udev-detect tsched=0/' ${config.hardware.pulseaudio.package}/etc/pulse/default.pa > $out

       … while evaluating the attribute 'content'
```

Elsewhere, I only have `hardware.pulseaudio.enable = true`.

Valodim:
*sigh* that one's on me. never a patch too trivial to double-check. sorry about that. looks like it's just missing some parenthesis, will fix today after work

PR #303, opened 2021-08-18
lenovo/thinkpad/e495: use native acpi backlight

ebbertd:
Saving/loading the backlight state requires the `acpi_backlight=native` kernel parameter.


Mic92:
Thanks!

PR #304, opened 2021-08-18
Lenovo ThinkPad E14

ebbertd:
Init profiles for the Lenovo ThinkPad E14.

This is mostly based on the L14 and not completely ready yet, it already seems to work quite well, but I haven't tested all details yet. I put it here for feedback and keeping track of the progress.

Disclosure: I only have an E14 with AMD, not with Intel.


Mic92:
evaluation on master is fixed again.

reiferschris:
Thank you! This fixed the backlight issue for me. (E14 with AMD)

PR #307, opened 2021-08-20
remove systemd-boot from hardware profiles

Mic92:
- We should not enable canTouchEfiVariables by default as this
  wears out the EFI storage.
- We should not set systemd-boot as default. This is up to the user to
  decide. There are exceptions when hardware only supports specific
  bootloaders so.

### DIFF
--- a/dell/xps/15-9560/xps-common.nix
+++ b/dell/xps/15-9560/xps-common.nix
@@ -1,10 +1,7 @@
 { lib, ... }:
 
 {
-
   # Boot loader
-  boot.loader.systemd-boot.enable = lib.mkDefault true;
-  boot.loader.efi.canTouchEfiVariables = lib.mkDefault true;
   boot.kernelParams = lib.mkDefault [ "acpi_rev_override" ];
 
   # This will save you money and possibly your life!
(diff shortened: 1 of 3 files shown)


dell/xps/15-9560/xps-common.nix
```diff
@@ -1,10 +1,7 @@
 { lib, ... }:
 
 {
-
   # Boot loader
```
Mic92, line 4:
```suggestion
```

samueldr (approved):

dwf:
Hey @Mic92, can you elaborate on what you meant by wearing out EFI storage? Does this go beyond ordinary SSD wear-and-tear in some fashion? Should it be documented on the option in nixpkgs?

(To be clear I am not disputing that this PR was a good idea, I agree that it was, just trying to learn something.)

Mic92:
> Hey @Mic92, can you elaborate on what you meant by wearing out EFI storage? Does this go beyond ordinary SSD wear-and-tear in some fashion? Should it be documented on the option in nixpkgs?
> 
> (To be clear I am not disputing that this PR was a good idea, I agree that it was, just trying to learn something.)

EFI storage is usually backed by some cheap flash memory chips on the motherboard (this is not your ssd or hdd). If one write them too often they could become read-only, which is bad because you cannot easy replace them without solid soldering skills.

Issue #317, opened 2021-09-27
Adding support for building installers?


samueldr:
More of a question for other maintainers, to see if there are strong reasons against.

I'm currently adding an ARM device to `nixos-hardware`, and have been thinking it would be nice to be able to build the NixOS installer images (whether SD image, or UEFI iso) with the profile baked-in.

With ARM devices, it would be sometimes necessary, as sometimes the kernel requires tweaks and such nonsense.

With your run-of-the-mill x86_64 system it might also be helpful. I'm thinking about those laptops with *annoying to setup Wi-Fi*, or maybe other "fun" quirks requiring "fun" configuration.

* * *

There's another dimension to this question: would it be acceptable to also provide some sort of "harness" to make cross-compilation easier? I'm thinking about the "first time setup" for a local AArch64 system, when the user has no AArch64 hardware to use to build that installer system. Basically slipstream [part of what I already have here](https://github.com/samueldr/cross-system/) so it's easier for end-users.

samueldr:
@Mic92 any opinion for or against?

Mic92:
I think providing installer would still fit into the scope of this project.
I am just not sure how we would provide the CI to make this happen.
Images itself can be hosted in  a github release as I demonstrate here: https://github.com/nix-community/nixos-images/releases/tag/nixos-unstable

PR #318, opened 2021-09-27
kobol/helios64: init

samueldr:
This adds support for the Kobol Helios64 to nixos-hardware.

This pull request does not configure anything related to either the *platform firmware* or the *bootloader*, by design.

The kernel used is mainline 5.10 (LTS), with a patchset [borrowed from Armbian](https://github.com/armbian/build/tree/744ea89a589d62cb6f409baab60fc6664520bc39/patch/kernel/archive/rockchip64-5.10). Only generic patches or Helios64 patches were kept from that directory. Their *overlay* patches were dropped.

> :warning: **The platform firmware requires usage of the proprietary RAM training!!** Using the default open source RAM training from U-Boot *will cause instability*. This instability is observed in memory being unreliable in different scenarios, but generally under load. *This instability can lead to data loss.* Under the same scenarios, and harsher scenarios, using the proprietary RAM training was rock solid, while the open source RAM training was pretty reproducibly problematic.

* * *

[I would like to make it easy for end-users to build an installer image](https://github.com/NixOS/nixos-hardware/issues/317) (either SD image or UEFI iso) before merging this.

* * *

I recommend my own project (Tow-Boot) for the *platform firmware*. Hopefully this is okay to do.

* * *

I have been playing with the SD image build, booted from USB, and it seems to work just fine for NAS use cases.

* * *

cc @angerman

### DIFF
--- a/kobol/helios64/README.md
+++ b/kobol/helios64/README.md
@@ -1,0 +1,57 @@
+Kobol Helios64
+==============
+
+The `default.nix` module provides the required setup for the system to work.
+
+The `recommended.nix` module adds recommended settings for the system.
+
+Status
+------
+
+### Works
+
+ - SATA hard drives
+ - Ethernet (1gbps)
+ - Serial through USB type-c (`ttyS2`)
+
+### Untested
+
+ - Ethernet (2.5gbps)
+ - DP video out
+ - UPS behaviour
+ - `rootfs` on SATA drives
+
+### Disabled
+
+Due to misbehaviour, `ttyS0` (`&uart0`, `serial@ff180000`) has been disabled
+via a kernel patch.
+
+Without this change, using, or attempting to use `ttyS0` will break serial
+output from `ttyS2`.
+
+
+Kernel
+------
+
+Only Linux 5.10 (LTS) is supported, using the patch set derived from Armbian.
+
+
+Requirements
+------------
+
+A *platform firmware* needs to be provided out of band for the system.
+
+The author recommends Tow-Boot, for which a [draft pull request](https://github.com/Tow-Boot/Tow-Boot/pull/54)
+adds support for the Helios64.
+
+Any other supported *platform firmware* should work too.
+
+
+Notes
+-----
+
+### Baud rate
+
+The serial baud rate is configured for `115200`, which is a more common default
+than the usual for Rockchip at `1500000`. See [the rationale for the decision](https://github.com/Tow-Boot/Tow-Boot/pull/33).
+
--- a/kobol/helios64/README.md
+++ b/kobol/helios64/README.md
@@ -46,6 +46,11 @@ adds support for the Helios64.
 
 Any other supported *platform firmware* should work too.
 
+> **NOTE**: at the time of writing (2021-10-10) the *platform firmware*
+> **must** make use of the proprietary ram training. The open source equivalent
+> will make the system unstable, and worse, will cause silent memory
+> corruption, in addition to loud memory corruption.
+
 
 Notes
 -----
--- a/kobol/helios64/modules/heartbeat.nix
+++ b/kobol/helios64/modules/heartbeat.nix
@@ -1,0 +1,20 @@
+{ pkgs, lib, ... }:
+{
+  systemd.services.heartbeat = {
+    enable = true;
+    description = "Enable heartbeat & network activity led on Helios64";
+    serviceConfig = {
+      Type = "oneshot";
+      ExecStart = ''
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo heartbeat | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:\\:status/trigger'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo netdev | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/trigger'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo eth0 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/device_name'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/link'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/rx'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/tx'
+      '';
+    };
+    after = [ "getty.target" ];
+    wantedBy = [ "multi-user.target" ];
+  };
+}
--- a/kobol/helios64/overlay.nix
+++ b/kobol/helios64/overlay.nix
@@ -1,0 +1,74 @@
+self: super: {
+  linux_5_10_helios64 = self.linux_5_10.override {
+
+    kernelPatches = [
+      {
+        name = "helios64-patch-set.patch";
+        patch = self.fetchurl {
+          # v5.10.68..v5.10-helios64-2021-10-10 @ https://github.com/samueldr/linux
+          # Hosted as a pre-rendered patch because `fetchpatch` strips added files.
+          url = "https://gist.githubusercontent.com/samueldr/1a409f88f2107054c87a70403686b871/raw/abee3d5d5415c466f8111371b63f759f614547c6/helios64.patch";
+          sha256 = "1gx2z345vb4r2mdfmydbzc5baj58rrn416rzb2fz7azxpyib5ym4";
+        };
+      }
+    ];
+
+    # Configuration mainly to remove unused platforms and things.
+    structuredExtraConfig = with self.lib.kernel; {
+      ARCH_ROCKCHIP = yes;
+
+      ARCH_ACTIONS = no;
+      ARCH_AGILEX = no;
+      ARCH_SUNXI = no;
+      ARCH_ALPINE = no;
+      ARCH_BCM2835 = no;
+      ARCH_BERLIN = no;
+      ARCH_BRCMSTB = no;
+      ARCH_EXYNOS = no;
+      ARCH_K3 = no;
+      ARCH_LAYERSCAPE = no;
+      ARCH_LG1K = no;
+      ARCH_HISI = no;
+      ARCH_MEDIATEK = no;
+      ARCH_MESON = no;
+      ARCH_MVEBU = no;
+      ARCH_MXC = no;
+      ARCH_QCOM = no;
+      ARCH_RENESAS = no;
+      ARCH_S32 = no;
+      ARCH_SEATTLE = no;
+      ARCH_STRATIX10 = no;
+      ARCH_SYNQUACER = no;
+      ARCH_TEGRA = no;
+      ARCH_SPRD = no;
+      ARCH_THUNDER = no;
+      ARCH_THUNDER2 = no;
+      ARCH_UNIPHIER = no;
+      ARCH_VEXPRESS = no;
+      ARCH_VISCONTI = no;
+      ARCH_XGENE = no;
+      ARCH_ZX = no;
+      ARCH_ZYNQMP = no;
+      ARCH_RANDOM = no;
+      ARCH_R8A77995 = no;
+      ARCH_R8A77990 = no;
+      ARCH_R8A77950 = no;
+      ARCH_R8A77951 = no;
+      ARCH_R8A77965 = no;
+      ARCH_R8A77960 = no;
+      ARCH_R8A77961 = no;
+      ARCH_R8A77980 = no;
+      ARCH_R8A77970 = no;
+      ARCH_R8A774C0 = no;
+      ARCH_R8A774E1 = no;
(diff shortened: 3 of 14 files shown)


kobol/helios64/overlay.nix
```diff
@@ -0,0 +1,73 @@
+self: super: {
+  linux_5_10_helios64 = self.linux_5_10.override {
+
+    kernelPatches = [
+      {
+        name = "helios64-patch-set.patch";
+        patch = ./kernel/helios64-5.10.patch;
```
Mic92, line 7: To keep this repo small, can you put the large patch in an external gist?
We can also upload it to https://github.com/NixOS/nixos-hardware/releases/tag/not-a-release

samueldr: I wasn't sure about the policy for patches, but yes it can be external.

Two propositions:

 - Fetching the patches directly from downstream Armbian; would be messy with ~20 or so fetchpatches. It's the more "semantically honest" to where the patches are directly coming from.
 - Fetching from my linux kernel fork as a `rev1...rev2` reference, less messy, and also better represents how I'm actually selecting and authoring the patches.

Mic92: I think the second option is easier to maintain.

kobol/helios64/modules/heartbeat.nix
```diff
@@ -0,0 +1,20 @@
...
+    description = "Enable heartbeat & network activity led on Helios64";
+    serviceConfig = {
+      Type = "oneshot";
+      ExecStart = ''
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo heartbeat | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:\\:status/trigger'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo netdev | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/trigger'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo eth0 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/device_name'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/link'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/rx'
+        ${lib.getBin pkgs.bash}/bin/bash -c 'echo 1 | ${lib.getBin pkgs.coreutils}/bin/tee /sys/class/leds/helios64\\:blue\\:net/tx'
```
Mic92, line 14: No strong opinion but system-tmpfiles allows to values like this to sysfs and looks more readable:

```nix
{
systemd.tmpfiles.rules = [
   "f     /file/to/create                          mode user group -           content"
];
}
```

I also don't know why `tee` is used here, echo should be enough. Also the `script` attribute could be used.

Mic92:
Moving this to draft mode so it's obvious this is not ready for review.

kobol/helios64/modules/heartbeat.nix
angerman: This packaging was a mostly verbatim translation from their armbian packages. I didn't try to make it fit nixos much, just getting it working was the primary motivation.

angerman:
@samueldr can you elaborate on the ramtraning note, and where you found that? Maybe that explains the occational instability issues I've run into?

samueldr:
@angerman through experimentation really.

I ran `memtester` a lot, nix evals, and ZFS, concurrently, as doing all these at once was sure to bring the system down at some point.

Additionally there is a thread that *points the problem out* on the armbian forums, but they themselves have not solved the problem. I extrapolated from the reports and what was observed.

At first I tried following the suggestions of the armbian forum, which were to disable frequency switching, and/or run at slower cpu frequencies. It helped, but never solved the problems *under heavy load*. Meanwhile, replacing the OSS ram training with the proprietary ram training solved the issues under all cases.

I'm soak testing, and this is my uptime since the last *non-crash* reboot:

```
[samuel@ednakrabapple:~]$ uptime
 23:22:42  up 19 days 10:42,  2 users,  load average: 0.06, 0.02, 0.00
```

A list of few instabilities I observed:

 - memtester seeing bad things (can't remember what exactly, but it found issues)
 - random segfaults
 - **Nix evaluation getting corrupted**
 - kernel subsystem oopsing (but no panics)
     - Especially ZFS with a scrub
 - kernel panics
 - **detectable** ZFS data corruption (thankfully on bogus test data)

angerman:
@samueldr huh, I've mostly just seen random kernel crashes (but yes, under load), this got a bit better with more recent kernels. Guess it's time to actually upgrade my helios with the proprietary ram trainer.

Does this changeset contain all the logic, or do you have some extra ontop?

samueldr:
The kernel patch set, contains *mainly* generic rk3399 changes from armbian and kobol64-specific changes. It's basically the same patch set you were using, but anything *obviously* not needed filtered out. It probably can be culled a bit more.

Everything in this PR is strictly about the operating system.

This means that the proprietary ram training is not part of this! The ram training is part of the platform firmware (U-Boot). I thought I had linked to the WIP PR for Tow-Boot that adds kobol64 support already.

 - https://github.com/Tow-Boot/Tow-Boot/pull/54

Note the current state in the commits, I started adding support for the whole proprietary rockchip miniloader at first, but I stopped early and thought "what if I use *just* the ram training?", and it worked.

I do not intend to add support for "bundling" a platform firmware in a disk image. This is not how things should work. [I am planning for a better NixOS on ARM already](https://discourse.nixos.org/t/planning-for-a-better-nixos-on-arm/15346), and part of this is the operating system is not managing the platform firmware. I would normally recommend installing the firmware to SPI Flash, but as I was still unsure about the results, I only tested on SD card. I still need to test the SPI build with proprietary ram training.

samueldr:
I guess my main issue right now, since @angerman tested recently, is figuring out where to go with all this.

I guess the path forward would be:

 - Update the pinned kernel (and rebase patches if necessary)
 - Further review patches to keep only what's needed
 - On Tow-Boot's side, get the SPI platform firmware going (some required work to help make this easier has been done recently).

Additionally, I think there's only one additional NixOS community user who got an Helios64. Considering they closed down, I suppose it's unlikely we're going to get more users for it. It's also unlikely we'll see any news from Kobol about the marginal behaviour observed with the RAM.

Mic92:
I know that @flokli also has the hardware :)

flokli:
> Additionally, I think there's only one additional NixOS community user who got an Helios64

Yup, one of those might be me. At the time of testing, I've been running 5.14.13 with the following patches:

```
  kernelPatches = [
    {
      name = "0001-arm64-dts-rockchip-Enable-HDD-power-on-helios64.patch";
      patch = ./0001-arm64-dts-rockchip-Enable-HDD-power-on-helios64.patch;
    }
    {
      name = "0002-arm64-dts-rockchip-fix-poweroff.patch";
      patch = ./0002-arm64-dts-rockchip-fix-poweroff.patch;
    }
    {
      name = "0003-arm64-dts-rockchip-helios64-rk808-add-missing-regula.patch";
      patch = ./0003-arm64-dts-rockchip-helios64-rk808-add-missing-regula.patch;
    }
    {
      name = "0001-arm64-dts-rockchip-helios64-define-usb-hub-and-2.5Gb.patch";
      patch = ./0001-arm64-dts-rockchip-helios64-define-usb-hub-and-2.5Gb.patch;
    }
    {
      name = "0001-Remove-overclock-from-helios64.patch";
      patch = ./0001-Remove-overclock-from-helios64.patch;
    }
  ];
```

Some of them might have already made its way upstream ;-)

I didn't really use the NixOS version since then, mostly because I didn't bother also adding fancontrol config (which is required to not have the system /also/ get unstable due to overheating.

I can confirm the armbian version also is unstable - the ram training stuff sounds really promising, thanks for the digging :-)

flokli:
@samueldr did you get `reboot` (not poweroff) to work with your set of patches? That was still missing from the kernel I booted back then…

samueldr:
@flokli depends what you mean exactly.

The `reboot` call works. But in my limited undiagnosed experience, the platform firmware stays stuck, I don't recall at which step exactly, but I think it's before U-Boot "proper" starts.

Is this what you're observing?

flokli:
When issuing a reboot, I've seen the system prepare for a reboot, but actually nothing happens after that - no serial output from uboot or an earlier stage at all.

I was suspecting the kernel not properly rebooting, since [a kernel patch was also required to fix poweroff](https://lore.kernel.org/all/20211220143041.521706721@linuxfoundation.org/).

Curiously, `reboot` works on Armbian.

samueldr:
Are you booting using UEFI? There are generic RK3399 issues with U-Boot's services that maye be making things harder to diagnose too.

flokli:
It seems armbian is not:

```
[    0.000000] efi: UEFI not found.
[    0.010091] EFI services will not be available.
```

GoogleBot42:
> Additionally, I think there's only one additional NixOS community user who got an Helios64.
I have one as well.

Mic92:
Please re-open this or anyone else, when this it is closer to completion. I try to keep the amount of PRs short in this repo to give fast review feedback cycles.

womfoo:
I've been running this on [my helios64](https://github.com/womfoo/dotfiles/blob/master/nixos/habilog/configuration.nix) for over a week now and it's been stable as its ever been. It used to crash at least ~2-3x a week during heavy load on kernel 5.9 without the ram training.

Given that it takes a lot of effort preparing patches from armbian and/or elsewhere, we could also consider merging this as-is as kernel 5.10 will be supported until 2026.

Mic92:
> Given that it takes a lot of effort preparing patches from armbian and/or elsewhere, we could also consider merging this as-is as kernel 5.10 will be supported until 2026.

Ok. Can you open a pull request and also set 5.10 as a kernel?

womfoo:
Sure, will test on stable and raise a PR

Mic92:
@womfoo what uboot firmware did you end up using?

GoogleBot42:
Not womfoo but I'm using tow-boot.  https://github.com/Tow-Boot/Tow-Boot/pull/54  You will need to build with the proprietary memory timings though for it to be stable.  It's been working well for me so far.

kobol/helios64/README.md
```diff
@@ -0,0 +1,62 @@
...
+> will make the system unstable, and worse, will cause silent memory
+> corruption, in addition to loud memory corruption.
+
+
+Notes
+-----
+
+### Baud rate
+
+The serial baud rate is configured for `115200`, which is a more common default
```
Mic92, line 60: Is not possible to have multiple baud rates? At least with getty this was possible.
Than one could default to one and potentially upgrade to a higher that is less laggy.

Mic92:
I just now don't have access to the hardware, but the basic workflow would be to first boot and flash towboot to the Emmc and than build sd-card image installer using https://github.com/nix-community/nixos-generators and this profile?
Also what would be the serial speed of the towboot installer, 115200 right?
Does towboot also picks up a m2 SSD for booting?

GoogleBot42:
Not sure if this is already known but it looks like some people have been going to a lot of work getting helios64 support upstreamed in the linux kernel.  If 5.16 is used it seems to have more features than the existing patches being used.  For example, the 2.5G ethernet works.  I think there are a few patches that only exist for 5.17 I'm not sure how important those are.  I'd like to use vanilla 5.16 so building the kernel isn't needed.

Although, if we use the upstream kernel, since some names are different the heatbeat and fan controls would need some small changes to work again.

samueldr:
I will not re-open this PR, and will never support this board.

After [stress-testing the device with the lowest ram speed training available](https://github.com/Tow-Boot/Tow-Boot/pull/54#issuecomment-1257067670), I have reproduced the fickleness of the device *yet again*.

I suspect that anyone having any form of success with this hardware either hasn't pushed the thing far enough to fail noisily, or worse, is living with silent corruption.

I am open to reconsidering this in the future, if someone has good examples of working and stable configurations. It would need to be proven stable under *stress* conditions. Notes on how I stressed the machine:

Ran each in their own loop:

 - nix-instantiate '<nixpkgs/nixos>' -A config.system.build.toplevel
 - memtester 2G
 - `grep -R a.c.e.g.i.k.m.o.q.s.u.w.y .` on a big dir
 - `tree` on a big dir

The "big dir" is stored on a single ZFS filesystem.

I also was using it heavily through NFS during testing to hasten an expected crash. Note that *NFS has already been verified to not be the cause*, I ran the system with other configurations and ended-up with instability there too.

Mic92:
I run the device without any issues and zfs: https://github.com/Mic92/dotfiles/blob/master/nixos/blob64/configuration.nix
I recently had different machines with nfs (arm and x86), that crashes on the 5.15 but not on 5.10 - so this is unrelated to the hardware I would say.

angerman:
I've been running with
```
Linux helios64 5.10.108 #1-NixOS SMP PREEMPT Wed Mar 23 08:13:30 UTC 2022 aarch64 GNU/Linux
```

I remember running stress tests on that machine a while back over days without being able to make it crash. But I'm still weary of silent corruption.

samueldr:
My current theory is that either a large batch, or all devices are marginal to a degree. I seem to remember you getting yours over a month before mine.

The problem is that without the vendor to help identify the problem and make statement about knowing what the problem is, what we have is *multiple proofs* (see the armbian forums discussions) that people are having issues, and no concrete proofs that some are actually stable.

So without a way to figure out the underlying problem, identify where it exists, cautiousness makes me think they all should be assumed bad. Especially since I ran mine fine for months, until it wasn't under load, multiple times.

PR #319, opened 2021-09-28
AMD: include 32-bit driver for Vulkan

j0hax:
In order for many Proton Games to work OOTB with modern AMD graphics cards `driversi686Linux.amdvlk` is needed.

I didn't include the package previously as it hadn't been backported into stable, now I think it's a good time to include it.

### DIFF
--- a/common/gpu/amd/default.nix
+++ b/common/gpu/amd/default.nix
@@ -10,6 +10,10 @@
     amdvlk
   ];
 
+  hardware.opengl.extraPackages32 = with pkgs; [
+    driversi686Linux.amdvlk
+  ];
+
   hardware.opengl = {
     driSupport = lib.mkDefault true;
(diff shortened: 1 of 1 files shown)


common/gpu/amd/default.nix
```diff
@@ -8,6 +8,7 @@
     rocm-opencl-icd
     rocm-opencl-runtime
     amdvlk
+    driversi686Linux.amdvlk
```
SuperSandro2000, line 11: Wrong option https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/configuration/gpu-accel.chapter.md#amd-sec-gpu-accel-vulkan-amd

j0hax: Good catch.

PR #320, opened 2021-10-03
Apply SSD optimizations for Dell XPS 13 9310

sirkonst:


SuperSandro2000:
According to https://www.dell.com/de-de/shop/laptops/xps-13-laptop/spd/xps-13-9310-laptop the default configuration is only with SSD.

SuperSandro2000 (approved):

PR #333, opened 2021-11-15
fix(intel): add opengl packages for 32bit

bobvanderlinden:
At the moment the OpenGL extraPackages for Intel are only added for the 64bit variant, but not the 32bit one.
The AMD does add its modules to the 32bit variant: https://github.com/NixOS/nixos-hardware/blob/a57fc74bc3c88537b7f80bc7bee428dd9e9a34ca/common/gpu/amd/default.nix#L13-L15

Note that `pkgsi686Linux` was used instead of `driversi686Linux`, because `intel-media-driver` was not defined for `driversi686Linux` in nixpkgs: https://github.com/nixos/nixpkgs/blob/87304645b876629a056d4e4b3043b99d125a054d/pkgs/top-level/all-packages.nix#L15882-L15892

### DIFF
--- a/common/cpu/intel/default.nix
+++ b/common/cpu/intel/default.nix
@@ -12,4 +12,11 @@
     libvdpau-va-gl
     intel-media-driver
   ];
+
+  hardware.opengl.extraPackages32 = with pkgs.pkgsi686Linux; [
+    vaapiIntel
+    vaapiVdpau
+    libvdpau-va-gl
+    intel-media-driver
+  ];
 }


common/cpu/intel/default.nix
```diff
@@ -12,4 +12,11 @@
     libvdpau-va-gl
     intel-media-driver
   ];
+
+  hardware.opengl.extraPackages32 = with pkgs.pkgsi686Linux; [
```
Mic92, line 16: Since 32bit is phasing out, should this be only added if `hardware.opengl. driSupport32Bit` is set?

bobvanderlinden: That would be strange I think. `extraPackages32` is only referred to if `driSupport32Bit` is set. Functionally there will be no change if `extraPackages32` is set while `driSupport32Bit` is not.

See https://github.com/NixOS/nixpkgs/blob/0699530f08290f34c532beedd66046825d9756fa/nixos/modules/hardware/opengl.nix#L142-L143 and https://github.com/NixOS/nixpkgs/blob/0699530f08290f34c532beedd66046825d9756fa/nixos/modules/hardware/opengl.nix#L18-L21

Mic92: Ok. If that is the case, than we don't need to make it conditional.

bobvanderlinden:
Thanks for the quick review and reply! 👍

maralorn:
Hello! I am sorry to say this PR is not compatible with nixos-21.05.

The intel-media-driver has a dependency of intel-gmmlib. intel-gmmlib does support i686 but only on nixpkgs master, because that has only been recently fixed by @SuperSandro2000. (See https://github.com/NixOS/nixpkgs/pull/136738) Sadly on nixos-21.05 currently intel-gmmlib only supports x86_64 linux.

I include nixos-hardware for my thinkpad in my config. When I tried to bump the pin right now I ran into this issue.

I am not sure how best to proceede and how many people will be affected by this.

* Maybe we could backport i686 support for intel-gmmlib to stable?
* Or we remove the intel-media-driver for now?
* Or we tell users to pin an older version of nixos-hardware, when they are on stable?

DavHau:
This just broke my 21.05 based config for the Thinkpad carbon x1 gen7 as well.
I suggest we roll this back asap, otherwise many people will waste their time debugging this. It is not obvious where the breakage comes from.
Afterwards, we can still discuss a solution on how to move this forward.

Mic92:
We can include this change conditionally based on the nixos version: 

```
lib.mkIf ((lib.versionOlder (lib.versions.majorMinor lib.version) "21.11")
```

bobvanderlinden:
Oof, sorry for the problems :( I didn't test this on earlier versions and I'm on unstable. I'll look into @Mic92 's suggestion.

PR #334, opened 2021-11-16
Use a more recent kernel for Thinkpad X1

mrkkrp:
This solves lagging noticeable on high-resolution screens.

### DIFF
--- a/lenovo/thinkpad/x1/9th-gen/default.nix
+++ b/lenovo/thinkpad/x1/9th-gen/default.nix
@@ -1,0 +1,11 @@
+{ lib, pkgs, ... }:
+{
+  imports = [
+    ../.
+  ];
+
+  # This solves lagging noticeable on high-resolution screens.
+  boot.kernelPackages = lib.mkIf
+    (lib.versionOlder pkgs.linux.version "5.15")
+    (lib.mkDefault pkgs.linuxPackages_latest);
+}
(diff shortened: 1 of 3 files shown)


lenovo/thinkpad/x1/default.nix
```diff
@@ -1,6 +1,9 @@
+{ lib, pkgs, ... }:
+
 {
   imports = [
     ../.
     ../../../common/cpu/intel
   ];
+  boot.kernelPackages = pkgs.linuxPackages_latest;
```
Mic92, line 8: Let's make this future proof.

```suggestion
  # This solves lagging noticeable on high-resolution screens.
  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.15")  pkgs.linuxPackages_latest;
```

Mic92:
Which generation of x1 are we talking about?

mrkkrp:
The model I have is ThinkPad X1 Carbon Gen 9. Should we put this in a separate directory?

mrkkrp:
@Mic92 Is there something else that should be done for this PR to become mergeable?

lenovo/thinkpad/x1/default.nix
```diff
@@ -1,6 +1,10 @@
+{ lib, pkgs, ... }:
+
 {
   imports = [
     ../.
     ../../../common/cpu/intel
   ];
+  # This solves lagging noticeable on high-resolution screens.
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.15")  pkgs.linuxPackages_latest;
```
Mic92, line 9:
```suggestion
  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.15")  (lib.mkDefault pkgs.linuxPackages_latest);
```

Mic92:
> The model I have is ThinkPad X1 Carbon Gen 9. Should we put this in a separate directory?

That might be a good idea. X1 has been on the market for a while. We probably should not force long-term users of older devices to a new kernel since this might make it harder to use some out-of-tree kernel modules.

mrkkrp:
Okay, I have squashed the commits and instead of modifying the shared config I created a `9th-gen` subdirectory.

lenovo/thinkpad/x1/9th-gen/default.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, pkgs, ... }:
+{
+  imports = [
+    ../.
+  ];
```
Mic92, line 5: Can you also reference this profile in the README as well as the flake?

mrkkrp: Done.

PR #342, opened 2021-11-25
nanopc-t4: remove deprecated serial speed

Mic92:
This option no longer has any effects. We can set kernel parameter now, but since I don't this device it's not clear to me which tty I would need to configure.

cc @jakubgs

### DIFF
--- a/friendlyarm/nanopc-t4/default.nix
+++ b/friendlyarm/nanopc-t4/default.nix
@@ -8,7 +8,7 @@
   };
 
   # UART debug console bitrates.
-  services.mingetty.serialSpeed = [ 1500000 115200 ];
+  boot.kernelParams = [ "console=ttyS2,1500000" ];
 
   # Enable additional firmware (such as Wi-Fi drivers).
   hardware.enableRedistributableFirmware = lib.mkDefault true;


friendlyarm/nanopc-t4/default.nix
```diff
@@ -7,9 +7,6 @@
     generic-extlinux-compatible.enable = lib.mkDefault true;
   };
 
-  # UART debug console bitrates.
-  services.mingetty.serialSpeed = [ 1500000 115200 ];
```
Mic92, line 11: One could have something like:

```
  boot.kernelParams = ["console=ttyS2,1500000"];
```

But reading the wiki it's not clear which tty number the UART debug console will get.
It might also depend what GPIOs the user wants to use for console vs other functionality.

jakubgs:
Is this related to the new `21.11` release?

friendlyarm/nanopc-t4/default.nix
jakubgs: As far as I can tell the device is creating two devices:
```
 > sudo dmesg | grep console
[    0.003454] Console: colour dummy device 80x25
[    0.003522] printk: console [tty0] enabled
[    0.180443] fbcon: Taking over console
[    1.836710] printk: console [ttyS2] enabled
[    2.390934] dw-apb-uart ff1a0000.serial: forbid DMA for kernel consol
```
The `/dev/tty0` and `/dev/ttysS2`, but there's others:
```
 > ls -li /dev/tty0 
     18 crw--w----    1 root     tty         4,   0 Feb 21  2013 /dev/tty0

 > ls -li /dev/ttyS*
    125 crw-rw----    1 root     dialout     4,  65 Feb 21  2013 /dev/ttyS1
    128 crw--w----    1 root     tty         4,  66 Feb 21  2013 /dev/ttyS2
    127 crw-rw----    1 root     dialout     4,  67 Feb 21  2013 /dev/ttyS3
```
So I'd say `ttyS2` is a good bet.

jakubgs: I tested your recommended setting:
```
  boot.kernelParams = ["console=ttyS2,1500000"];
```
And I've confirmed it works with `picocom`:
```
picocom -b 1500000 /dev/ttyUSB0
```
```
 > sudo dmesg | grep model
[    0.000000] Machine model: FriendlyElec NanoPC-T4

~
 > 
*** baud: 1500000
*** flow: none
*** parity: none
*** databits: 8
*** stopbits: 1
*** dtr: up
*** rts: up
*** mctl: DTR:1 DSR:0 DCD:0 RTS:1 CTS:0 RI:0
```

jakubgs: Actually, I tested without either of the settings - `boot.kernelParams` or `services.mingetty.serialSpeed` - and it worked.

Mic92:
> Is this related to the new `21.11` release?

Yes. But it's also an issue on unstable right now.

PR #345, opened 2021-12-10
raspberry-pi/4: Add option to turn on the ARM I2C bus (i2c1).

miangraham:
This adds an option to turn on and enable access to i2c-1, the ARM I2C bus on Pi 4. Also supports i2c0, the VideoCore I2C bus.

### Motivation

This was the shortest path for me to get to the result below and came only after a full day of banging my head against other config options. I'm contributing to hopefully save someone else the time in the future.

### Usage

```nix
{
  imports = [
  ];

  hardware.raspberry-pi."4".i2c1.enable = true;
}
```

### Testing

After rebuilding and rebooting, `/dev/i2c-1` should exist.

```console
$ ls -l /dev/i2c-1
crw-rw---- 1 root i2c 89, 1 Dec 11 01:07 /dev/i2c-1
```

Any user with access should be able to query capabilities of the device, even without anything plugged in.

```console
$ nix-shell -p i2c-tools --run "i2cdetect -F 1"
Functionalities implemented by /dev/i2c-1:
I2C                              yes
SMBus Quick Command              yes
SMBus Send Byte                  yes
SMBus Receive Byte               yes
SMBus Write Byte                 yes
SMBus Read Byte                  yes
SMBus Write Word                 yes
SMBus Read Word                  yes
SMBus Process Call               yes
SMBus Block Write                yes
SMBus Block Read                 no
SMBus Block Process Call         no
SMBus PEC                        yes
I2C Block Write                  yes
I2C Block Read                   yes
```

### Hardware Support

I've tested on a Pi 4 Model B which is the only one I've got.

### See Also

Heavily inspired by #261 and #301. Reviewers of similar PRs: @Mic92, @samueldr, @domenkozar


samueldr:
I don't really have the ability to *test*, but given your photo, I don't doubt it's working well.

Additionally, implementation looks alright, considering.

Do you know if *other* i2c bus are enabled the same way? Any way we could add them all? (Are there even more than one i2c bus available?)

miangraham:
From https://github.com/raspberrypi/linux/tree/rpi-5.10.y/arch/arm/boot/dts:

```
$ ls -l i2c*
.rw-r--r--  156 ian users  9 Dec 03:49 -- i2c-bcm2708-overlay.dts
.rw-r--r--  894 ian users  9 Dec 03:49 -- i2c-gpio-overlay.dts
.rw-r--r-- 2.2k ian users  9 Dec 03:49 -- i2c-mux-overlay.dts
.rw-r--r--  421 ian users  9 Dec 03:49 -- i2c-pwm-pca9685a-overlay.dts
.rw-r--r-- 5.6k ian users  9 Dec 03:49 -- i2c-rtc-common.dtsi
.rw-r--r--  694 ian users  9 Dec 03:49 -- i2c-rtc-gpio-overlay.dts
.rw-r--r--  529 ian users  9 Dec 03:49 -- i2c-rtc-overlay.dts
.rwxr-xr-x 5.2k ian users  9 Dec 03:49 -- i2c-sensor-overlay.dts
.rw-r--r-- 1.3k ian users  9 Dec 03:49 -- i2c0-overlay.dts
.rw-r--r--  653 ian users  9 Dec 03:49 -- i2c1-overlay.dts
.rw-r--r--  524 ian users  9 Dec 03:49 -- i2c3-overlay.dts
.rw-r--r--  524 ian users  9 Dec 03:49 -- i2c4-overlay.dts
.rw-r--r--  532 ian users  9 Dec 03:49 -- i2c5-overlay.dts
.rw-r--r--  528 ian users  9 Dec 03:49 -- i2c6-overlay.dts
```

There are a handful of busses, problem is each one has its own unique settings in the .dts so I don't see a generic way to capture all of them and, say, just change the device number to choose a different one. For broad support there's probably a better, more automated way to do all of this but it's tough given the Pi's semi-opaque firmware build. I agree with your note from a related PR that it would be nice to have a one-line equivalent of just setting a dtparam for the simple cases.

i2c1, the ARM bus, is by far the most common I2C interface and most people use it for prototyping from what I can tell. In fact, the official pi config TUI utility raspi-config [only has an option to turn on i2c1](https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L902) (i2c_arm is an alias for i2c1), so I'm guessing >90% of I2C usage should be covered by supporting that one case.

2nd place is the VideoCore bus under i2c0 which is a bit weird and maps to a different number in the final device tree. With a quick test I was able to spin it up and query caps without issue, so I can easily copy-paste a separate setting for that one while I'm at it. The wider I spread the support the thinner my testing capacity gets so I'm hesitant to do much more rather than just encouraging others to use what's here as an example and expand to meet concrete personal needs.

samueldr:
Nah, it's fine not to, it was an open question: I did not know.

And looking at the dts for the overlays, I agree it doesn't look trivial enough.

I'd rather someone carefully implements it for their need, rather than have a batch of maybe-correct-but-unverified changes.

miangraham:
Sounds reasonable. Thanks for the feedback. I'm doing a few quick tests to see if I can simplify inlining the .dts a bit (a helper function for the easy one-liner case would be nice at least) and will undraft once I'm satisfied.

miangraham:
Added some tweaks and support for i2c0. Retested both.

After some testing, turns out my .dts didn't need to specify as much as I thought so a shared function is easy after all.

The helper fn could also be used for audio.nix in theory but it changes the .dtbo name which felt like it might break someone. Thought better of touching it for now.

I'm new here so shout if I'm being dumb about any style/factoring issues. Also let me know if you guys don't squash, I can rebase and reopen if needed.

samueldr:
I would put the trivial i2c definitions in a single file (named i2c.nix) and inline the `simple-pi4-overlay` in it for now. Otherwise looks alright I think?

I guess it would be nice to provide the helper function in some form or other to re-use, but that may not be entirely right. Anyway it would be better to figure out if we can use the source files as-they-are at some point, instead of managing them locally.

miangraham:
Good call, here's a pass. Dovetailing them eliminates some repetition.

samueldr (approved):
Looks alright, I can't test as I don't know what to do with an i2c bus :).

PR #347, opened 2021-12-11
Add Dell Inspiron 5509

maydayv7:
- [x] Ran `tests/run.py` successfully

### DIFF
--- a/dell/inspiron/5509/default.nix
+++ b/dell/inspiron/5509/default.nix
@@ -1,0 +1,23 @@
+{ lib, ... }:
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+  ];
+
+  # Essential Firmware
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Cooling Management
+  services.thermald.enable = lib.mkDefault true;
+
+  services.xserver = {
+    # GPU Driver Setting
+    videoDrivers = lib.mkDefault [ "intel" ];
+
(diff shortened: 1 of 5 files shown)


dell/inspiron/5509/default.nix
```diff
@@ -0,0 +1,23 @@
...
+
+  # Essential Firmware
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Cooling Management
+  services.thermald.enable = lib.mkDefault true;
+
+  services.xserver = {
+    # GPU Driver Setting
+    videoDrivers = lib.mkDefault [ "intel" ];
```
Mic92, line 17: We usually modsettings for intel these days. Any reason the intel xorg driver is required here?

maydayv7: Sorry, I had just copy-pasted it without realizing it, even my original configuration uses `modesetting`, I have changed it now!

dell/inspiron/5509/default.nix
```diff
@@ -0,0 +1,23 @@
...
+
+  # Essential Firmware
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Cooling Management
+  services.thermald.enable = lib.mkDefault true;
+
+  services.xserver = {
+    # GPU Driver Setting
+    videoDrivers = lib.mkDefault [ "modesetting" ];
```
Mic92, line 17: This is actually the default:

```suggestion
```

maydayv7: Okay, I've removed it, thanks!

Issue #352, opened 2021-12-26
Remove references to dezgeg.me


arkivm:
The binary cache hosted at http://nixos-arm.dezgeg.me/channel is not longer functional and redirects to a landing page.

samueldr:
List:

 - https://github.com/NixOS/nixos-hardware/search?q=dezgeg

Which gives:

 - https://github.com/NixOS/nixos-hardware/blob/master/raspberry-pi/2/README.org
 - https://github.com/NixOS/nixos-hardware/blob/2a76e1204f3a605f8d8d2f323671e1a295a5246d/inversepath/usbarmory/default.nix#L28-L31

For content to fix.

Mic92:
This has been resolved.

PR #353, opened 2021-12-28
drop usbarmory

Mic92:
The kernel has not been updated in ages (since 2016) and the used cached is
no longer reachable. Hence I assume this is no longer used. cc @ehmry


Mic92:
related to https://github.com/NixOS/nixos-hardware/issues/352

Mic92:
fixes https://github.com/NixOS/nixos-hardware/issues/352

PR #356, opened 2022-01-09
add xps 17 9710

NelsonJeppesen:
# Add XPS 17 9710

Mostly just to share the fix for non-working sound

Tested with Kernel 5.15.12 on the non-nvidia xps17

### DIFF
--- a/dell/xps/17-9710/intel/default.nix
+++ b/dell/xps/17-9710/intel/default.nix
@@ -1,0 +1,13 @@
+{ lib, pkgs, ... }: {
+  imports = [
+    ../../../../common/cpu/intel
+    ../../../../common/pc/laptop
+    ../../../../common/pc/ssd
+  ];
+
+  # Includes the Wi-Fi and Bluetooth firmware
+  hardware.enableRedistributableFirmware = true;
+
+  # Requires at least 5.12 for working sound
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.12") (lib.mkDefault pkgs.linuxPackages_latest);
+}
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ See code for all available configurations.
 | [Dell XPS 15 9560, nvidia only](dell/xps/15-9560/nvidia)            | `<nixos-hardware/dell/xps/15-9560/nvidia>`         |
 | [Dell XPS 15 9500](dell/xps/15-9500)                                | `<nixos-hardware/dell/xps/15-9500>`                |
 | [Dell XPS 15 9500, nvidia](dell/xps/15-9500/nvidia)                 | `<nixos-hardware/dell/xps/15-9500/nvidia>`         |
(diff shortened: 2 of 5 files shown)


dell/xps/17-9710/intel/default.nix
```diff
@@ -0,0 +1,27 @@
...
+    ../../../../common/pc/ssd
+  ];
+
+  # patch the kernel to enable sound
+  # Tested with 5.15.12
+  boot.kernelPatches = lib.mkDefault [{
+    name = "xps17-9710-fix-sound";
+    patch = null;
+    extraConfig = ''
+      SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES y
```
Mic92, line 14: How does this option fix sound? https://cateee.net/lkddb/web-lkddb/SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES.html

Mic92: Ok. get it. It's a dependency for `SND_SOC_INTEL_SOUNDWIRE_SOF_MACH`.
Can you make a PR for nixpkgs to include both options in nixpkgs master?

dell/xps/17-9710/intel/default.nix
```diff
@@ -0,0 +1,27 @@
...
+  }];
+
+  # Includes the Wi-Fi and Bluetooth firmware
+  hardware.enableRedistributableFirmware = true;
+
+  # Requires at least 5.12 for working sound
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.12") (lib.mkDefault pkgs.linuxPackages_latest);
+
+  # Allows for updating firmware via `fwupdmgr`.
+  services.fwupd.enable = true;
```
Mic92, line 26: Maybe fwupd is supported by this device in the README instead and leave it to the user if they want to enable it permanently or not. Without documentation that this service exists, it wouldn't be very useful anyway.

NelsonJeppesen: :+1: Sounds reasonable; I'll take this as an excuse to create readme

dell/xps/17-9710/intel/default.nix
NelsonJeppesen: Add it to `nixpkgs`, like here? https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/linux/kernel/common-config.nix#L290

Mic92: Yes.

NelsonJeppesen: @Mic92 I've created and tested a nixpkgs pr locally
https://github.com/NixOS/nixpkgs/pull/154220

NelsonJeppesen:
Moving to draft, pending approval or rejection of nixpkgs pr
https://github.com/NixOS/nixpkgs/pull/154220

Mic92:
Now we just need to drop the kernel options from this PR: https://github.com/NixOS/nixpkgs/pull/154287

NelsonJeppesen:
Added simple wiki
Removed kernel patch

Mic92:
Thanks!

PR #359, opened 2022-01-10
thinkpad/x1/9th-gen: enable fwupd

brandonweeks:
The X1 Gen 9 supports `fwupd` for multiple types of firmware. I've successfully updated UEFI, disk, and touchpad firmware using `fwupdmgr`.

### DIFF
--- a/lenovo/thinkpad/x1/9th-gen/default.nix
+++ b/lenovo/thinkpad/x1/9th-gen/default.nix
@@ -9,4 +9,7 @@
   boot.kernelPackages = lib.mkIf
     (lib.versionOlder pkgs.linux.version "5.15")
     (lib.mkDefault pkgs.linuxPackages_latest);
+  
+
+  services.fwupd.enable = true;
 }


lenovo/thinkpad/x1/9th-gen/default.nix
```diff
@@ -9,4 +9,7 @@
   boot.kernelPackages = lib.mkIf
     (lib.versionOlder pkgs.linux.version "5.15")
     (lib.mkDefault pkgs.linuxPackages_latest);
+  
+
+  services.fwupd.enable = true;
```
SuperSandro2000, line 14:
```suggestion

  services.fwupd.enable = true;
```

Mic92:
So far we mainly advertise support for fwupd in a README related to the hardware. Just enabling the service would not be so useful because user are not aware of its existence and not everyone would have it enabled at all times.
Also see https://github.com/NixOS/nixos-hardware/pull/356#discussion_r780736483

brandonweeks:
Closing for now until I figure out what generations of Thinkpad X1 Carbon machines to document as supporting `fwupd`.

PR #365, opened 2022-01-24
hp/elitebook/2560p: init

aviallon:
First HP device? Interesting.

Add some required tweaks to make things work properly, and some toggle-able config for the iGPU.


hp/elitebook/2560p/igpu.nix
```diff
@@ -0,0 +1,28 @@
+{ config, pkgs, lib, ... }:
+with lib;
+let
+  xcfg = config.services.xserver;
+  cfg = config.hardware.hp.elitebook.graphics;
+in
+{
```
Mic92, line 7: Also I would be interested why you want to toggle the iGPU? Does the board also comes with dedicated graphic cards?

hp/elitebook/2560p/default.nix
```diff
@@ -0,0 +1,22 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  xcfg = config.services.xserver;
+in
+{
+  imports = [
+    ../../../common/cpu/intel
```
Mic92, line 8: By including this profile, you are currently enabling intel graphics unconditionally.

hp/elitebook/2560p/igpu.nix
aviallon: Actualy, I don't want to toggle the iGPU, but only the hardware acceleration (vaapi, etc).
Someone might want to disable it for some reason.

hp/elitebook/2560p/default.nix
aviallon: See above comment.

hp/elitebook/2560p/igpu.nix
aviallon: Oh wait, I'm stupid.

Mic92:
fixed in https://github.com/NixOS/nixos-hardware/pull/371

Issue #362, opened 2022-01-18
Surface Pro 8 support


kiana-S:
The `microsoft/surface` module currently does not provide support for the keyboard and touchpad of the Microsoft Surface Pro 8, despite the fact that the [linux-surface feature matrix](https://github.com/linux-surface/linux-surface/wiki/Supported-Devices-and-Features#feature-matrix) states that it should.

I'm not entirely sure why this is, but I believe that it is because that module uses the Linux 5.13 kernel, and the patches for the Surface Pro 8 are [only present on release 5.14](https://github.com/linux-surface/linux-surface/issues/616#issuecomment-955812230) (I could be wrong on that, though).

Updating the Linux kernel and patches to a newer version may solve this issue.

Mic92:
Patches are welcome if you update the linux kernel linked here.

kiana-S:
Fixed by #368 (merged into #385)

PR #368, opened 2022-01-25
Surface Linux kernel 5.13.4 --> 5.16.11

kiana-S:
Potentially fixes issue #362


kiana-S:
The build process seems to fail during patching. I don't see any reason why the 5.13 patches should work and the 5.16 patches don't, other than the 5.16 patches being incorrect in some way.

Mic92:
Are you still working on this?

kiana-S:
I've been trying on-and-off for a while now, but the patches are failing and I'm not sure why.

I'll look at it again this afternoon, and if I can't get the patches to work I'll probably close it.

kiana-S:
After trying and failing to get the patches to work properly, I decided to reformat the entire linux derivation to use the pre-patched kernel at [linux-surface/kernel/v5.16-surface](https://github.com/linux-surface/kernel).

I have no idea if this actually fully works, but at least the kernel starts building.

Mic92:
Ok. Let me know if the resulting kernel runs stable for you.

kiana-S:
Complete success! The kernel runs stable, and #362 is fixed. I'll submit the pull request for review at this point.

microsoft/surface/repos.nix
```diff
@@ -1,9 +1,7 @@
 { lib, pkgs, fetchFromGitHub, fetchurl }: {
-  linux-surface = fetchFromGitHub {
-    owner = "linux-surface";
-    repo = "linux-surface";
-    rev = "4130746c31606aebaccef58997b650beda9049f7";
-    sha256 = "0r74z0hj3rg8bn2arh2zj8p7awk7fi14gpyrs5h44lwxw4351xky";
+  linux-surface-kernel = fetchurl {
+    url = "https://github.com/linux-surface/kernel/archive/v5.16-surface.tar.gz";
```
Mic92, line 3: This is a branch, we should only reference commits or tags here. Otherwise the checksum will break.

Mic92: Also I don't quite understand the switch away from `fetchFromGitHub` here.

kiana-S: That's my bad; I misunderstood what `fetchFromGitHub` did and accidentally thought it didn't apply in this situation (I'm still a beginner, and nix isn't easy to pick up).
I'll fix it so it points to a specific revision too.

Mic92:
merged in https://github.com/NixOS/nixos-hardware/pull/385

PR #372, opened 2022-02-02
feat(framework): Add initial support

### LINKED ISSUE #323: Framework Support
I think getting support would be cool opening an issue for tracking.

https://community.frame.work/t/using-the-ax210-with-linux-on-the-framework-laptop/1844/65

Setting the kernel to 5.12.15 for now for example rather than recommending latest.

Emiller88:
- Enable deep sleep
- Use kernel that's 5.12 or latest
- Enable fingerprint scanner
- HiDPI

closes #323

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -39,6 +39,7 @@
       dell-xps-15-9560-intel = import ./dell/xps/15-9560/intel;
       dell-xps-15-9560-nvidia = import ./dell/xps/15-9560/nvidia;
       dell-xps-17-9710-intel = import ./dell/xps/17-9710/intel;
+      framework = import ./framework;
       friendlyarm-nanopc-t4 = import ./friendlyarm/nanopc-t4;
       google-pixelbook = import ./google/pixelbook;
       gpd-micropc = import ./gpd/micropc;
--- a/framework/default.nix
+++ b/framework/default.nix
@@ -1,0 +1,27 @@
+{ lib, pkgs, ... }: {
+  imports = [
+    ../common/cpu/intel
+    ../common/pc/laptop
+    ../common/pc/laptop/ssd
+  ];
+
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -37,6 +37,7 @@
       dell-xps-15-9500 = import ./dell/xps/15-9500;
       dell-xps-15-9500-nvidia = import ./dell/xps/15-9500/nvidia;
       dell-xps-17-9710-intel = import ./dell/xps/17-9710/intel;
+      framework = import ./framework;
```
Mic92, line 40: Please also add this to the readme, so that our unittest framework picks it up.

framework/default.nix
```diff
@@ -0,0 +1,27 @@
...
+  # https://community.frame.work/t/using-the-ax210-with-linux-on-the-framework-laptop/1844/89
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.12") (lib.mkDefault pkgs.linuxPackages_latest);
+
+  # For fingerprint support
+  services.fprintd.enable = true;
+
+  # HiDPI
+  hardware.video.hidpi.enable = lib.mkDefault true;
+  services.xserver.dpi = 200;
+  environment.variables = {
```
Mic92, line 22: I think its fine to document these option but we should not set them by default, because they are not necessary with wayland/GNOME and have unintended effects.

Emiller88: I followed what was set in https://github.com/NixOS/nixos-hardware/blob/master/lenovo/thinkpad/x1/6th-gen/QHD/default.nix let me know if that doesn't work!

Mic92:
convert to draft until https://github.com/NixOS/nixos-hardware/pull/372#discussion_r799441383 is fixed.

PR #379, opened 2022-02-19
flake.nix add missing profiles

kilimnik:
I noticed, that some of the profiles where missing from the `flake.nix` file. 
I also ordered the profiles alphabetically.


Mic92:
Thanks!

PR #380, opened 2022-02-20
asus/battery.nix: fix chargeUpto after suspend/resume, make script optional

### LINKED ISSUE #373: hardware.asus.battery.chargeUpto not reliable after sleep/resume
The current implementation in https://github.com/NixOS/nixos-hardware/blob/master/asus/battery.nix seems to work fine on boot but doesn't stick after suspending/resuming.

Something like this seems to work for me.

```
  systemd.services.battery-charge-threshold = {
    wantedBy = [ "multi-user.target" "suspend.target" ];
    after = [ "multi-user.target" "suspend.target" ];
    description = "Set the battery charge threshold";
    startLimitBurst = 5;
    serviceConfig = {
      Type = "oneshot";
      Restart = "on-failure";
      ExecStart = "/bin/sh -c 'echo 70 > /sys/class/power_supply/BAT0/charge_control_end_threshold'";
    };
  };
```

LunNova:
Fixes chargeUpto setting getting reset after suspend/resume cycle by using a systemd service which runs after suspend instead of tmpfiles.

Fixes #373

### DIFF
--- a/asus/battery.nix
+++ b/asus/battery.nix
@@ -3,19 +3,35 @@ let
   p = pkgs.writeScriptBin "charge-upto" ''
     echo ''${0:-100} > /sys/class/power_supply/BAT0/charge_control_end_threshold
   '';
-  cfg = config.hardware.asus;
+  cfg = config.hardware.asus.battery;
 in
 
 {
-  options.hardware.asus.battery.chargeUpto = lib.mkOption {
-    description = "Maximum level of charge for your battery, as a percentage.";
-    default = 100;
-    type = lib.types.int;
+  options.hardware.asus.battery = {
+    chargeUpto = lib.mkOption {
+      description = "Maximum level of charge for your battery, as a percentage.";
+      default = 100;
+      type = lib.types.int;
+    };
+    addChargeUptoScript = lib.mkOption {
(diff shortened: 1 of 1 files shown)


asus/battery.nix
```diff
@@ -3,19 +3,35 @@ let
...
+    systemd.services.battery-charge-threshold = {
+      wantedBy = [ "local-fs.target" "suspend.target" ];
+      after = [ "local-fs.target" "suspend.target" ];
+      description = "Set the battery charge threshold to ${toString cfg.chargeUpto}%";
+      startLimitBurst = 5;
+      startLimitIntervalSec = 1;
+      serviceConfig = {
+        Type = "oneshot";
+        Restart = "on-failure";
+        ExecStart = "/bin/sh -c 'echo ${toString cfg.chargeUpto} > /sys/class/power_supply/BAT0/charge_control_end_threshold'";
```
Mic92, line 33:
```suggestion
        ExecStart = "${pkgs.runtimeShell} -c 'echo ${toString cfg.chargeUpto} > /sys/class/power_supply/BAT0/charge_control_end_threshold'";
```

LunNova:
Anything else I need to do before this can be merged?

PR #374, opened 2022-02-06
Fix/ms surface/ath10k firmware source

mexisme:
The upstream firmware source had changed, which means the hash of the source + the patch needed to be updated.

Issue #288, opened 2021-07-05
Bad source for surface go wifi


ursi:
https://github.com/NixOS/nixos-hardware/blob/41775780a0b6b32b3d32dcc32bb9bc6df809062d/microsoft/surface/firmware/surface-go/ath10k/ath10k-replace.nix#L5

^ this doesn't exist

https://github.com/NixOS/nixos-hardware/blob/41775780a0b6b32b3d32dcc32bb9bc6df809062d/microsoft/surface/repos.nix#L10-L15

https://github.com/kvalo/ath10k-firmware/tree/84b47062aab31d67156e0a3ef593a6999a12864b

samueldr:
cc @mexisme who contributed this.

mexisme:
Thanks.
Will try to look tonight; I have an idea what may have happened.

ursi:
I take it this has been fixed?

Mic92:
> I take it this has been fixed?

see https://github.com/NixOS/nixos-hardware/pull/374 maybe you can push the fix over the finish line.

Mic92:
fixed.

PR #382, opened 2022-02-20
Refactor ath10k-replace to no longer need an outputHash

ursi:
The outputHash would change and frequently cause this to not work. This
gets around the issue by using the built version firmwareLinuxNonfree as
a source instead of trying to patch it directly.


ursi:
#288 #374

PR #383, opened 2022-02-24
Add rpi4 pwm0 support

ettom:
This adds hardware PWM support on GPIO 18. There's probably a better way to do this, but this should be good enough for a start.

Usage: `hardware.raspberry-pi."4".pwm0.enable = true;`


Mic92:
Feel free to ping me again in case @samueldr is too busy to review this one.

PR #385, opened 2022-03-02
Surface kernel

Mic92:


dotlambda:
Why did you stop using patches? Now we rely on linux-surface to backport e.g. security fixes?

PR #386, opened 2022-03-06
add legion 5 (lenovo-legion-15arh05h)

sielicki:
Please feel free to drop the other two commits, if they are not welcome on the same PR.


Mic92:
Thanks. Refactoring looks also good.

PR #389, opened 2022-03-07
extract GPU stuff from common/cpu/intel to common/gpu/intel.nix

### LINKED ISSUE #388: VDPAU_DRIVER broken for NVIDIA + Intel CPU
since https://github.com/NixOS/nixos-hardware/commit/1da3ba110c8940146588c9adf52811e70f449eb1, Chrome doesn't start anymore. See https://github.com/NixOS/nixpkgs/issues/162499

Should `common/gpu/nvidia.nix` just set `VDPAU_DRIVER` to `nvidia`?

@ilkecan

bbigras:
Is this a good idea?

Should all the profiles using `common/cpu/intel` be modified to also have `common/gpu/intel`? If the same computer model can sometime have an Nvidia card, it could cause the problem I'm trying to fix.

An alternative could be to use an optional `withGPU ? true` .

Another alternative could be to use `mkOverride 900 "nvidia"` with the `common/gpu/nvidia` (see https://github.com/NixOS/nixos-hardware/issues/388#issuecomment-1060933059)

fix #388

### DIFF
--- a/common/cpu/intel/default.nix
+++ b/common/cpu/intel/default.nix
@@ -1,19 +1,8 @@
 { config, lib, pkgs, ... }:
 
 {
-  boot.initrd.kernelModules = [ "i915" ];
-
-  environment.variables = {
-    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
-  };
-
-  hardware.cpu.intel.updateMicrocode =
-    lib.mkDefault config.hardware.enableRedistributableFirmware;
-  
-  hardware.opengl.extraPackages = with pkgs; [
-    vaapiIntel
-    vaapiVdpau
-    libvdpau-va-gl
-    intel-media-driver
+  imports = [
+    ./cpu-only.nix
+    ../../gpu/intel.nix
   ];
 }
--- a/common/gpu/intel.nix
+++ b/common/gpu/intel.nix
@@ -1,0 +1,15 @@
+{ config, lib, pkgs, ... }:
+
+{
+  boot.initrd.kernelModules = [ "i915" ];
+
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
+  };
+
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiIntel
+    libvdpau-va-gl
+    intel-media-driver
+  ];
+}
--- a/flake.nix
+++ b/flake.nix
@@ -121,6 +121,7 @@
 
       common-cpu-amd = import ./common/cpu/amd;
       common-cpu-intel = import ./common/cpu/intel;
+      common-cpu-intel-cpu-only = import ./common/cpu/intel/cpu-only;
       common-cpu-intel-kaby-lake = import ./common/cpu/intel/kaby-lake;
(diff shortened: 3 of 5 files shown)


Mic92:
I would say for all desktop and laptop cpus (non-xeons) we should include both or else it will break backwards compatibility. 
We only have a few server boards in this repo, so this will be pretty much each profile that uses intel.

priegger:
I have nvidia graphics and intel graphics but only use nvidia (via a bios setting).

Right now I get some unnecessary configuration by using the cpu/intel profile. A clean solution might be to have something like platform/intel which includes the cpu and gpu parts, but this is not backwards compatible and might break a lot of setups.

Not sure if there's a good way to fix all use-cases. Maybe provide cpu/intel/cpu-only.nix with the cpu config and let cpu/intel/default.nix import cpu_only.nix and the gpu stuff?

Mic92:
Right, a cpu-only profile might be more pragmatic since we have many intel laptops, that need both. Than in setups with nvidia, one could go for the other profiles.

Mic92:
Any news on this?

bbigras:
> Any news on this?

ready

common/gpu/intel.nix
```diff
@@ -0,0 +1,16 @@
+{ config, lib, pkgs, ... }:
```
Mic92, line 1: I think we include this in all intel laptops at least.

bbigras: Did you mean to comment on `boot.initrd.kernelModules = [ "i915" ];`?

Mic92: I was more thinking about your suggestion here:

> Should all the profiles using common/cpu/intel be modified to also have common/gpu/intel? If the same computer model can sometime have an Nvidia card, it could cause the problem I'm trying to fix.

Since most of devices are laptops, this is a sane assumption. I can only think about some super micro server boards, where we can skip this.

common/cpu/intel/default.nix
```diff
@@ -1,19 +1,8 @@
...
-  environment.variables = {
-    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
-  };
-
-  hardware.cpu.intel.updateMicrocode =
-    lib.mkDefault config.hardware.enableRedistributableFirmware;
-  
-  hardware.opengl.extraPackages = with pkgs; [
-    vaapiIntel
-    vaapiVdpau
```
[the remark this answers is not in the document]
bbigras: Both `vaapiIntel` and `vaapiVdpau` or only `vaapiVdpau`?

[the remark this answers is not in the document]
gou4shi1: only vaapiVdpau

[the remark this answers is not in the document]
bbigras: Done

flake.nix
```diff
@@ -121,6 +121,7 @@
 
       common-cpu-amd = import ./common/cpu/amd;
       common-cpu-intel = import ./common/cpu/intel;
+      common-cpu-intel-cpu-only = import ./common/cpu/intel/cpu-only;
```
amesgen, line 124:
```suggestion
      common-cpu-intel-cpu-only = import ./common/cpu/intel/cpu-only.nix;
```

bbigras:
Please review.

bbigras:
Thanks!

PR #399, opened 2022-03-29
New device: Pine64's Pinebook Pro

aciceri:
@samueldr @colemickens as promised here I am to try to work on samueldr/wip-pinebook-pro#25

The structure is the same of `wip-pinebook-pro` but the default kernel now is a custom kernel based on 5.16.18 with the patches from Manjaro. This way we have a working HDMI over Type-C.
I tried directly building from megous' kernel with poor results. I asked for help in the IRC/Matrix channel and in the end they suggested me to use the Manjaro's kernel PKGBUILD as reference. 

I also slightly changed the README with this PR.

What do you think?

However please wait some days before merging PR since I want to better test this kernel, I had some kernel panics using 5.17.1 and I want to be sure they do not show up with this version .


pine64/pinebook-pro/default.nix
```diff
@@ -0,0 +1,90 @@
+# This configuration file can be safely imported in your system configuration.
+{ config, pkgs, lib, ... }:
+
+{
+  nixpkgs.overlays = [
+    (import ./overlay.nix)
+  ];
+
+  boot.kernelPackages = lib.mkDefault pkgs.pinebookpro-manjaro-kernel;
```
samueldr, line 9: I would personally suggest making `linuxPackages_latest` the default, with documentation about using the Manjaro kernel, describing what benefits it provides.

pine64/pinebook-pro/kernel/default.nix
```diff
@@ -0,0 +1,90 @@
...
+    "3339-drm-rockchip-cdn-dp-Disable-CDN-DP-on-disconnect.patch"
+    "3355-usb-typec-fusb302-Set-the-current-before-enabling-pu.patch"
+    "3359-usb-typec-fusb302-Update-VBUS-state-even-if-VBUS-int.patch"
+    "3361-usb-typec-fusb302-Add-OF-extcon-support.patch"
+    "3362-usb-typec-fusb302-Fix-register-definitions.patch"
+    "3363-usb-typec-fusb302-Clear-interrupts-before-we-start-t.patch"
+    "3364-usb-typec-typec-extcon-Add-typec-extcon-bridge-drive.patch"
+    "3365-phy-rockchip-typec-Make-sure-the-plug-orientation-is.patch"
+    "3372-phy-rockchip-inno-usb2-More-robust-charger-detection.patch"
+    "3373-usb-typec-extcon-Don-t-touch-charger-proprties.patch"
```
samueldr, line 62: Shouldn't we only include the RK3399/Pinebook Pro patches?

pine64/pinebook-pro/kernel/default.nix
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  create-patch = p: {
+    name = p;
+    patch = "${manjaro-patches}/${p}";
+  };
+
+  manjaro-kernel = pinnedNixpkgs.linuxPackagesFor (pinnedNixpkgs.linuxKernel.kernels.linux_5_16.override {
+    argsOverride = rec {
+      kernelPatches = map create-patch patches;
```
samueldr, line 78: This will leave out the default NixOS kernel patches I think, no?

(Not that I'm 100% confident, the kernel builder infra in NixOS is a huge behemoth to grok.)

pine64/pinebook-pro/default.nix
```diff
@@ -0,0 +1,90 @@
...
+    pkgs.pinebookpro-ap6256-firmware
+  ];
+
+  systemd.tmpfiles.rules = [
+    # Tweak the minimum frequencies of the GPU and CPU governors to get a bit more performance
+    # https://github.com/elementary/os/blob/05a5a931806d4ed8bc90396e9e91b5ac6155d4d4/build-pinebookpro.sh#L288-L294
+    "w- /sys/devices/system/cpu/cpufreq/policy0/scaling_min_freq - - - - 1200000"
+    "w- /sys/devices/system/cpu/cpufreq/policy4/scaling_min_freq - - - - 1008000"
+    "w- /sys/class/devfreq/ff9a0000.gpu/min_freq - - - - 600000000"
+  ];
```
samueldr, line 86: Since this is a list, overriding values is hard; is `mkAfter` enough with tmpfiles rules to override a value? These were my personally opinionated settings, but maybe users will want different settings?

pine64/pinebook-pro/kernel/default.nix
Mic92: I wouldn't mind if we could have one kernel in here for multiple devices...

pine64/pinebook-pro/kernel/default.nix
Mic92: One could still collect patches from `linuxPackages.kernel.patches` or the more appropriate version.

pine64/pinebook-pro/sound/reset-sound.rb
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env nix-shell
```
Mic92, line 1: Shouldn't be this installed? Does the sound reset need to be applied on every boot?

pine64/pinebook-pro/kernel/default.nix
aciceri: @samueldr this was my initial idea too, but people I talked with in the Pinebook Pro IRC chat told me to not worry and that applying the other patches is only useless, but not harmful. However I could try building it again using only a subset of patches (but I'm not sure whether to consider some or not).

@Mic92 this is a nice idea, if you prefer I could keep the kernel definition in another folder and import it in Pinebook Pro's configuration. But where would you put it? `common/kernel`?

pine64/pinebook-pro/sound/reset-sound.rb
aciceri: > Does the sound reset need to be applied on every boot?

No, I remember to have used that script only one time (maybe the first time I installed NixOS on the Pinebook Pro). Maybe there are some conditions that make necessary to run it again. Or maybe it's simply not necessary anymore. Since I was undecided I put it here exactly as it was in `wip-pinebook-pro`.

> Shouldn't be this installed?

Since I'm not even sure if it has to be executed I think it can be left as is. Otherwise i could add it to the overlay to make it easier to install.

@samueldr Do you know when is necessary to run this script?

pine64/pinebook-pro/default.nix
aciceri: > is `mkAfter` enough with tmpfiles rules to override a value? 

I don't know.

> These were my personally opinionated settings, but maybe users will want different settings?

I think we could stay a little opinionated, Pinebook Pro isn't a thunder, getting a bit more performance is a common wish in my opinion. However I'll refer to your experience, if you prefer I'll change it.

pine64/pinebook-pro/default.nix
aciceri: I was undecided about this point to be honest, a reason for my choice was that I don't know how fragile can be `linuxPackages_latest`. How likely is that when one update its `nixpkgs` the Pinebook Pro stops bootinkg/working?
Using a fixed and pinned version of the kernel is surely anti-pattern (compared to the Nix philosophy) but the chances of this possibility decrease. 

However I'll make this change if you deem it appropriate.

> (I guess mainly DP alt mode on type-c?)

It's the only benefit as far as I know but maybe there are other.

miniBill:
Can't wait for this to be merged ^_^
@samueldr's repo worked very well for me as soon as I added `console=tty0` to the kernel command line

pine64/pinebook-pro/sound/reset-sound.rb
samueldr: It is necessary to run only once to set things up correctly. Running it more than once will *reset* audio parameters.

This something that, AFAIUI, should be handled by ALSA UCM normally, but I don't know if there is any effort in the community to handle that.

Other distros do something that I personally think is kinda worse: they have a copy of the `/var/lib/alsa/asound.state` file that they pre-seed.

pine64/pinebook-pro/kernel/default.nix
samueldr: Yes, applying extra patches is generally "useless", that is right.

My thoughts here is that this needlessly adds surface area to review and validate, if users want to do so. Though I'm not the one to decide here, as I'll keep using the mainline kernel.

aciceri:
The custom patched kernel isn't the default anymore since I had kernel panics several times. I tried many kernel versions with many different patches subsets but I didn't solved.
However I left the custom kernel (adding `unstable` to the name) in the overlay so if someone needs the DP alt mode over Type-C (I also added an "use at your own risk" in the README).

In my opinion this can be merged now, it's not what I hoped for, but at least it's a first step.

pine64/pinebook-pro/default.nix
Mic92: We can revisit this if somebody wants to change these settings.

pine64/pinebook-pro/firmware/ap6256-firmware/default.nix
```diff
@@ -0,0 +1,41 @@
...
+, fetchFromGitLab
+, fetchurl
+, runCommandNoCC
+}:
+
+let
+  src = fetchFromGitLab {
+    domain = "gitlab.manjaro.org";
+    owner = "manjaro-arm";
+    repo = "packages/community/ap6256-firmware";
```
Mic92, line 11: The original firmware was deleted but I recovered the firmware from a fork and mirrored it here: https://github.com/nix-community/rkwifibt Can we use this? It looks a bit less random than pulling firmware files from some gitlab mirror.

pine64/pinebook-pro/kernel/default.nix
Mic92: I would say:

```suggestion
      kernelPatches = (map create-patch patches) ++ linuxPackages_5_16.kernel.patches;
```

pine64/pinebook-pro/overlay.nix
```diff
@@ -0,0 +1,11 @@
+final: super:
+
+let
+  inherit (final) callPackage kernelPatches linuxPackagesFor;
+  kernels = callPackage ./kernel { };
+in
+{
+  pinebookpro-unstable-manjaro-kernel = kernels.manjaro-kernel;
+  pinebookpro-ap6256-firmware = callPackage ./firmware/ap6256-firmware { };
+  pinebookpro-keyboard-updater = callPackage ./keyboard-updater { };
```
Mic92, line 10: I don't think any of this needs to be an overlay. The module can just do `callPackage` directly. It's also easier to read because those packages don't just pop out of no where.
This can help for situations where people share a nixpkgs instance between different nixos machines and set: `nixpkgs.pkgs = nixpkgs;`

Mic92:
> The custom patched kernel isn't the default anymore since I had kernel panics several times. I tried many kernel versions with many different patches subsets but I didn't solved. However I left the custom kernel (adding `unstable` to the name) in the overlay so if someone needs the DP alt mode over Type-C (I also added an "use at your own risk" in the README).
> 
> In my opinion this can be merged now, it's not what I hoped for, but at least it's a first step.

In this case I would not include the kernel for now as it will have no users, will bitrot and is already known not to work.

pine64/pinebook-pro/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Current state
+
+Currently everything seems to work but this is not guaranteed it will be forever.
+
+### Kernel
+
+By default this uses the latest upstream kernel but the overlay makes available `pinebookpro-unstable-manjaro-kernel`
+which is built from a pinned kernel version and makes available the DP alt mode over Type-C.
+This kernel is unstable and some sporadic kernel panics may occur, use at your own risk.
```
Mic92, line 25:
```suggestion
This module defaults to the latest upstream kernel. Attempts have been made to use manjaro's kernel that makes available the DP alt mode over Type-C, but it turns out that the kernel is unstable and some sporadic kernel panics may occur. If you want to test version, you can grab the nix expression from this commit: https://github.com/NixOS/nixos-hardware/blob/6d1bd5bc2e8b9992a3f57e416ba50fbed5516db6/pine64/pinebook-pro/kernel/default.nix
```

aciceri: I didn't know that is possible to link commits from forks as they were in the forked repo.
Then, I suppose that we can't squash all the commits into a single one before merging,

pine64/pinebook-pro/firmware/ap6256-firmware/default.nix
aciceri: Nice idea, I'm changing it.

pine64/pinebook-pro/overlay.nix
aciceri: I agree and I applied this change regarding the AP6256 firmware but I left the keyboard updater in the overlay since it's not directly used by anything in the configuration.

pine64/pinebook-pro/firmware/ap6256-firmware/default.nix
aciceri: @Mic92 there is a problem, In the repo you mirrored I can't find a needed file (`brcmfmac43456-sdio.clm_blob`). I found some repos on github that provides it but I can't find the better/more upstream one. What do you suggest?

pine64/pinebook-pro/README.md
Mic92: The commit will persisted by github, even if you squash it.

miniBill:
I think https://nixos.wiki/wiki/NixOS_on_ARM/PINE64_Pinebook_Pro should be updated now. I don't know how to do that, feel free to do it or point me in the correct direction

aciceri:
@miniBill I just added a reference to `nixos-hardware` in the wiki *before*  the other two repositories since I think the idea is that this place should become the new home for the Pinebook Pro. I think that page is really essential and any kind of information you want to add would be useful.

Maybe we should also open a PR to `wip-pinebook-pro` to change its README putting a link to this repository as *the nix-community's unopinionated overlay* (although they are the same currently).

Also, maybe this is not the best place to talk about this, still not sure about which place we could use.

Mic92:
> I think https://nixos.wiki/wiki/NixOS_on_ARM/PINE64_Pinebook_Pro should be updated now. I don't know how to do that, feel free to do it or point me in the correct direction

The wiki has a login via github for quick sign/up.

Issue #344, opened 2021-12-09
Document tuning of swappiness


AndrewKvalheim:
I noticed that the PC configurations default to either low ([10](https://github.com/NixOS/nixos-hardware/blob/4c9f07277bd4bc29a051ff2a0ca58c6403e3881a/common/pc/hdd/default.nix#L5)) or very low ([1](https://github.com/NixOS/nixos-hardware/blob/4c9f07277bd4bc29a051ff2a0ca58c6403e3881a/common/pc/ssd/default.nix#L5)) values of `vm.swappiness`. I don’t know much about this topic so it’s not obvious to me why this default is preferable to the kernel’s.

Can someone who understands the tuning please add an explanatory comment?

Thanks.

samueldr:
> swappiness
> 
> This control is used to define how aggressive the kernel will swap
> memory pages.  Higher values will increase aggressiveness, lower values
> decrease the amount of swap.  A value of 0 instructs the kernel not to
> initiate swap until the amount of free and file-backed pages is less
> than the high water mark in a zone.
> 
> The default value is 60.
> 

https://www.kernel.org/doc/Documentation/sysctl/vm.txt

Basically, it is less likely to pre-emptively swap out stuff when swappiness is low. Generally this is cargo-culted config for reducing wear on SSDs.

PR #400, opened 2022-03-31
Remove swappiness config

vroad:
Setting very low value to swapiness will not probably benefit most users, so I just removed those configs in hdd/ssd module to revert to kernel's default value.

https://askubuntu.com/questions/420026/why-is-the-default-swappiness-60-what-would-be-the-effect-if-i-lower-the-swappi/438363#438363
https://askubuntu.com/questions/184217/why-most-people-recommend-to-reduce-swappiness-to-10-20/184221#184221

Lower swappiness could just make the system slower when it runs out of memory, or even cause freezes: https://askubuntu.com/questions/184217/why-most-people-recommend-to-reduce-swappiness-to-10-20/1028193#1028193

With higher swappiness value the system copies infrequently used region of physical memory to the disk much earlier, recuding the latency of memory allocation in low-memory situations.

I did not remove common/pc/hdd/default.nix, as that could cause confusion to users using this file directly, I instead would like to show warnings to users who are using this file, but I don't know how.

Releated issue: #344

### DIFF
--- a/common/pc/hdd/default.nix
+++ b/common/pc/hdd/default.nix
@@ -1,7 +1,3 @@
 { lib, ... }:
 
-{
-  boot.kernel.sysctl = {
-    "vm.swappiness" = lib.mkDefault 10;
-  };
-}
+{}
--- a/common/pc/ssd/default.nix
+++ b/common/pc/ssd/default.nix
@@ -1,9 +1,5 @@
 { lib, ... }:
 
 {
-  boot.kernel.sysctl = {
-    "vm.swappiness" = lib.mkDefault 1;
-  };
-
   services.fstrim.enable = lib.mkDefault true;
 }


Mic92:
Given that NVME now becomes more and more common, this seems a reasonable thing to do.

vroad:
Given that swappiness config is now removed, can #344 be also closed?

vroad:
> Given that NVME now becomes more and more common, this seems a reasonable thing to do.

I doubt if older config (=swappiness lower than default) really helps in improving performance even on HDDs, it just delays page out until the system really runs out of memory. Delaying usually will just cause spike of page out and system instability.
And even with default swappiness config the kernel tries to avoid using swapping as much as possible as it's a relatively expensive operation.

PR #401, opened 2022-04-01
Add Dell XPS 17 9700

workflow:
Fixes https://github.com/NixOS/nixos-hardware/issues/290

See https://github.com/NixOS/nixos-hardware/pull/291

### DIFF
--- a/dell/xps/17-9700/common.nix
+++ b/dell/xps/17-9700/common.nix
@@ -18,7 +18,6 @@
       SND_SOC_INTEL_SOUNDWIRE_SOF_MACH m
       SND_SOC_RT1308 m
     '';
-    ignoreConfigErrors = true;
   }];
 
   boot.kernelPackages =
--- a/dell/xps/17-9700/common.nix
+++ b/dell/xps/17-9700/common.nix
@@ -1,0 +1,27 @@
+{ lib, pkgs, ... }:
+
+{
+  # Boot loader
+  boot.kernelParams = lib.mkDefault [ "acpi_rev_override" ];
+  boot.kernelModules = lib.mkDefault [ "kvm-intel" ];
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
+
+  boot.kernelPatches = [{
+    name = "enable-soundwire-drivers";
+    patch = null;
+    extraConfig = ''
+      SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES y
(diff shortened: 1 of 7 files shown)


dell/xps/17-9700/common.nix
```diff
@@ -0,0 +1,27 @@
...
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
+
+  boot.kernelPatches = [{
+    name = "enable-soundwire-drivers";
+    patch = null;
+    extraConfig = ''
+      SND_SOC_INTEL_USER_FRIENDLY_LONG_NAMES y
+      SND_SOC_INTEL_SOUNDWIRE_SOF_MACH m
+      SND_SOC_RT1308 m
```
Mic92, line 19: Can you also upstream these options to nixpkgs?

dell/xps/17-9700/common.nix
```diff
@@ -0,0 +1,27 @@
...
+    '';
+    ignoreConfigErrors = true;
```
Mic92, line 21: What error are you seeing here?

dell/xps/17-9700/common.nix
workflow: Good catch, seems to pass fine without now:
4a0cdc97c85cb941e8ff4a91226bbc92ca20cd9e

dell/xps/17-9700/common.nix
workflow: Good catch, seems to pass fine without now:
4a0cdc97c85cb941e8ff4a91226bbc92ca20cd9e

dell/xps/17-9700/common.nix
workflow: As default options for everyone?

Mic92: Yes. This is only building modules from the looks of it. So it's safe to have by default.

Mic92: Than users for this hardware no longer have to build their own kernel...

IvanMalison:
Hey @workflow Sorry to ping you here, but wondering if you are running off of unstable/which graphics option you are using and if you have started encountering any diffculties as of late.

It's not clear to me what is going on, and interestingly my display manager seems to be working, but whenever i start an X11 graphical session it seems like things arent working anymore. Not sure what could have changed to make this happen.

workflow:
@IvanMalison was still running on 22.05 when last I used the XPS, so probably wouldn't have run into it.
Maybe open a separate issue? Happy to test it at some point if you ping me there and hopefully help you reproduce 🤞

PR #408, opened 2022-04-15
Add acpi_call for lenovo thinkpad x270

koalp:
This pull request adds acpi call for lenovo thinkpad x270, thus allowing to use tlp.

I also use `<nixos-hardware/common/pc/laptop/ssd/default.nix>` but not all x270 have an SSD (though the majority have it I think)
I wonder if I can add the ssd config even if all x270 don’t have an SSD

- [x] maybe add ssd ?

### DIFF
--- a/lenovo/thinkpad/x270/default.nix
+++ b/lenovo/thinkpad/x270/default.nix
@@ -2,6 +2,8 @@
   imports = [
     ../.
     ../../../common/cpu/intel
+    ../../../common/pc/laptop/acpi_call.nix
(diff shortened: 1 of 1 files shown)


anthr76:
I just sold my x270 but it indeed have a NVME drive.

Mic92:
Our ssd profile does not do much these days: https://github.com/NixOS/nixos-hardware/blob/master/common/pc/ssd/default.nix#L4
fstrim should not cause trouble on hdd-based systems.

PR #412, opened 2022-05-14
Add Dell Latitude 5520

j-steinbach:


dell/latitude/5520/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  ];
+
+  # Essential Firmware
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  # Cooling Management
+  services.thermald.enable = lib.mkDefault true;
+
+  # Video Drivers (Intel Iris Xe)
+  services.xserver.videoDrivers = lib.mkDefault [ "intel" ];
```
Mic92, line 17: So the `intel` is better here then using the  Kernel modsettings that we do on other platforms?

Mic92: Maybe make a longer comment describing why this is the case since it's different from what we normally do.

dell/latitude/5520/default.nix
```diff
@@ -0,0 +1,21 @@
...
+
+  # Enable fwupd
+  services.fwupd.enable = true;
```
Mic92, line 20:
```suggestion
  services.fwupd.enable = lib.mkDefault true;
```

dell/latitude/5520/default.nix
j-steinbach: I have no idea which is better. But this is how it is done for the related [Dell Latitude 3480](https://github.com/NixOS/nixos-hardware/commit/d7d897fe2499df6a4ea561d7ed642d1c1f30c98e), so I thought that is is supposed to be that way. If this is not the default, I will change that.

dell/latitude/5520/default.nix
j-steinbach: Good point, will change

dell/latitude/5520/default.nix
Mic92: Maybe just drop the line and it will use the builtin kms driver. If you experience any issues that are fixed by this option. We can still add it back.

j-steinbach:
Ok, I fixed things up and hope I didn't brick anything. (Still learning about git..)

PR #429, opened 2022-06-28
thinkpad/x1-extreme/gen4: add module

xfix:
This doesn't configure NVIDIA graphics card settings as it's an optional component in this laptop and frankly I have no idea how to configure it - the problem is that HDMI port on laptops with NVIDIA card is connected to NVIDIA graphics card, and simply using `offload` won't make this work, but at the same time disabling power management is probably bad for battery if you don't need HDMI.


README.md
```diff
@@ -149,6 +149,7 @@ See code for all available configurations.
 | [Lenovo ThinkPad X1 (7th Gen)](lenovo/thinkpad/x1/7th-gen)          | `<nixos-hardware/lenovo/thinkpad/x1/7th-gen>`      |
 | [Lenovo ThinkPad X1 (9th Gen)](lenovo/thinkpad/x1/9th-gen)          | `<nixos-hardware/lenovo/thinkpad/x1/9th-gen>`      |
 | [Lenovo ThinkPad X1 Extreme Gen 2](lenovo/thinkpad/x1-extreme/gen2) | `<nixos-hardware/lenovo/thinkpad/x1-extreme/gen2>` |
+| [Lenovo ThinkPad X1 Extreme Gen 4](lenovo/thinkpad/x1-extreme/gen4) | `<nixos-hardware/lenovo/thinkpad/x1-extreme/gen4>` |
```
Mic92, line 152: Can you also add it to the flake.nix?

xfix: Done.

PR #435, opened 2022-07-15
Adds support for Thinkpad T460p

mkgvt:
This PR adds support for the Thinkpad T460p. It is identical to T460 and T460s. It's main purpose is to make it clear to users which to select to configure for that specific model without having to look at the other T460 variants and decide that they are all identical. It has been tested on a T460p.

### DIFF
--- a/lenovo/thinkpad/t460p/default.nix
+++ b/lenovo/thinkpad/t460p/default.nix
@@ -1,0 +1,9 @@
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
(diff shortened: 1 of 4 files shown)


lenovo/thinkpad/t460p/default.nix
```diff
@@ -0,0 +1,9 @@
+{ config, lib, pkgs, ... }:
```
Mic92, line 1: Please also add this to the README.md in the table and flake.nix.

PR #438, opened 2022-07-30
Adding amd pstate handling

pasqui23:


common/cpu/amd/pstate.nix
```diff
@@ -0,0 +1,7 @@
+{
+  import = [ ./. ];
```
Mic92, line 2:
```suggestion
  imports = [ ./. ];
```

common/cpu/amd/pstate.nix
```diff
@@ -0,0 +1,7 @@
+{
```
Mic92, line 1: I thinks this module should explain what it does.
Not everyone is a x86 expert. It would be also interesting to know, how this module differs from what Linux does on AMD by default.

Mic92: Upstream docs: https://www.kernel.org/doc/html/latest/admin-guide/pm/amd-pstate.html

common/cpu/amd/pstate.nix
```diff
@@ -0,0 +1,10 @@
+{ lib, config, ... }: {
+  #Enables the amd cpu scaling https://www.kernel.org/doc/html/latest/admin-guide/pm/amd-pstate.html
+  #On recent AMD cupus this can be more energy efficient
```
SuperSandro2000, line 3:
```suggestion
  # Enables the amd cpu scaling https://www.kernel.org/doc/html/latest/admin-guide/pm/amd-pstate.html
  # On recent AMD cupus this can be more energy efficient
```

common/cpu/amd/pstate.nix
```diff
@@ -0,0 +1,10 @@
...
+
+  imports = [ ./. ];
+  boot = lib.mkIf (!lib.versionOlder config.boot.kernelPackages.kernel.version "5.17") {
```
SuperSandro2000, line 6:
```suggestion
  boot = lib.mkIf (lib.versionAtLeast config.boot.kernelPackages.kernel.version "5.17") {
```

PR #440, opened 2022-08-02
raspberry-pi-4: add i2c clock-frequency option

carlossless:
Recently I had an application where I wanted to tweak the default 100khz clock frequency of the i2c0 bus to 1mhz, so I decided to expand the i2c dts'es here to include that option.

I'm still very new to nix expressions, so I am sure this code can be simplified in many different ways - feedback is appreciated.

## Test

Applying the following:
```nix
{
  options.hardware.raspberry-pi."4" = {
    i2c0 = {
      enabled = true;
      frequency = 1000000;
    };
  };
}
```

Results in:
```
xxd -g4  /proc/device-tree/soc/i2c@7e804000/clock-frequency
00000000: 000f4240 # 1_000_000
```

### DIFF
--- a/raspberry-pi/4/i2c.nix
+++ b/raspberry-pi/4/i2c.nix
@@ -2,7 +2,8 @@
 
 let
   cfg = config.hardware.raspberry-pi."4";
-  simple-overlay = { target, status }: {
+  optionalProperty = name: value: lib.optionalString (value != null) "${name} = <${builtins.toString value}>;";
+  simple-overlay = { target, status, frequency }: {
     name = "${target}-${status}-overlay";
     dtsText = ''
       /dts-v1/;
@@ -13,6 +14,7 @@ let
           target = <&${target}>;
           __overlay__ {
             status = "${status}";
+            ${optionalProperty "clock-frequency" frequency}
           };
         };
       };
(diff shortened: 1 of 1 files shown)


raspberry-pi/4/apply-overlays-dtmerge.nix
```diff
@@ -0,0 +1,30 @@
+# modification of nixpkgs deviceTree.applyOverlays to resolve https://github.com/NixOS/nixpkgs/issues/125354
```
Mic92, line 1: Its currently not clear to me where this file is used in the code. Also if it fixes an issue in the nixpkgs version, can we just merge into nixpkgs directly?

Mic92: Ah, it's used in the other PR: https://github.com/NixOS/nixos-hardware/pull/442

Mic92:
cc @06kellyjac for review.

raspberry-pi/4/apply-overlays-dtmerge.nix
Mic92: So this needs to be removed now.

06kellyjac: yup

06kellyjac:
my only thought is the frequency *could* default to null and then if null we don't add that line to the dtb but that's probably more effort than it's worth

raspberry-pi/4/apply-overlays-dtmerge.nix
carlossless: Done

raspberry-pi/4/i2c.nix
```diff
@@ -21,26 +22,52 @@ let
...
-      After a reboot, i2c-tools (e.g. i2cdetect -F 1) should work for root or any user in i2c.
-    '';
+    i2c0 = {
+      enable = lib.mkEnableOption ''
+        Turn on the VideoCore I2C bus (maps to /dev/i2c-22) and enable access from the i2c group.
+        After a reboot, i2c-tools (e.g. i2cdetect -F 22) should work for root or any user in i2c.
+      '';
+      frequency = lib.mkOption {
+        type = lib.types.int;
+        default = 100000;
```
Mic92, line 32: I assume this is the default frequency for raspberry pi just now?

carlossless: https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/bcm2711-rpi-4-b.dts#L490-L498

Yes, but I think I like @06kellyjac proposal more, where we wouldn't need to be making assumptions like that.

Mic92:
Just undraft it, when you made your change.

carlossless:
@06kellyjac @Mic92 alright I made this change, please have another look.

raspberry-pi/4/i2c.nix
```diff
@@ -2,7 +2,10 @@
 
 let
   cfg = config.hardware.raspberry-pi."4";
-  simple-overlay = { target, status }: {
+  optionalProperty = name: value: if value != null
+    then "${name} = <${builtins.toString value}>;"
+    else "";
```
Mic92, line 7:
```suggestion
  optionalProperty = name: value: lib.optionalString (value != null) "${name} = <${builtins.toString value}>;";
```

PR #444, opened 2022-08-10
pine64-pinebook-pro: remove superfluous bright/sleep keys

tomfitzhenry:
Upstream udev has had that since 2020: https://github.com/systemd/systemd/commit/f25e30dec4958ba3ab972b823210fcbc9dd231da

Tested on my Pinebook Pro. Brightness keys still work.

### DIFF
--- a/pine64/pinebook-pro/default.nix
+++ b/pine64/pinebook-pro/default.nix
@@ -43,14 +43,6 @@
   ];
 
   services.udev.extraHwdb = lib.mkMerge [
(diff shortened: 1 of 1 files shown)


aciceri:
@tomfitzhenry I can't try this now but if you tested it and these keys work even without these lines I would say we can remove them. Anyway I've no idea why this was needed in the past and now is not needed anymore.

samueldr (approved):
I have not tested, but can't see this being wrong.

PR #445, opened 2022-08-10
pine64-pinebook-pro: remove inappropriate overriding of min_freq

tomfitzhenry:
Power management should be left to a power management daemon (e.g. gnome-power-manager) or for users to set/override themselves (e.g. by setting a CPU governor).

I use governor 'schedutil', which is big.LITTLE aware, and thus can sensibly downclock, but min_freq prevents that.

### DIFF
--- a/pine64/pinebook-pro/default.nix
+++ b/pine64/pinebook-pro/default.nix
@@ -77,14 +77,6 @@
     (pkgs.callPackage ./firmware/ap6256-firmware { })
   ];
 
-  systemd.tmpfiles.rules = [
-    # Tweak the minimum frequencies of the GPU and CPU governors to get a bit more performance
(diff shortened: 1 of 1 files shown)


samueldr:
Makes sense, but also would invite discussion about how to expose such settings as a user-friendly way for systems builds where there is no userspace component to manage this.

As @aciceri said, this comes from *my opinionated* config.

This is also missing a detail about how I'm using the OP1 OPPs, basically overclocking/re-clocking the RK3399 in the Pinebook Pro [it actually adds an additional step]. So it's not even the complete view of the problem.

Mic92:
> Setting `powerManagement.cpuFreqGovernor` is a simple and effective way to control power management, e.g. `powersave` or `performance`. Shall I add that to the README?

I don't think this is needed. This is a common interface used in many places.

PR #450, opened 2022-09-01
lenovo/z13: add Lenovo Thinkpad Z13

marijanp:


toastal:
https://github.com/NixOS/nixos-hardware/blob/master/lenovo/thinkpad/default.nix#L10

```nix
services.fprintd.enable = true;
```

The fingerprint reader should likely be on by default.

toastal:
Unsure, but it may be useful to explicitly name the trackpoint

```nix
{
  hardware.trackpoint.device = "TPPS/2 Elan TrackPoint";
}
```

toastal:
TODOs:

* Trackpad support with later kernels(?)

toastal:
Missing in the `flake.nix`

toastal:
Would be nice to see `lenovo/z-series/z13`

flake.nix
```diff
@@ -109,6 +109,7 @@
       lenovo-thinkpad-x260 = import ./lenovo/thinkpad/x260;
       lenovo-thinkpad-x270 = import ./lenovo/thinkpad/x270;
       lenovo-thinkpad-x280 = import ./lenovo/thinkpad/x280;
+      lenovo-thinkpad-z13 = import ./lenovo/thinkpad/z-series/z13;
```
toastal, line 112: Suggestion:

```diff
- lenovo-thinkpad-z13 = import ./lenovo/thinkpad/z-series/z13;
+ lenovo-thinkpad-z-series = import ./lenovo/thinkpad/z-series;
+ lenovo-thinkpad-z-series-z13 = import ./lenovo/thinkpad/z-series/z13;
```

Though I could see an argument for dropping the `z-series-*` from the `z13` for brevity (though you'll likely ever only type this once in your config).

toastal (changes requested):

toastal:
> ThinkPad Z13 Gen 1 Machine types (MT): 21D2 and 20D3
> ThinkPad Z16 Gen 1 Machine types (MT): 21D4 and 20D5
> 
> - https://download.lenovo.com/pccbbs/mobiles_pdf/z13_z16_gen1_hmm_en.pdf

For other reviewers: we're not entirely sure if the `z13` naming convention from the unit or not

flake.nix
toastal: None of the other Lenovo laptops mention `*-series`, should we consider just `lenovo-thinkpad-z` + `lenovo-thinkpad-z13`??

toastal:
Booted successfully with this configuration.

toastal:
`README` is missing though.

lenovo/thinkpad/z/z13/default.nix
```diff
@@ -0,0 +1,5 @@
+{ config, lib, pkgs, ... }: {
+  imports = [
+    ../../../../lenovo/thinkpad/z
```
toastal, line 3: Maybe prefer `../z`?

PR #455, opened 2022-09-05
inspiron-7405: init

Weathercold:


Weathercold:
Don’t merge yet, I have to check whether pstate is functioning

lukegb:
> Don’t merge yet, I have to check whether pstate is functioning

Marked it as a draft to make it clear; mark it ready for review when you're ready :)

dell/inspiron/7405/default.nix
```diff
@@ -0,0 +1,32 @@
...
+
+{
+  imports = [
+    ../../../common/cpu/amd/pstate.nix
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+  ];
+
+  boot = {
+    initrd.availableKernelModules = [ "nvme" "xhci_pci" "ahci" "usb_storage" "usbhid" "sd_mod" ];
```
Mic92, line 13: Usually we don't have kernel modules that are also detected by `nixos-generate-config` since names might change over time and people might swap different storage devices in their laptops.

dell/inspiron/7405/default.nix
```diff
@@ -0,0 +1,32 @@
...
+    # TSC is unstable
+    kernelParams = [ "notsc" "trace_clock=local" ];
+    kernelModules = [ "kvm-amd" ];
```
Mic92, line 16: This is also usually detected.

Weathercold:
P-State is not working, it seems that many laptops don’t have support for it.

dell/inspiron/7405/default.nix
```diff
@@ -0,0 +1,28 @@
...
+  services.xserver = {
+    # Touchpad
+    libinput.enable = mkDefault true;
+
+    # No nvidia driver
+    videoDrivers = mkDefault [
+      "amdgpu"
+      "radeon"
+      "fbdev"
+    ];
```
rehno-lindeque, line 26: Two quick thoughts on `videoDrivers`:

* Are you sure the older `"radeon"` driver supports this gpu? 
(I think it's not listed under "Kernel Drivers" in https://linux-hardware.org/?id=pci:1002-1636-1002-1636 if I interpreted that correctly)

* Would it make sense to import `../../../common/gpu/amd`? I see some other configs do that, I'm not sure what is considered best practice here.

Mic92: Good points.

Weathercold: > Are you sure the older "radeon" driver supports this gpu?

No, it uses amdgpu. Now that you pointed it out I'm surprised how I didn't notice this earlier lol.

> Would it make sense to import ../../../common/gpu/amd?

I thought it was for discrete gpus as nearly every laptop config I saw doesn't import gpu modules.

rehno-lindeque: > I thought it was for discrete gpus as nearly every laptop config I saw doesn't import gpu modules.

Does this laptop have integrated graphics? 

E.g. I know on macbook pro 11-5, it has `i915` for intel integrated graphics and the `amdgpu` driver for discrete graphics.

Weathercold: It has only integrated graphics. I have imported `../../../common/gpu/amd` and it worked without issues.

PR #462, opened 2022-09-13
Add lenovo/thinkpad/p16s/amd/gen1

flosse:
I'm not a NixOS specialist, so I don't know if this enough but at least for me it works.

### DIFF
--- a/lenovo/thinkpad/p16s/amd/gen1/default.nix
+++ b/lenovo/thinkpad/p16s/amd/gen1/default.nix
@@ -1,0 +1,6 @@
+{ config, lib, pkgs, ... }:
+
+{
+  # For mainline support of rtw89 wireless networking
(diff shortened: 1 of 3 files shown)


lenovo/thinkpad/p16s/amd/gen1/default.nix
```diff
@@ -0,0 +1,6 @@
+{ config, lib, pkgs, ... }:
+
+{
+  # For mainline support of rtw89 wireless networking
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") pkgs.linuxPackages_latest;
```
Mic92, line 5: Please also reference this file in flake.nix and the README.md.

flosse: sure!

PR #478, opened 2022-10-07
raspberrypi/4: allow pi libs to detect pi 4

illegalprime:
Lot of libraries have checks to detect that they are running on a Raspberry Pi 4 and fail otherwise.
See https://github.com/gpiozero/gpiozero/issues/837.

This also happens with the current NixOS image, specifically:
```
$ python3
>>> from RPi import GPIO
>>> GPIO.setup(9, GPIO.IN)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Not running on a RPi!
>>>
```

This sets the CPU revision which is the newest way these libraries do their detection.
I've been running this for a while with great results, finally remembered to post something here.

### DIFF
--- a/raspberry-pi/4/cpu-revision.nix
+++ b/raspberry-pi/4/cpu-revision.nix
@@ -1,0 +1,25 @@
+{ ... }:
+{
+  hardware.deviceTree.overlays = [
+    {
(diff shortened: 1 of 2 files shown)


Mic92:
bors merge

PR #482, opened 2022-10-15
Add support for ssd to Lenovo ThinkPad T480s

dav-dap:
I just realised that ssd support is missing despite the model having ssd


Mic92:
bors merge

PR #486, opened 2022-10-20
apple/macbook-pro: add support for 14,1

vs49688:
Extracted from my system configs. There's still a few issues with this machine:

1. Audio is a lost cause. Will probably need to bribe an Apple or Cirrus engineer for the datasheet.
2. The thunderbolt module will oops upon system resume, and subsequently refuse to work until next reboot.
3. The d3cold state needs to be disabled on the NVME controller for it to wake up.
4. The Bluetooth UART (/dev/ttyS0) is created and then deleted by udev in early boot. I am yet to figure out why. Hack around it by reloading the 8250_dw module, causing it to be re-created.

Any ideas for an alternative to the `/dev/ttyS0` hack are welcome,.

### DIFF
--- a/apple/macbook-pro/14-1/README.md
+++ b/apple/macbook-pro/14-1/README.md
@@ -1,0 +1,17 @@
+# MacBook Pro 14,1
+
+## Audio
+
+Audio is a lost cause. Bribe an Apple or Cirrus engineer for the datasheet. ;)
+
+## Thunderbolt
+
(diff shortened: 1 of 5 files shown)


Mic92:
bors merge

Mic92:
bors merge

Mic92:
bors abort

yisraeldov:
I was unable to get this to work.

Re: sound, you might want to look at what they are doing in winesapOs. There everything was working, touchbar sound camera

PR #487, opened 2022-10-22
add thinkpad x1 yoga 7th-gen

MayNiklas:
The Thinkpad X1 yoga 7th gen comes with an intel Alder Lake CPU.
The default Kernel in nixos-unstable did not work with this device:
dmesg had the `hardware newer than driver` error.
After trying out one kernel at a time, at least 5.19 seems to be needed.
Without, only a TTY would show up.

The specific device being used for testing: https://www.lenovo.com/de/de/laptops/thinkpad/thinkpad-x1/ThinkPad-X1-Yoga-Gen-7-14-inch-Intel/p/LEN101T0010


Mic92:
bors merge

PR #488, opened 2022-10-26
raspberry-pi/4: add pcie_brcmstb and reset-raspberrypi to kernelParams

MatthewCroughan:
###### Description of changes

This is required for things like USB Boot to work correctly. Without it, NixOS will fail to discover the disk in Stage 1 of boot.

###### Things done

### DIFF
--- a/raspberry-pi/4/default.nix
+++ b/raspberry-pi/4/default.nix
@@ -16,7 +16,13 @@
 
   boot = {
     kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi4;
-    initrd.availableKernelModules = [ "usbhid" "usb_storage" "vc4" ];
+    initrd.availableKernelModules = [
+      "usbhid"
+      "usb_storage"
+      "vc4"
+      "pcie_brcmstb"      # required for the pcie bus to work
+      "reset-raspberrypi" # required for vl805 firmware to load
+    ];
 
     loader = {
       grub.enable = lib.mkDefault false;


raspberry-pi/4/default.nix
```diff
@@ -16,6 +16,10 @@
 
   boot = {
     kernelPackages = lib.mkDefault pkgs.linuxPackages_rpi4;
+    kernelParams = lib.mkDefault [
```
samueldr, line 19:
```suggestion
    initrd.availableKernelModules = [
```

Adding those as kernel params would do nothing.

Also note that using `mkDefault` with lists means it likely won't work as expected. Lists will generally use the unqualified priority when implemented, there is no way to compose a "`mkDefault`ed" list with one that isn't.

samueldr (changes requested):
Can you describe how you have tested these changes? They probably don't work or do nothing and you have something else solving the problem.

raspberry-pi/4/default.nix
MatthewCroughan: Ah, the reason I'm using mkDefault is because I saw this pattern in `2/default.nix`

samueldr:
I meant: how did you test *these changes in this PR*?

MatthewCroughan:
@samueldr

The https://github.com/MatthewCroughan/raspberrypi-nixos-example/ flake has two attributes:
1. `images.pi`
2. `nixosConfigurations.pi`

When USB Booting, if you were to boot the sdImage from `images.pi`, then `nixos-rebuild switch` to the `nixosConfiguration`, it would fail to boot that generation, due to missing modules. So @cleverca22 disected what the difference was between 1 and 2, and found that these were the missing modules. We found the modules given by this PR.

Admittedly, I have not actually tested that importing and using nixos-hardware from my fork here, actually makes the same difference, I will confirm this shortly in a comment.

samueldr:
> Admittedly, I have not actually tested that importing and using nixos-hardware from my fork here, actually makes the same difference, I will confirm this shortly in a comment.

When contributing untested changes, *please state it was untested*.

Especially when it's about specific details that not everyone is able to test on their end, for example because they lack the specific hardware combination to test.

MatthewCroughan:
@samueldr it was tested extensively that these modules are required for USB Boot. I'll add a PR template to avoid this in future.

MatthewCroughan:
I've created a pull request to add a template, and I've adjusted my initial comment in this PR to make use of it.

MatthewCroughan:
I have tested the changes end-to-end, by including the fork that created this PR in the flake input that produced my nixosConfiguration. Without this PR, USB Boot will fail unless the kernel modules I've added in this PR are present, on a Raspberry Pi 4 Model B.

samueldr (approved):

Mic92:
bors merge

PR #494, opened 2022-10-31
Dell Latitude 3340

quinn-dougherty:
###### Description of changes

The Dell Latitude 3340 does not have the same touchpad situation as later Latitudes. 

###### Things done

- [x] Tested the changes in your own NixOS Configuration [here](https://github.com/quinn-dougherty/configuration.nix/tree/main/agent-latitude)

(I'm not actually sure what an end to end test would be)

### DIFF
--- a/dell/latitude/3340/default.nix
+++ b/dell/latitude/3340/default.nix
@@ -1,0 +1,11 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services.xserver.videoDrivers = lib.mkDefault [ "intel" ];
+}
--- a/dell/latitude/3340/default.nix
+++ b/dell/latitude/3340/default.nix
@@ -6,6 +6,4 @@
     ../../../common/pc/laptop
     ../../../common/pc/laptop/ssd
   ];
-
-  services.xserver.videoDrivers = lib.mkDefault [ "intel" ];
 }
--- a/flake.nix
+++ b/flake.nix
@@ -102,16 +102,13 @@
       lenovo-thinkpad-t590 = import ./lenovo/thinkpad/t590;
       lenovo-thinkpad-x1 = import ./lenovo/thinkpad/x1;
       lenovo-thinkpad-x1-yoga = import ./lenovo/thinkpad/x1/yoga;
-      lenovo-thinkpad-x1-yoga-7th-gen =
-        import ./lenovo/thinkpad/x1/yoga/7th-gen;
+      lenovo-thinkpad-x1-yoga-7th-gen = import ./lenovo/thinkpad/x1/yoga/7th-gen;
       lenovo-thinkpad-x1-6th-gen = import ./lenovo/thinkpad/x1/6th-gen;
       lenovo-thinkpad-x1-7th-gen = import ./lenovo/thinkpad/x1/7th-gen;
       lenovo-thinkpad-x1-9th-gen = import ./lenovo/thinkpad/x1/9th-gen;
       lenovo-thinkpad-x1-extreme = import ./lenovo/thinkpad/x1-extreme;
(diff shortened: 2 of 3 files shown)


Mic92:
End-to-end means you are importing your nixos-hardware fork and testing the options on actual hardware.

flake.nix
```diff
@@ -24,6 +24,7 @@
       dell-inspiron-5509 = import ./dell/inspiron/5509;
       dell-inspiron-5515 = import ./dell/inspiron/5515;
       dell-inspiron-7405 = import ./dell/inspiron/7405;
+      dell-latitude-3340 = import ./dell/latitude/3340;
```
Mic92, line 27: Please also import the profile in your flake.nix.

dell/latitude/3340/default.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services.xserver.videoDrivers = lib.mkDefault [ "intel" ];
```
Mic92, line 10: By default many systems use `modsettings` driver for intel graphics because it is better maintained.
Are there specific reasons you are using intel here instead. Also see our 19.03 changelog:

> - The `intel` driver has been removed from the default list of [X.org video drivers](options.html#opt-services.xserver.videoDrivers). The `modesetting` driver should take over automatically, it is better maintained upstream and has less problems with advanced X11 features. This can lead to a change in the output names used by `xrandr`. Some performance regressions on some GPU models might happen. Some OpenCL and VA-API applications might also break (Beignet seems to provide OpenCL support with `modesetting` driver, too). Kernel mode setting API does not support backlight control, so `xbacklight` tool will not work; backlight level can be controlled directly via `/sys/` or with `brightnessctl`. Users who need this functionality more than multi-output XRandR are advised to add \`intel\` to \`videoDrivers\` and report an issue (or provide additional details in an existing one)

dell/latitude/3340/default.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
```
Mic92, line 5: This profile usually already provides sane defaults for the builtin graphic card.

flake.nix
```diff
@@ -101,13 +102,16 @@
       lenovo-thinkpad-t590 = import ./lenovo/thinkpad/t590;
       lenovo-thinkpad-x1 = import ./lenovo/thinkpad/x1;
       lenovo-thinkpad-x1-yoga = import ./lenovo/thinkpad/x1/yoga;
-      lenovo-thinkpad-x1-yoga-7th-gen = import ./lenovo/thinkpad/x1/yoga/7th-gen;
+      lenovo-thinkpad-x1-yoga-7th-gen =
+        import ./lenovo/thinkpad/x1/yoga/7th-gen;
```
Mic92, line 106: Please avoid reformatting of unrelated profiles. If we want to change the style it should better done by enforcing a code formatter test in CI.

quinn-dougherty: oops, that was my IDE automatically, will revert it out of the diff

dell/latitude/3340/default.nix
quinn-dougherty: ok i'll do some manual testing with my 3340 (which is rolling a basic gnome build) with this removed, to verify. 

The reason for line 10 here is that it's in the other latitudes' code.

flake.nix
quinn-dougherty: sorry, uncertain what this means. there are no imports in this `flake.nix`.

Mic92: Ah. sorry. I mean the README maybe.

Mic92: I actually no longer remember what I was on about here.

Mic92:
I made a new PR which was a squashed version of yours. This has been merged now: https://github.com/NixOS/nixos-hardware/pull/500

PR #502, opened 2022-11-13
common/gpu/nvidia*: Migrate to common/gpu/nvidia/* and add non-prime

### LINKED ISSUE #338: Provide NVIDIA non PRIME profile
I have been using nixos-hardware for quite some time with the common profiles, I have recently changed from an AMD to NVIDIA GPU and the first thing that I did was trying to use the NVIDIA profile, but it expects a PRIME environment, which in my desktop, I do not have.

Would be cool to have a `nixos-hardware/common/gpu/nvidia-nonprime.nix` with:

```
{ lib,  ... }:

{
  # For desktops
  # Non-Optimus mode
  # You need an MXM card
  services.xserver.videoDrivers = lib.mkDefault  [ "nvidia" ];
 
  # For high res tty
  hardware.nvidia.modesetting.enable = lib.mkDefault true;
 
  # Suspend/hibernate
  hardware.nvidia.powerManagement.enable = true;
}
``` 

Will happily do a PR if that's the path.

Ref:

- https://nixos.wiki/wiki/Nvidia
- https://nixos.org/manual/nixos/stable/#sec-x11-graphics-cards-nvidia

TLATER:
###### Description of changes

Fixes #338

Annoyingly, the flake module can't be deprecated cleanly. Suggestions welcome.

###### Things done

Note I can't properly test changes that involve prime/disable, so I'd appreciate confirmation that that works, as simple as this change looks. Also can't test the module changes, of course.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -157,8 +157,9 @@
       common-gpu-amd-sea-islands = import ./common/gpu/amd/sea-islands;
       common-gpu-amd-southern-islands = import ./common/gpu/amd/southern-islands;
       common-gpu-intel = import ./common/gpu/intel.nix;
-      common-gpu-nvidia = import ./common/gpu/nvidia.nix;
-      common-gpu-nvidia-disable = import ./common/gpu/nvidia-disable.nix;
+      common-gpu-nvidia = import ./common/gpu/nvidia/prime.nix;
+      common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
+      common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
       common-pc = import ./common/pc;
       common-pc-hdd = import ./common/pc/hdd;
       common-pc-laptop = import ./common/pc/laptop;
--- a/asus/rog-strix/g733qs/default.nix
+++ b/asus/rog-strix/g733qs/default.nix
@@ -2,7 +2,7 @@
 {
   imports = [
     ../../../common/cpu/amd/pstate.nix
-    ../../../common/gpu/nvidia.nix
+    ../../../common/gpu/nvidia/prime.nix
     ../../../common/pc/laptop
     ../../../common/pc/ssd
     ../../battery.nix
--- a/asus/zephyrus/ga401/default.nix
+++ b/asus/zephyrus/ga401/default.nix
@@ -3,7 +3,7 @@
 {
   imports = [
     ../../../common/cpu/amd
-    ../../../common/gpu/nvidia.nix
+    ../../../common/gpu/nvidia/prime.nix
     ../../../common/pc/laptop
     ../../../common/pc/ssd
   ];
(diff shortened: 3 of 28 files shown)


Mic92:
I would wait a bit in the hope that someone with an nvidia gpu can test this, if not I will try to hunt one down.

flake.nix
```diff
@@ -157,8 +157,9 @@
       common-gpu-amd-sea-islands = import ./common/gpu/amd/sea-islands;
       common-gpu-amd-southern-islands = import ./common/gpu/amd/southern-islands;
       common-gpu-intel = import ./common/gpu/intel.nix;
-      common-gpu-nvidia = import ./common/gpu/nvidia.nix;
-      common-gpu-nvidia-disable = import ./common/gpu/nvidia-disable.nix;
+      common-gpu-nvidia = import ./common/gpu/nvidia/prime.nix;
+      common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
+      common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
```
thiagokokada, line 162: Could we reverse this to keep backwards compatibility?
```suggestion
      common-gpu-nvidia = import ./common/gpu/nvidia;
      common-gpu-nvidia-nonprime = import ./common/gpu/nvidia/prime.nix;
      common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
```

flake.nix
```diff
@@ -157,8 +157,9 @@
       common-gpu-amd-sea-islands = import ./common/gpu/amd/sea-islands;
       common-gpu-amd-southern-islands = import ./common/gpu/amd/southern-islands;
       common-gpu-intel = import ./common/gpu/intel.nix;
-      common-gpu-nvidia = import ./common/gpu/nvidia.nix;
-      common-gpu-nvidia-disable = import ./common/gpu/nvidia-disable.nix;
+      common-gpu-nvidia = import ./common/gpu/nvidia/prime.nix;
+      common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
```
thiagokokada, line 161: In general LGTM, but I think we need to bikeshed this name a little. However I can also not think about a better name, so for now I am going to approve.

But if someone can think about a better name, just put it here.

thiagokokada (approved):

flake.nix
TLATER: For reference, the vendor terms would be `-offload` and `-output`, but those terms are confusing to anyone without prior knowledge about prime, so I think mentioning prime is important - that way a user with no knowledge knows what to read up on.

Just having `-output` would be very confusing anyway.

`-desktop` and `-laptop` would mimick existing terminology, but is confusing for anyone actually using their iGPU.

I'd like for it to be swapped, and have a `-prime` instead of `-nonprime`, especially since this mimics the module names, but that can't be done without a breaking change.

Ultimately I left it like this because it's the best I can come up with.

Mic92:
bors merge

Issue #504, opened 2022-11-17
Fails to build microsoft-surface Kernel


Yeshey:
I manage my system with a flake, and included the surface configuration in a similar fashion to the one suggested in the readme ([here is my configuration](https://github.com/Yeshey/nixOS-Config)).

I have a Surface Pro 7, and it would build the kernel fine, but recently, after running `nix flake update --commit-lock-file`, it fails to build the surface kernel linux-5.16.11

```
...
  CC [M]  drivers/iio/light/zopt2201.o
  LD [M]  drivers/infiniband/hw/hfi1/hfi1.o
  GEN     .version
  CHK     include/generated/compile.h
  LD      vmlinux.o
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: warning: arch/x86/power/hibernate_asm_64.o: missing .note.GNU-stack section implies executable stack
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
  MODPOST vmlinux.symvers
  MODINFO modules.builtin.modinfo
  GEN     modules.builtin
  LD      .tmp_vmlinux.btf
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: warning: arch/x86/power/hibernate_asm_64.o: missing .note.GNU-stack section implies executable stack
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: warning: .tmp_vmlinux.btf has a LOAD segment with RWX permissions
  BTF     .btf.vmlinux.bin.o
  LD      .tmp_vmlinux.kallsyms1
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: warning: .btf.vmlinux.bin.o: missing .note.GNU-stack section implies executable stack
/nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
[...]

PR #507, opened 2022-11-28
Refactor for //microsoft/surface/kernel/* - DRY, enable end-user to select from kernels.

mexisme:
###### Description of changes

Refactor the `//microsoft/surface/kernel/*` code to replace repetitive code (that's frequently cut'n'pasted) with a common pattern + functions.

Make it possible for the user to pick from a selection of kernels, esp. to support for the selection of stable vs. part-fixed vs. bleeding edge kernels, for devices with newer or unstable fixes.

Tidy-up / improve readability and consistency of code.

Supports fixes for #504.
Enables #511.

-----

I rewrote it this way for a few reasons:

- A big chunk of the 5.19.x series seemed to not compile correctly, and I had to switch several times trying to work out whether it was different patches, or kernel forks, or versions.

- When 6.0 started being released, it was not quite as stable on some of my other devices, so I limited installing it to only those ones which seemed to need it.

- More broadly, it seems there have been several times over last few years where particular kernels don't work so well for all Surface devices, so having a few versions has been helpful while bugs are shaken-out.
Therefore, it would be nice to make it possible/easy to share newer versions or different patches without breaking the "default" one, similarly to "feature flags".

- This uses built-in features of NixOS, so no complicated or clever code has been written to support it.

- Each kernel build is mostly independent, and can therefore be worked on independently.

###### Things done


Mic92:
cc @@christophcharles for review

PR #508, opened 2022-12-03
Add support for GA402

aacebedo:
###### Description of changes

This PR add the support for the ASUS Zephyrus G14 402 models.
This hardware  needs the asusd package that shall be merged from https://github.com/NixOS/nixpkgs/pull/201081.

@K900 as requested I tag you here for review.

###### Things done

### DIFF
--- a/asus/zephyrus/ga402/default.nix
+++ b/asus/zephyrus/ga402/default.nix
@@ -1,0 +1,20 @@
+{ ... }:
+
+{
+  imports = [
+    ../../../common/cpu/amd
+    ../../../common/cpu/amd/pstate.nix
+    ../../../common/gpu/amd
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/acpi_call.nix
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services = {
+    asusd.enable = true;
+  };
+
+  boot = {
+    kernelParams = [ "pcie_aspm.policy=powersupersave" "acpi.prefer_microsoft_dsm_guid=1" ];
+  };
+}
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ See code for all available configurations.
 | [Apple MacBook Pro 14,1](apple/macbook-pro/14-1)                    | `<nixos-hardware/apple/macbook-pro/14-1>`          |
 | [Asus ROG Strix G733QS](asus/rog-strix/g733qs)                      | `<nixos-hardware/asus/rog-strix/g733qs>`           |
 | [Asus ROG Zephyrus G14 GA401](asus/zephyrus/ga401)                  | `<nixos-hardware/asus/zephyrus/ga401>`             |
+| [Asus ROG Zephyrus G14 GA402](asus/zephyrus/ga402)                  | `<nixos-hardware/asus/zephyrus/ga402>`             |
 | [Asus ROG Zephyrus G15 GA503](asus/zephyrus/ga503)                  | `<nixos-hardware/asus/zephyrus/ga503>`             |
 | [Asus TUF FX504GD](asus/fx504gd)                                    | `<nixos-hardware/asus/fx504gd>`                    |
 | [BeagleBoard PocketBeagle](beagleboard/pocketbeagle)                | `<nixos-hardware/beagleboard/pocketbeagle>`        |
--- a/flake.nix
+++ b/flake.nix
@@ -17,6 +17,7 @@
       asus-fx504gd = import ./asus/fx504gd;
       asus-rog-strix-g733qs = import ./asus/rog-strix/g733qs;
       asus-zephyrus-ga401 = import ./asus/zephyrus/ga401;
+      asus-zephyrus-ga402 = import ./asus/zephyrus/ga402;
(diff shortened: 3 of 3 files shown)


asus/zephyrus/ga402/default.nix
```diff
@@ -0,0 +1,20 @@
+{ ... }:
+
+{
+  imports = [
+    ../../../common/cpu/amd
+    ../../../common/cpu/amd/pstate.nix
+    ../../../common/gpu/amd
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/acpi_call.nix
```
K900, line 9: Why is this necessary?

asus/zephyrus/ga402/default.nix
```diff
@@ -0,0 +1,20 @@
...
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services = {
+    asusd.enable = true;
+  };
+
+  boot = {
+    kernelParams = [ "pcie_aspm.policy=powersupersave" "acpi.prefer_microsoft_dsm_guid=1" ];
```
K900, line 18: The second option shouldn't be necessary with recent firmware updates.

asus/zephyrus/ga402/default.nix
```diff
@@ -0,0 +1,20 @@
...
+    ../../../common/cpu/amd
+    ../../../common/cpu/amd/pstate.nix
+    ../../../common/gpu/amd
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/acpi_call.nix
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services = {
+    asusd.enable = true;
```
K900, line 14: Should probably `mkDefault` here so people can opt out if they want.

asus/zephyrus/ga402/default.nix
```diff
@@ -0,0 +1,20 @@
+{ ... }:
+
+{
+  imports = [
+    ../../../common/cpu/amd
+    ../../../common/cpu/amd/pstate.nix
+    ../../../common/gpu/amd
+    ../../../common/pc/laptop
```
K900, line 8: This just does nothing.

aacebedo: oh why ? it does not apply their content ? 
It is done on other files like here https://github.com/NixOS/nixos-hardware/blob/master/asus/zephyrus/ga503/default.nix

asus/zephyrus/ga402/default.nix
aacebedo: agreed

asus/zephyrus/ga402/default.nix
K900: The module itself doesn't do anything on modern NixOS setups.

asus/zephyrus/ga402/default.nix
aacebedo: Which firmware ? I still saw the problem with BIOS 315. I'll try 316 then

asus/zephyrus/ga402/default.nix
aacebedo: I applied them because the options they contains see to be applicable to the platform (especially for the battery)

asus/zephyrus/ga402/default.nix
K900: I'm unable to reproduce the issue on 316 or 501.U4.

asus/zephyrus/ga402/default.nix
K900: The `acpi_call` module is only really needed for old hybrid Nvidia setups and can definitely be removed. `common/pc/laptop` can probably be left in as a no-op in case generic laptop configuration is ever desired.

aacebedo: amd* are still useful right ?

K900: Yes.

asus/zephyrus/ga402/default.nix
aacebedo: ok going to upgrade then I'll remove the flag.

I saw a difference in term of power consumption with the first but unsure what is the drawback of using it.

asus/zephyrus/ga402/default.nix
aacebedo: fixed

asus/zephyrus/ga402/default.nix
aacebedo: ok removed acpi_call module

K900:
Should be good to merge now, though I'd probably still remove the ACPI parameter and force `linuxPackages_latest` instead (which will be 6.1 very soon).

asus/zephyrus/ga402/default.nix
```diff
@@ -0,0 +1,19 @@
...
+    ../../../common/pc/laptop/ssd
+  ];
+
+  services = {
+    asusd.enable = lib.mkDefault true;
```
Mic92, line 13: We still need to wait until the channel moved before merging this.

K900: All channels should have moved by now.

Mic92:
bors merge

PR #525, opened 2022-12-24
Add LENOVO Yoga 6 13ALC6 82ND

me00001:
~~**Don't merge it, It's done for me but CONTRIBUTING.md is not implemented yet**~~ Ready for Review

###### Description of changes
Add support for the LENOVO Yoga 6 13ALC6 82ND
 Ryzen 7 5700U

###### Things done
* Easy 10 hrs of battery
* Enable or disable charging limit by TLP config or manually, `echo 1 > /sys/bus/platform/drivers/ideapad_acpi/VPC2004:00/conservation_mode`
* Enable hardware video acceleration in Firefox
* Use latest kernel to make wifi and other things work

### DIFF
--- a/lenovo/yoga/6/13ALC6/default.nix
+++ b/lenovo/yoga/6/13ALC6/default.nix
@@ -12,8 +12,7 @@
     libvdpau-va-gl
   ];
 
-  # Latest Kernel
-  boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
+  boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "6.1") pkgs.linuxPackages_latest;
 
   # energy savings
   hardware.bluetooth.powerOnBoot = lib.mkDefault false;
--- a/lenovo/yoga/6/13ALC6/default.nix
+++ b/lenovo/yoga/6/13ALC6/default.nix
@@ -12,6 +12,7 @@
     libvdpau-va-gl
   ];
 
+  # latest kernel needed to make wifi work
   boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") pkgs.linuxPackages_latest;
 
   # energy savings
--- a/lenovo/yoga/6/13ALC6/default.nix
+++ b/lenovo/yoga/6/13ALC6/default.nix
@@ -1,36 +1,24 @@
 { lib, pkgs, ...  }:
 
 {
-  boot.initrd.kernelModules = [ 
-    "amdgpu"
-    "ideapad_laptop"
+  imports = [
+    ../../../thinkpad/yoga.nix
+    ../../../../common/gpu/amd/default.nix
   ];
-  services.xserver.videoDrivers = [ "amdgpu" ];
+
+  boot.initrd.kernelModules = [ "ideapad_laptop" ];
   hardware.opengl.extraPackages = with pkgs; [
-    rocm-opencl-icd
-    rocm-opencl-runtime
-    amdvlk
     vaapiVdpau
     libvdpau-va-gl
   ];
-  hardware.opengl.extraPackages32 = with pkgs; [
-    driversi686Linux.amdvlk
-  ];
-  hardware.opengl = {
-    driSupport = lib.mkDefault true;
-    driSupport32Bit = lib.mkDefault true;
-  };
-  environment.variables.AMD_VULKAN_ICD = lib.mkDefault "RADV";
 
   # Latest Kernel
-  boot.kernelPackages = pkgs.linuxPackages_latest;
+  boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
 
+  # energy savings
   hardware.bluetooth.powerOnBoot = lib.mkDefault false;
   services.power-profiles-daemon.enable = false;
   services.tlp.enable = lib.mkDefault true;
-  # automatic screen orientation, only works in X11
-  hardware.sensor.iio.enable = true;
-  # energy savings
   boot.kernelParams = ["mem_sleep_default=deep" "pcie_aspm.policy=powersupersave"];
   powerManagement.enable = lib.mkDefault true;
   powerManagement.cpuFreqGovernor = lib.mkDefault "ondemand";
--- a/lenovo/yoga/6/13ALC6/default.nix
+++ b/lenovo/yoga/6/13ALC6/default.nix
@@ -1,0 +1,51 @@
+{ lib, pkgs, ...  }:
+
+{
+  boot.initrd.kernelModules = [ 
+    "amdgpu"
+    "ideapad_laptop"
+  ];
+  services.xserver.videoDrivers = [ "amdgpu" ];
+  hardware.opengl.extraPackages = with pkgs; [
+    rocm-opencl-icd
+    rocm-opencl-runtime
+    amdvlk
+    vaapiVdpau
+    libvdpau-va-gl
+  ];
+  hardware.opengl.extraPackages32 = with pkgs; [
+    driversi686Linux.amdvlk
+  ];
+  hardware.opengl = {
+    driSupport = lib.mkDefault true;
+    driSupport32Bit = lib.mkDefault true;
+  };
+  environment.variables.AMD_VULKAN_ICD = lib.mkDefault "RADV";
+
+  # Latest Kernel
+  boot.kernelPackages = pkgs.linuxPackages_latest;
+
+  hardware.bluetooth.powerOnBoot = lib.mkDefault false;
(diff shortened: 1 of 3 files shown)


lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,51 @@
+{ lib, pkgs, ...  }:
+
+{
+  boot.initrd.kernelModules = [ 
+    "amdgpu"
+    "ideapad_laptop"
+  ];
+  services.xserver.videoDrivers = [ "amdgpu" ];
```
Mic92, line 8: It would be nice to re-use some amd gpu hardware module rather than copy and pasting all those.

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,51 @@
...
+
+  hardware.bluetooth.powerOnBoot = lib.mkDefault false;
+  services.power-profiles-daemon.enable = false;
+  services.tlp.enable = lib.mkDefault true;
+  # automatic screen orientation, only works in X11
+  hardware.sensor.iio.enable = true;
+  # energy savings
+  boot.kernelParams = ["mem_sleep_default=deep" "pcie_aspm.policy=powersupersave"];
+  powerManagement.enable = lib.mkDefault true;
+  powerManagement.cpuFreqGovernor = lib.mkDefault "ondemand";
```
Mic92, line 36: I think we also had some energy managment options in the repository somewhere.

me00001: didn't find anything

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,39 @@
...
+  ];
+
+  boot.initrd.kernelModules = [ "ideapad_laptop" ];
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiVdpau
+    libvdpau-va-gl
+  ];
+
+  # Latest Kernel
+  boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
```
Mic92, line 16: We usually set a minimum kernel that supports the current hardware:

```suggestion
  boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "6.1") pkgs.linuxPackages_latest;
```

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,39 @@
...
+
+  # energy savings
+  hardware.bluetooth.powerOnBoot = lib.mkDefault false;
+  services.power-profiles-daemon.enable = false;
+  services.tlp.enable = lib.mkDefault true;
```
Mic92, line 21: Can you add a specific comment here why tlp was chosen and not the power-profiles-daemon?
Usually we also enable tlp using through the laptop module: `common/pc/laptop/default.nix` because this still allows gnome user to use power-management that is integrated with UI.

lenovo/yoga/6/13ALC6/default.nix
Mic92: Since this model is 2018, do we need the latest kernel here at all? If it is just your personal preference to use the latest kernel than this would be better left your own configuration.

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    CPU_SCALING_GOVERNOR_ON_AC="performance";
+    CPU_BOOST_ON_AC=1;
+    CPU_BOOST_ON_BAT=0;
+    # Stop charging battery at 60%, ideapad_laptop driver required
+    #STOP_CHARGE_THRESH_BAT0=1;
+    # Stop charging battery at 100%
+    STOP_CHARGE_THRESH_BAT0=0;
+    # GPU
+    RADEON_DPM_PERF_LEVEL_ON_AC="auto";
+    RADEON_DPM_PERF_LEVEL_ON_BAT="low";
```
Mic92, line 37: It would also be good to have some comment here on why these settings were chosen.

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,39 @@
...
+  # energy savings
+  hardware.bluetooth.powerOnBoot = lib.mkDefault false;
+  services.power-profiles-daemon.enable = false;
+  services.tlp.enable = lib.mkDefault true;
+  boot.kernelParams = ["mem_sleep_default=deep" "pcie_aspm.policy=powersupersave"];
+  powerManagement.enable = lib.mkDefault true;
+  powerManagement.cpuFreqGovernor = lib.mkDefault "ondemand";
+  services.tlp.settings = {
+    # CPU
+    CPU_SCALING_GOVERNOR_ON_BAT="ondemand";
```
Mic92, line 27: Is schedutil not the new gold standard for AMD or is this something that only exists for newer generations?

lenovo/yoga/6/13ALC6/default.nix
me00001: this model is 2021 and latest kernel needed to make wifi work

lenovo/yoga/6/13ALC6/default.nix
me00001: i am a gnome user and i found tlp to be more powerful at saving power for this device than p-p-d
the goal of buying this device is to saveing power and get at least 10 hours, but p-p-d does not do this

lenovo/yoga/6/13ALC6/default.nix
me00001: Already wrote, to save energy
https://github.com/me00001/nixos-hardware/blob/a30c4834f48a8d4da5c5d46926287cda4665078a/lenovo/yoga/6/13ALC6/default.nix#L17

lenovo/yoga/6/13ALC6/default.nix
me00001: I wouldn't mind if schedutil saves energy better

me00001: But I don't have time to test such things as the current configuration is best for getting 10 hours of battery easily

lenovo/yoga/6/13ALC6/default.nix
Mic92: It is fair enough if this is your preference but we also have other users and having this enable unconditionally had led to evaluation errors. This why we introduced this module. If you have your own preferences, you can put those in your own configuration but we should try to have settings that work for everyone.

lenovo/yoga/6/13ALC6/default.nix
Mic92: Ok. than please drop all tlp settings from now and leave them in your own configuration. It is otherwise to hard to maintain a configuration where we don't know why certain setting where chosen.

lenovo/yoga/6/13ALC6/default.nix
Mic92: Tlp is already supposed to save energy and it has default settings from which you seem to diverge. However the reason why the default settings are not good enough is not clear.

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,38 @@
...
+    ../../../../common/gpu/amd/default.nix
+  ];
+
+  boot.initrd.kernelModules = [ "ideapad_laptop" ];
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiVdpau
+    libvdpau-va-gl
+  ];
+
+  boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") pkgs.linuxPackages_latest;
```
Mic92, line 15:
```suggestion
  # latest kernel needed to make wifi work
  boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") pkgs.linuxPackages_latest;
```

lenovo/yoga/6/13ALC6/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiVdpau
+    libvdpau-va-gl
+  ];
+
+  # latest kernel needed to make wifi work
+  boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") pkgs.linuxPackages_latest;
+
+  # energy savings
+  hardware.bluetooth.powerOnBoot = lib.mkDefault false;
```
Mic92, line 19: I would also leave this for your own configuration.nix. Some people might use bluetooth keyboards for logging in...

```suggestion
```

Mic92:
bors merge

PR #526, opened 2022-12-30
Sleep resume/wait unload module script

mexisme:
###### Description of changes

Provides a common utility script for unloading a module (kernel driver) and then waiting for the unload.
This (currently) is to support Dell Inspiron 5515 (#527) and Dell XPS 13/9300 (#510) profiles.

###### Things done


mexisme:
cc @symphorien

mexisme:
Note:
#510 and #527 depend on this code — even though both have it merged into their own branches, I would assume it would be preferable to merge this one first?

common/utilities/sleep-resume/wait-unload-module/default.nix
```diff
@@ -0,0 +1,51 @@
...
+        echo ok
+      else
+        echo fail
+      fi
+    }
+
+    # NOTE: If using "systemd-inhibit", please use "--mode=delay" to minimise the risk of sleep
+    #       being completely blocked.
+    main() {
+      local module="$1"
```
mexisme, line 27:
```suggestion
    # NOTE: If using "systemd-inhibit", please use "--mode=delay" to minimise the risk of sleep
    #       being completely blocked.

    unload_module() {
      local module="$1"

      if modprobe -r "$module" 2>&1; then
        echo ok
      else
        echo fail
      fi
    }

    main() {
      local module="$1"
```

common/utilities/sleep-resume/wait-unload-module/default.nix
```diff
@@ -0,0 +1,52 @@
...
+
+writeShellApplication {
+  name = "wait-unload-module";
+  runtimeInputs = [
+    coreutils
+    kmod
+  ];
+  text = ''
+    # NOTE: If using "systemd-inhibit", please use "--mode=delay" to minimise the risk of sleep
+    #       being completely blocked.
```
mexisme, line 15:
```suggestion
    # This script attempts to unload a kernel module, waiting for it to no longer be in-use by
    # other modules.
    # It does so by trying to "modprobe -r" the module once every second until the command
    # no longer fails with a "is in use" message.
    #
    # NOTE: If using "systemd-inhibit", please use "--mode=delay" to minimise the risk of sleep
    #       being completely blocked.
    #       The setting causes "systemd-inhibit" to place a time-out on the completion of the
    #       script, and will cancel the inhibit it if the script exceeds the time-out.
```

common/utilities/sleep-resume/wait-unload-module/default.nix
```diff
@@ -0,0 +1,59 @@
...
+  text = ''
+    # This script attempts to unload a kernel module, waiting for it to no longer be in-use by
+    # other modules.
+    # It does so by trying to "modprobe -r" the module once every second until the command
+    # no longer fails with a "is in use" message.
+    #
+    # NOTE: If using "systemd-inhibit", please use "--mode=delay" to minimise the risk of sleep
+    #       being completely blocked.
+    #       The setting causes "systemd-inhibit" to place a time-out on the completion of the
+    #       script, and will cancel the inhibit it if the script exceeds the time-out.
```
mexisme, line 22:
```suggestion
    #       The setting causes "systemd-inhibit" to place a time-out on the completion of
    #       the script, and will cancel the inhibit if the script exceeds the time-out.
```

symphorien:
Recent versions of modprobe learned a --wait flag that could be used instead of this script. https://lwn.net/Articles/899526/
It's availabe in 22.11, and 22.05 stops being supported in 2 days so this looks like the way to go.

mexisme:
Oh wow, thanks @symphorien !
Will check it out and adjust my PRs as appropriate.

symphorien:
also a heads up that the implem for postResumeCommands might change https://github.com/NixOS/nixpkgs/pull/207990/files
This breaks the usage of systemd inhibit in what I contributed, and presumably what you intend to contribute (because systemd-inhibit fails when run during suspend hooks as you can't inhibit something that's already happening).

mexisme:
> also a heads up that the implementation for postResumeCommands might change [NixOS/nixpkgs#200874](https://github.com/NixOS/nixpkgs/pull/200874)
> This breaks the usage of systemd inhibit in what I contributed, and presumably what you intend to contribute (because systemd-inhibit fails when run during suspend hooks as you can't inhibit something that's already happening).

Ah, that's interesting.
I'd used that behaviour in other distros, but it seemed like Systemd units were preferred in NixOS.

If I understand the implication of https://github.com/NixOS/nixpkgs/pull/200874, I'm better off converting from system units to scripts provided to the `powerManagement.resumeCommands` and/or `powerManagement.powerUpCommands` options, as that will take advantage of these new changes seamlessly?

symphorien:
I'm not sure one way is better than another.

PR #532, opened 2023-01-05
Add common-cpus amd-kvm and intel-kvm

jcmdln:
###### Description of changes

This PR adds support for two additional common-cpu variants: `common-cpu-amd-kvm` and `common-cpu-intel-kvm`.

I currently carry these in my personal NixOS flake though I assume others may also wish to enable the relevant `kvm` module for their CPU type in a convenient way. I'll admit that it's extremely trivial for someone to set this in their own configuration which may be why these `kvm` variants do not already exist, and it saves me 3 lines in my config.

###### Things done

- [:heavy_check_mark:] Tested the changes in your own NixOS Configuration
- [:heavy_check_mark:] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via `<nixos-hardware>` or Flake input

### DIFF
--- a/common/cpu/amd/kvm.nix
+++ b/common/cpu/amd/kvm.nix
@@ -1,0 +1,4 @@
+{
+  imports = [ ./. ];
+  boot.kernelModules = [ "kvm-amd" ];
+}
--- a/common/cpu/intel/kvm.nix
+++ b/common/cpu/intel/kvm.nix
@@ -1,0 +1,4 @@
+{
+  imports = [ ./. ];
+  boot.kernelModules = [ "kvm-intel" ];
+}
--- a/flake.nix
+++ b/flake.nix
@@ -156,8 +156,10 @@
       tuxedo-infinitybook-v4 = import ./tuxedo/infinitybook/v4;
 
       common-cpu-amd = import ./common/cpu/amd;
+      common-cpu-amd-kvm = import ./common/cpu/amd/kvm.nix;
       common-cpu-amd-pstate = import ./common/cpu/amd/pstate.nix;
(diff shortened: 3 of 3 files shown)


common/cpu/amd/kvm.nix
```diff
@@ -0,0 +1,4 @@
+{
+  imports = [ ./. ];
+  boot.kernelModules = [ "kvm-amd" ];
```
Mic92, line 3: I think if we include this in nixos-hardware it can go directly into the amd and intel profile. The reason this has been not done so far is that the nixos-generate-config module already will add this to `hardware-configuration.nix`.
So this is the actual question, should we rather add a more comprehensive intel/amd profile to nixos and just enable a flag in both `hardware-configuration.nix` and `nixos-hardware` when included or do we just live with the duplication between the two?

jcmdln: I think I might have misunderstood the role of `hardware-configuration.nix` in a "pure flake" NixOS config, and what the goals are of `nixos-hardware`. If others are generating and referencing `hardware-configuration.nix` using `--impure` then I think I confused myself somewhere while reading about using a NixOS flake and made things more difficult than they needed to be.

Mic92: You should not need `--impure` if you are using flakes or tools like niv for importing nixos-hardware in your configuration.
Just now we are trying to supplement things that are not detected by `nixos-generate-config`.

Mic92:
Closing this PR for now since, we might not need these profiles given our current nixos configuration generator already covering this part. But I am open to discussion how we can improve this part and maybe change our focus.

Issue #522, opened 2022-12-21
Surface Laptop 4 AMD: touchpad, keyboard, brightness control not working


tmnvanderberg:
Hi  there, and thanks for your work. :)

On updated channel with import `<nixos-hardware/microsoft/surface>`:  keyboard, touchpad and brightness control are not working for me on my new Surface Laptop 4 AMD.

`nixos-rebuild` did cause a kernel compilation, but at this point I'm wondering if it succeeded and if I am on the right kernel, since it's reported as working on the upstream linux-surface repo. Doing another `nixos-rebuild` doesn't change anything.

Best,
Timon

tmnvanderberg:
lsmod and dmesg output here: https://gist.github.com/tmnvanderberg/a75b4431472e1348317e9cfb7e71d8e3

Mic92:
cc @mexisme

Mic92:
There are two kernels two try as well: https://github.com/NixOS/nixos-hardware/pull/511/files#diff-82eaaa50adefad877ba7f54b5a444cf4942804e8d44d474a4fbe3a0cf47c5d5dR12

tmnvanderberg:
Sorry, my nix-foo is a little limited and I can't figure out the right 'namespaces' to set `kernelVersion`.
`microsoft-suface.kernelVersion`? `microsoft.surface.kernelVersion`?

mexisme:
Hi @tmnvanderberg,
Full disclosure: I'm afraid I don't have access to an AMD Surface, so I may not be able to debug this adequately by myself.
But thanks for dropping `lsmod` and `dmesg`, and I'll take a look at that, now.

In the meantime:

Current `master` of nixos-hardware supports building kernel 6.0.11 and 5.19.17.
The default is currently 6.0.11, mostly because it has quite a few fixes — incl. several for the newer AMD 6000 series CPUs, e.g. internal keyboard  ;-)
You can override the default, and try kernel 5.19.17 by adding `microsoft-surface.kernelVersion = "5.19.17";` to your laptop's `configuration.nix` (or `flake.nix` perhaps).

tmnvanderberg:
Thanks both, setting `options.microsoft-surface.kernelVersion = "5.19.17"` atleast fixes the keyboard and trackpad.

mexisme:
> Thanks both, setting `microsoft-surface.kernelVersion = "5.19.17"` atleast fixes the keyboard and trackpad.

OK, that sounds like kernel-specific.
[...]

PR #537, opened 2023-01-10
microsoft/surface profile -- change dir structure to support easier specialisation by model

mexisme:
###### Description of changes

When I created the microsoft/surface profile, there weren't that many differences between the various models of Surface.

Now, however, we have a much wider variety of chipsets, incl. models with some of the newer AMD CPUs, and this is breaking small things in annoying ways for more people.

This is a refactor/restructure to try to make it easy to support specialisation by model.

This also:
- Creates a model specialisation of `microsoft/surface` profile for MS Surface Go.
- Creates a simple/generic model specialisation of `microsoft/surface` profile for MS Surface Pro.
- Creates a simple/generic model specialisation of `microsoft/surface` profile for MS Surface Laptop (AMD).

###### Things done


mexisme:
I expect I'll need to replace the warning in `microsoft/surface/default.nix` with something that won't break tests, @Mic92 ?

microsoft/surface/default.nix
```diff
@@ -1,46 +1,12 @@
...
 
 let
-  inherit (lib) mkDefault;
+  inherit (lib) warn;
 
 in {
   imports = [
-    ./kernel
-    ./firmware/surface-go/ath10k
+    ( warn
```
Mic92, line 8: There is a `warnings` nixos module option which might not break the test here. Otherwise we might need a workaround in our test script.

microsoft/surface/old/default.nix
```diff
@@ -0,0 +1,22 @@
...
+    ../surface-go/firmware/ath10k
+  ];
+
+  boot.extraModprobeConfig = mkDefault ''
+    options i915 enable_fbc=1 enable_rc6=1 modeset=1
+    options snd_hda_intel power_save=1
+    options snd_ac97_codec power_save=1
+    options iwlwifi power_save=Y
+    options iwldvm force_cam=N
+  '';
```
Mic92, line 18: Given this module still module still has some option is this maybe some generic intel model, that you trying to target here?

Mic92: I think as of now it's a bit unclear to where people have to migrate after switching to the old module?

microsoft/surface/old/README.md
```diff
@@ -0,0 +1,51 @@
...
+### "Common" modules
+
+All the "common" modules that were once in the top-level of the microsoft/surface profile have moved
+under the `common/` directory.
+
+Tools / services that are shared among several models are now extracted to their own module under
+`common/` and imported by `common/default.nix`.
+These new "common" modules now have an `enable` option, which is `false` by default.
+
+### "Old Behaviour" module
```
Mic92, line 36: I would use `deprecated` instead of `old` here to make the intent more clear. Old could also mean legacy hardware, or is this the intent?

microsoft/surface/old/default.nix
Mic92: Are you trying to encourage people to add their own more specific models instead of using the `old` module?

microsoft/surface/default.nix
mexisme: > There is a `warnings` nixos module option which might not break the test here. Otherwise we might need a workaround in our test script.

Yeah, it wasn't clear whether this would be preferred over `lib.warn`.
I'll take a look.

Alternatively, I can perhaps use a simpler `builtins.trace` instead?

microsoft/surface/old/default.nix
mexisme: Ah, damn:  I had intended to add doc-comments to this file, to explain its intention as well as the `old/README.md`.
I will add that.

The intention is to support the "old behaviour" that most people have probably been relying on, and provide an additional deprecation warning asking people to change.
The old behaviour included everything Surface related, and expected various kernel modules and other Surface support tools to fail gracefully depending on the hardware they were running on.

I feel like the change is big enough that having some sort-of deprecation process is appropriate.

microsoft/surface/old/README.md
mexisme: > I would use `deprecated` instead of `old` here to make the intent more clear. Old could also mean legacy hardware, or is this the intent?

Happy to change the name.

microsoft/surface/old/default.nix
mexisme: > The intention is to support the "old behaviour" that most people have probably been relying on, and provide an additional deprecation warning asking people to change. The old behaviour included everything Surface related, and expected various kernel modules and other Surface support tools to fail gracefully depending on the hardware they were running on.
> [...]
> I feel like the change is big enough that having some sort-of deprecation process is appropriate.

Now that I've had a brain-break, I'm thinking a better option would be to drop the `deprecated` module, and instead make the original `microsoft/surface/default.nix` fail with an explanation with how to update.

Is there a way to support this, and still make sure the CI tests succeed?

microsoft/surface/default.nix
mexisme: > > There is a `warnings` nixos module option which might not break the test here. Otherwise we might need a workaround in our test script.
> 
> Yeah, it wasn't clear whether this would be preferred over `lib.warn`. I'll take a look.

I tested this, and it also triggers a fail, the same as using `lib.warn` directly.

> Alternatively, I can perhaps use a simpler `builtins.trace` instead?

If I want to provide a deprecation assertion or warning, it seems like this is my only option.
And it's not a really great option, realistically.

mexisme: > There is a `warnings` nixos module option which might not break the test here. Otherwise we might need a workaround in our test script.

Ok, to hack around this, I've kept the `//microsoft/surface/default.nix` file, but removed the entry from the `//README.md` file.
I've also converted the `warnings` to `assertions` instead.

WDYT?

microsoft/surface/default.nix
```diff
@@ -1,46 +1,10 @@
...
-  imports = [
-    ./kernel
-    ./firmware/surface-go/ath10k
+{ ... }:
+
+{
+  assertions = [
+    {
+      assertion = false;
+      message = "Importing microsoft/surface/ (default.nix) directly is deprecated! See microsoft/surface/OLD-BEHAVIOUR-DEPRECATED.md for more details.";
```
Mic92, line 7:
```suggestion
      message = "Importing microsoft/surface/ (default.nix) directly is deprecated! See https://github.com/NixOS/nixos-hardware/blob/master/microsoft/surface/OLD-BEHAVIOUR-DEPRECATED.md for more details.";
```

Mic92: This is easier to click/read in the terminal.

mexisme: > This is easier to click/read in the terminal.

That's an excellent idea, thanks!

Mic92:
bors merge

mexisme:
> bors merge

Oops, sorry:  I quickly added an AMD variant, to help support #522

Mic92:
bors merge

PR #548, opened 2023-01-25
macbookpro 3,1: init

Krutonium:
###### Description of changes
Added what's needed to make the MacBook Pro 3,1 17" that I have work. (If anyone happens to know how to make Nouveau agree with this device, I'd love to know.)

###### Things done

### DIFF
--- a/apple/macbook-pro/3-1/README.wiki
+++ b/apple/macbook-pro/3-1/README.wiki
@@ -1,0 +1,5 @@
+= Apple MacBook Pro 3,1 =
+
+This laptop is a weird one. It has an nVidia 8600M GT, which is supported in driver 340, which is (at the moment) marked as broken because OpenGL is broken with it. Nouveau doesn't work either, as it fails to pull the firmware from the GPU to turn it on.
+
+That said, the rest of it all just... works.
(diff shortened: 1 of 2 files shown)


Krutonium:
....And I just realized I edited the wrong files.

Krutonium:
(if anyone does happen to be or know someone who is good with nouveau issues, hmu)

Krutonium:
So I chatted with some people in the nouveau IRC, I was wondering what the right thing to do would be if the solution, realistically, was to distribute a copy of the VBIOS, since it's difficult to extract, and the GPU is non functional without it.

apple/macbook-pro/3-1/README.wiki
```diff
@@ -0,0 +1,5 @@
+= Apple MacBook Pro 3,1 =
```
Mic92, line 1: Can you also add this to the README and the flake.nix as described in https://github.com/NixOS/nixos-hardware/blob/master/CONTRIBUTING.md#2-adding-entry ?

Mic92:
> So I chatted with some people in the nouveau IRC, I was wondering what the right thing to do would be if the solution, realistically, was to distribute a copy of the VBIOS, since it's difficult to extract, and the GPU is non functional without it.

There are some places like archive.org where these things can be uploaded... or some throw-away github account in a github repository release. This is a grey area surely but legal risks are not to high for older hardware.

toastal:
> 3,1

Why not 3.1?

Mic92:
Please reopen if you think this is ready.

PR #550, opened 2023-01-31
Add amdvlk and opencl options to amdgpu & Fix proton crash when Dual-Direct GFX enabled for lenovo legion 16ach6h

LostAttractor:
###### Description of changes
When I play the game through proton, I found that in the case of Dual-Direct GFX enabled (dGPU disabled), proton will crash directly
But in the case of hybrid, the game runs fine with or without `nvidia-offload`
After investigation, this is because when writing the specialization of Dual-Direct GFX, I did not completely remove all packages for amd igpu. I only removed `amdgpu` from `services.xserver.videoDrivers` by overriding. This is because the specialization of nix cannot implement such an operation as canceling an import. In the end, if it is enabled in Dual-Direct GFX In the absence of amd igpu, the `amdvlk` package caused the proton to crash.
In order to solve this problem, I add the option of whether to enable `amdvlk` to the configuration file of amd gpu, and open it by default, and turn it off in specialization, so as to delete `amdvlk` package and other packages for amd igpu in specialization.
At the same time, I also added an option to amdgpu's opencl runtime.

###### Things done

### DIFF
--- a/lenovo/legion/16ach6h/default.nix
+++ b/lenovo/legion/16ach6h/default.nix
@@ -7,6 +7,12 @@
     # This specialisation is for the case where "DDG" (Dual-Direct GFX, A hardware feature that can enable in bios) is enabled, since the amd igpu is blocked at hardware level and the built-in display is directly connected to the dgpu, we no longer need the amdgpu and prime configuration.
     system.nixos.tags = [ "Dual-Direct-GFX-Mode" ];
     services.xserver.videoDrivers = [ "nvidia" ]; # This will override services.xserver.videoDrivers = lib.mkDefault [ "amdgpu" "nvidia" ];
-    hardware.nvidia.prime.offload.enable = false;
+    hardware = {
+      nvidia.prime.offload.enable = false;
+      amdgpu = {
+        amdvlk = false;
+        opencl = false;
+      };
+    };
   };
 }
--- a/lenovo/legion/16ach6h/nvidia/default.nix
+++ b/lenovo/legion/16ach6h/nvidia/default.nix
@@ -3,5 +3,11 @@
 {
   imports = [ ../hybrid ];
   services.xserver.videoDrivers = [ "nvidia" ]; # This will override services.xserver.videoDrivers = lib.mkDefault [ "amdgpu" "nvidia" ];
-  hardware.nvidia.prime.offload.enable = false;
+  hardware = {
+    nvidia.prime.offload.enable = false;
+    amdgpu = {
+      amdvlk = false;
+      opencl = false;
+    };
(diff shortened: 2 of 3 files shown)


lenovo/legion/16ach6h/default.nix
```diff
@@ -7,6 +7,12 @@
     # This specialisation is for the case where "DDG" (Dual-Direct GFX, A hardware feature that can enable in bios) is enabled, since the amd igpu is blocked at hardware level and the built-in display is directly connected to the dgpu, we no longer need the amdgpu and prime configuration.
     system.nixos.tags = [ "Dual-Direct-GFX-Mode" ];
     services.xserver.videoDrivers = [ "nvidia" ]; # This will override services.xserver.videoDrivers = lib.mkDefault [ "amdgpu" "nvidia" ];
-    hardware.nvidia.prime.offload.enable = false;
+    hardware = {
+      nvidia.prime.offload.enable = false;
+      amdgpu = {
+        amdvlk = false;
+        opencl = false;
```
Mic92, line 14: Can you write a simple comment similar to what you described in your PR also here and the other case below where you disable the options?

lenovo/legion/16ach6h/nvidia/default.nix
```diff
@@ -3,5 +3,11 @@
 {
   imports = [ ../hybrid ];
   services.xserver.videoDrivers = [ "nvidia" ]; # This will override services.xserver.videoDrivers = lib.mkDefault [ "amdgpu" "nvidia" ];
-  hardware.nvidia.prime.offload.enable = false;
+  hardware = {
```
Mic92, line 6:
```suggestion
  # When I play the game through proton, I found that in the case of Dual-Direct GFX
  # enabled (dGPU disabled), proton will crash directly. But in the case of hybrid,
  # the game runs fine with or without nvidia-offload After investigation, this is
  # because when writing the specialization of Dual-Direct GFX, I did not completely
  # remove all packages for amd igpu. I only removed amdgpu from
  # services.xserver.videoDrivers by overriding. This is because the specialization
  # of nix cannot implement such an operation as canceling an import. In the end, if
  # it is enabled in Dual-Direct GFX In the absence of amd igpu, the amdvlk package
  # caused the proton to crash. In order to solve this problem, I add the option of
  # whether to enable amdvlk to the configuration file of amd gpu, and open it by
  # default, and turn it off in specialization, so as to delete amdvlk package and
  # other packages for amd igpu in specialization. At the same time, I also added an
  # option to amdgpu's opencl runtime.
  hardware = {
```

Mic92:
bors merge

PR #554, opened 2023-02-11
gpd/win-2: init

necrophcodr:
###### Description of changes

Adds the bare minimums for running NixOS on a GPD WIN 2 without the system being all weird and the display being incorrectly rotated.

###### Things done

### DIFF
--- a/gpd/win-2/default.nix
+++ b/gpd/win-2/default.nix
@@ -1,0 +1,19 @@
+{ config, lib, pkgs, ...}:
+{
+  imports = [
+    ../../common/cpu/intel
+    ../../common/pc/ssd
+    ];
+
+  boot.kernelParams = [
+    "fbcon=rotate:1"
+    "video=eDP-1:panel_orientation=right_side_up"
+  ];
+
+  services.tlp.enable = lib.mkDefault ((lib.versionOlder (lib.versions.majorMinor lib.version) "21.05")
+                                       || !config.services.power-profiles-daemon.enable);
+
+  # Required for grub to properly display the boot menu.
+  boot.loader.grub.gfxmodeEfi = lib.mkDefault "720x1280x32";
+
+}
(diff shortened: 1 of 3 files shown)


gpd/win-2/default.nix
```diff
@@ -0,0 +1,18 @@
+{ lib, pkgs, ...}:
+{
+  imports = [
+    ../../common/cpu/intel
+    ../../common/pc/laptop
+    ../../common/pc/ssd
+    ];
+
+  boot.initrd.availableKernelModules = [ "xhci_pci" "ahci" "usbhid" "usb_storage" "sd_mod" "sdhci_pci" ];
```
Mic92, line 9: For now we only extend nixos-generate-config since the needed kernel modules may change with kernel versions. This might be open for debate but than we should consistently handle this in all hardware modules in the same way.

```suggestion
```

necrophcodr: @Mic92 I'd really not like to remove the kernel modules here since the builtin SD card reader simply wouldn't work without them. If it's dependent on kernel versions we could set this based on kernel versioning instead?

I'm not sure what the best solution is, but removing functionality of the laptop without informing the user seems like an actually bad idea to me.

Mic92: If you run `nixos-generate-config` on the device are those not added by the script in `hardware-configuration.nix`?

necrophcodr: Oh, I misunderstood. If it is required that the user boots a LiveCD first and runs nixos-generate-config before using the code in this repository, then it should definitely be removed yes. I've done so.

Mic92:
bors merge

gpd/win-2/default.nix
Mic92: Ideally we could even add more auto-detection to the `nixos-generate-config` code or some similar implementation or have this script just put hardware metadata into the configuration that can be than matched against a hardware module. But until this happens this repository is a nice & easy way to put the `archlinux wiki` into code form.

PR #556, opened 2023-02-13
Add NXP i.MX8 SOC family support.

unbel13ver:
Currently there are two devices supported:
* imx8qm-mek
* imx8qxp-mek

Signed-off-by: Ivan Nikolaenko <ivan.nikolaenko@unikie.com>

###### Description of changes
Changes made in the way it makes very easy to add another mx8-based device or device-specific configuration.

###### Things done

### DIFF
--- a/nxp/imx8qm-mek/overlay.nix
+++ b/nxp/imx8qm-mek/overlay.nix
@@ -1,0 +1,3 @@
+final: _prev: {
+  inherit ( final.callPackage ./bsp/u-boot/imx8/imx-uboot.nix { pkgs = final; targetBoard = "imx8qm"; }) ubootImx8 imx-firmware;
+}
--- a/nxp/imx8qxp-mek/overlay.nix
+++ b/nxp/imx8qxp-mek/overlay.nix
@@ -1,0 +1,3 @@
+final: prev: {
+  inherit ( final.callPackage ./bsp/u-boot/imx8/imx-uboot.nix { pkgs = final; targetBoard = "imx8qxp"; }) ubootImx8 imx-firmware;
+}
--- a/flake.nix
+++ b/flake.nix
@@ -139,6 +139,7 @@
       microsoft-surface-pro-3 = import ./microsoft/surface-pro/3;
       msi-gs60 = import ./msi/gs60;
       msi-gl62 = import ./msi/gl62;
+      nxp-imx8qm-mek = import ./nxp/imx8qm-mek;
       omen-en00015p = import ./omen/en00015p;
       onenetbook-4 = import ./onenetbook/4;
       pcengines-apu = import ./pcengines/apu;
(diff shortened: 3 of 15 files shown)


necrophcodr:
@unbel13ver is the firmware files completely public domain to use, or are they under some license agreement that the user needs to agree to?

Mic92:
I think in either case we would be fine to have the nix expression in this repository because the binaries are public accessible and we are not hosting repackaged versions of them (hence no copy). Might be nice though to have a license in `meta.license` if available.

unbel13ver:
> @unbel13ver is the firmware files completely public domain to use, or are they under some license agreement that the user needs to agree to?

Thanks for quick reply. Yes, the firmware is publicly available, but one needs to accept NXP's EULA:
> Welcome to NXP imx-seco-3.8.6.bin
> 
> You need to read and accept the EULA before you can continue.
> 
> LA_OPT_NXP_Software_License v34 February 2022

I am not quilte familiar with legal stuff, so I would appreciate a lot if you point me what kind of license information I should add.

Mic92:
Does not look like it's redistributable so I would go with `lib.licenses.unfree`

unbel13ver:
> Does not look like it's redistributable so I would go with `lib.licenses.unfree`

Addressed

Mic92:
bors merge

nxp/imx8qm-mek/overlay.nix
```diff
@@ -0,0 +1,3 @@
+final: _prev: {
+  inherit ( final.callPackage ./bsp/u-boot/imx8/imx-uboot.nix { pkgs = final; targetBoard = "imx8qm"; }) ubootImx8 imx-firmware;
```
tattvam, line 2: Is `./bsp/u-boot/imx8/imx-uboot.nix ` correct path?

nxp/imx8qxp-mek/overlay.nix
```diff
@@ -0,0 +1,3 @@
+final: prev: {
+  inherit ( final.callPackage ./bsp/u-boot/imx8/imx-uboot.nix { pkgs = final; targetBoard = "imx8qxp"; }) ubootImx8 imx-firmware;
```
tattvam, line 2: `./bsp/u-boot/imx8/imx-uboot.nix` no u-boot/imx8 directory in basp. Also relative path of bsp doesn't seem ok.

nxp/imx8qm-mek/overlay.nix
unbel13ver: Hi, sorry for this. I'll fix paths and improve README

nxp/imx8qxp-mek/overlay.nix
unbel13ver: Relative paths are ok as long as they point within the repo directory structure.

PR #563, opened 2023-02-22
t14s,p14s: remove rtw89-firmware

ncfavier:
###### Description of changes

The package was removed in https://github.com/NixOS/nixpkgs/pull/217196 and is now present in linux-firmware.

cc @lovesegfault @QuantMint, should we guard this against some version check? I'm not sure what exactly is needed for wifi to work: is having a 5.16+ kernel enough or would people *also* need a 20230210+ linux-firmware?

###### Things done

### DIFF
--- a/lenovo/thinkpad/p14s/amd/gen2/default.nix
+++ b/lenovo/thinkpad/p14s/amd/gen2/default.nix
@@ -12,7 +12,7 @@
 
 
   # Wifi support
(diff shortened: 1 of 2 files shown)


lovesegfault:
You need both, yeah, so a kernel version guard would be good here

ncfavier:
Added a guard against the linux-firmware version since that's what determines whether the package exists or not.

PR #566, opened 2023-02-23
Remove rtw89 firmware

etu:
###### Description of changes
The rtw89-firmware package have been removed and replaced with a throw, however I didn't just want to remove it in case people have older versions of nixpkgs so I made it conditional if the package evals or not.

With this in place I updated my flake input to use that branch and I can build the system without getting the throw error from this package.

###### Things done


HeapUnderfl0w:
Isn't this PR related to #563?
(also not to be bikeshedding, but i feel like switching on the version is better than trying to evaluate it).

etu:
Seems to be a duplicate yes

PR #569, opened 2023-02-25
Enable thermald for Dell XPS 15 7590

ghisvail:
###### Description of changes

Enable thermald.

###### Things done

### DIFF
--- a/dell/xps/15-7590/default.nix
+++ b/dell/xps/15-7590/default.nix
@@ -1,3 +1,5 @@
+{ lib, ... }:
+
 {
   imports = [
     ../../../common/cpu/intel
@@ -20,6 +22,9 @@
 
   # Prevent small EFI partiion from filling up
   boot.loader.grub.configurationLimit = 10;
+
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
 
   # The 48.ucode causes the Killer wifi card to crash.
   # The iwlfwifi-cc-a0-46.ucode works perfectly


ghisvail:
I also noticed that previous and next XPS models have ` boot.kernelParams = lib.mkDefault [ "acpi_rev_override" ];` but not this one. I wonder whether it's intentional or a mistake.

Mic92:
bors merge

Mic92:
Don't know if we need the acpi flag here or not: 
https://cateee.net/lkddb/web-lkddb/ACPI_REV_OVERRIDE_POSSIBLE.html

ghisvail:
Another thing is that the `nixpkg.overlay` section to override the WiFI module does not seem to be required anymore. I was able to configure my laptop just fine without it.

Should I submit a PR to remove it?

Mic92:
> Another thing is that the `nixpkg.overlay` section to override the WiFI module does not seem to be required anymore. I was able to configure my laptop just fine without it.
> 
> Should I submit a PR to remove it?

sure.

dell/xps/15-7590/default.nix
```diff
@@ -21,6 +21,9 @@
   # Prevent small EFI partiion from filling up
   boot.loader.grub.configurationLimit = 10;
 
+  # This will save you money and possibly your life!
+  services.thermald.enable = lib.mkDefault true;
```
Mic92, line 25: `lib` needs to be added to the header of this module:

```
+ { lib, ...}:
```

ghisvail: my bad, missed it when copy pasting from my config

Mic92:
bors merge

PR #571, opened 2023-02-27
common/gpu/intel*: Migrate to common/gpu/intel/* and add disable

Kiskae:
###### Description of changes

Recently had to disable the iGPU, but it turns out that even if you blacklist the module it will still get loaded by the KMS subsystem. Added a module that disables both modelled after the nvidia modules.

###### Things done

### DIFF
--- a/common/gpu/intel/default.nix
+++ b/common/gpu/intel/default.nix
@@ -1,0 +1,15 @@
+{ config, lib, pkgs, ... }:
+
+{
+  boot.initrd.kernelModules = [ "i915" ];
+
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
+  };
+
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiIntel
+    libvdpau-va-gl
+    intel-media-driver
+  ];
+}
--- a/common/cpu/intel/default.nix
+++ b/common/cpu/intel/default.nix
@@ -1,6 +1,6 @@
 {
   imports = [
     ./cpu-only.nix
-    ../../gpu/intel.nix
+    ../../gpu/intel
   ];
 }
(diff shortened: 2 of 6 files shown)


common/gpu/intel/default.nix
```diff
@@ -0,0 +1,15 @@
...
+
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
+  };
+
+  hardware.opengl.extraPackages = with pkgs; [
+    vaapiIntel
+    libvdpau-va-gl
+    intel-media-driver
+  ];
```
SuperSandro2000, line 14:
```suggestion
  hardware.opengl.extraPackages = with pkgs; [
    intel-media-driver
    intel-ocl
    vaapiIntel
  ];
```

see https://search.nixos.org/options?channel=unstable&show=hardware.opengl.extraPackages&from=0&size=50&sort=relevance&type=packages&query=hardware.opengl.extraPackages

Kiskae: `intel-ocl` would be a part of the intel cpu module, not the gpu. In addition it does not replace ` libvdpau-va-gl` which implements vdpau on top of `libva` and `libGL`.

Mic92: So this means that there are intel CPUs without GPU that have opencl support?

Kiskae: Technically yes[^1], but it appears to be part of an SDK and is supposed to be linked statically into applications.

It seems that `intel-ocl` is GPU-specific and could probably be added to the existing intel module, but this PR does not change the behavior of existing modules so that should probably be a separate PR.

[^1]: https://www.intel.com/content/www/us/en/developer/articles/tool/opencl-drivers.html#cpu-section

Mic92: fair enough.

Mic92:
bors merge

PR #573, opened 2023-03-04
common/gpu/intel/opencl: add opencl

Kiskae:
###### Description of changes

Updates intel libraries as per https://github.com/NixOS/nixpkgs/pull/205648
Adds module for intel opencl currently only supporting GPU

It is possible to add CPU OpenCL support, but only using an older version of the driver: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=intel-opencl-runtime

###### Things done

### DIFF
--- a/common/gpu/intel/opencl.nix
+++ b/common/gpu/intel/opencl.nix
@@ -1,0 +1,14 @@
+{ config, lib, pkgs, ... }:
+
+{
+  options.hardware.intel.opencl.enable = lib.mkEnableOption (lib.mkDoc
+    "intel opencl runtime (Install icd for Intel GPUs)"
+  );
+
+  config = lib.mkIf config.hardware.intel.opencl.enable {
+    hardware.opengl.extraPackages = with pkgs; [ 
+      intel-compute-runtime
+      intel-ocl
+    ];
+  };
+}
--- a/common/gpu/intel/default.nix
+++ b/common/gpu/intel/default.nix
@@ -1,5 +1,10 @@
 { config, lib, pkgs, ... }:
-
+let 
+  selectPackages = pkgs: with pkgs; [
+    intel-media-driver
+    vaapiIntel
+  ];
+in
 {
   boot.initrd.kernelModules = [ "i915" ];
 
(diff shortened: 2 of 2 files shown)


Kiskae:
The `intel-ocl` package is unfree and not required for newer processors, should it be gated behind an unfree/legacy option?

Mic92:
> The `intel-ocl` package is unfree and not required for newer processors, should it be gated behind an unfree/legacy option?

How old? `hardware.enableRedistributableFirmware` could be used potentially used if the license allows for redistribution.

common/gpu/intel/opencl.nix
```diff
@@ -0,0 +1,14 @@
+{ config, lib, pkgs, ... }:
+
+{
+  options.hardware.intel.opencl.enable = lib.mkEnableOption (lib.mkDoc
+    "intel opencl runtime (Install icd for Intel GPUs)"
+  );
```
Mic92, line 6: No strong opinion here...:

```suggestion
  options.hardware.intel.opencl.enable = lib.mkEnableOption (lib.mkDoc
    "intel opencl runtime (Install icd for Intel GPUs)"
  ) // {
    enable = true;
  };
```

Mic92:
Potentially makes sense to have the opencl option in nixos itself: https://github.com/NixOS/nixpkgs/pull/219657

common/gpu/intel/opencl.nix
```diff
@@ -0,0 +1,14 @@
+{ config, lib, pkgs, ... }:
+
+{
+  options.hardware.intel.opencl.enable = lib.mkEnableOption (lib.mkDoc
```
Mic92, line 4:
```suggestion
  options.hardware.intel.opencl.enable = lib.mkEnableOption (lib.mdDoc
```

Kiskae:
> > The `intel-ocl` package is unfree and not required for newer processors, should it be gated behind an unfree/legacy option?
> 
> How old? `hardware.enableRedistributableFirmware` could be used potentially used if the license allows for redistribution.

Both the `intel-media-driver` and `intel-compute-runtime` support starts with Broadwell, aka 5th gen intel and newer. The [licence](https://www.intel.com/content/dam/support/us/en/documents/server-products/intel-eula-(distribution-only).pdf) does not include any provisions for redistribution.

In addition I realized that `intel-ocl` is a very old version of the CPU OpenCL driver. It also includes a GPU driver but that isn't actually packaged correctly.

common/gpu/intel/opencl.nix
Pandapip1: Also, the module name is `hardware.intelgpu`, not `hardware.intel`.

Issue #552, opened 2023-02-09
Surface Pro 5 Touch does not work with updated version of IPTS


phpcitizen:
Surface pro 5 touch support does not work with the updated version of iptsd 1.0.1

Works with [iptsd 0.5.1](https://github.com/linux-surface/iptsd/releases/tag/v0.5.1)

It updates automatically.

Mic92:
cc @mexisme

mexisme:
> Surface pro 5 touch support does not work with the updated version of iptsd 1.0.1
> 
> Works with [iptsd 0.5.1](https://github.com/linux-surface/iptsd/releases/tag/v0.5.1)
> 
> It updates automatically.
> 
> 
> **Edit:** 
> The error is DEVICE not found in the service.
> 
> I created another service:
> ```
> let
>   # For iptsd 0.5.1
>   pkg-0_5_1 = import
>     (builtins.fetchTarball {
>       url = "https://github.com/NixOS/nixpkgs/archive/ee01de29d2f58d56b1be4ae24c24bd91c5380cea.tar.gz";
>       # Hash obtained using `nix-prefetch-url --unpack <url>`
> 
>       sha256 = "0829fqp43cp2ck56jympn5kk8ssjsyy993nsp0fjrnhi265hqps7";
>     })
>     {
>       system = "x86_64-linux";
>     };
> in
> {
>   systemd.services.iptsd2 = {
>     description = "IPTSD";
>     path = [
>       pkg-0_5_1.iptsd
>     ];
>     script = "iptsd";
>     wantedBy = [ "multi-user.target" ];
>   };
> ```
> and it works.
> 
>  I also realized the microsoft-surface.ipts.enable = false; option does not work as it seems to be config is already hard-coded to true. I don't know the case for microsoft-surface.surface-control.enable.
> 

Hardcoded is def. unhelpful.
I'll try to take a look early this coming week (juat flying, atm...)/

phpcitizen:
Forgot to mention that the ipts version issue happens in unstable channel.

dotlambda:
It's sufficient to add `iptsd` to `services.udev.packages` and `systemd.packages`.

phpcitizen:
Will not unstable branch still will fetch the version of iptsd 1.0.1?

dotlambda:
> Will not unstable branch still will fetch the version of iptsd 1.0.1?

Yes, and that version (and all later ones) ship with the appropriate udev rule and systemd version.
[...]

PR #584, opened 2023-03-19
Only create a systemd service config for IPTSd earlier than 1.0

mexisme:
###### Description of changes

IPTSd versions before 1.0 didn't include systemd or udev configs, but later ones do.
This PR attempts to only install a systemd service config for the earlier versions, and instead use the configs provided by `iptsd` package.

See #552 

###### Things done

### DIFF
--- a/microsoft/surface/common/ipts/default.nix
+++ b/microsoft/surface/common/ipts/default.nix
@@ -1,7 +1,7 @@
 { config, lib, pkgs, ... }:
 
 let
-  inherit (lib) mkDefault mkEnableOption mkIf mkMerge;
+  inherit (lib) mkDefault mkEnableOption mkIf mkMerge versionOlder;
 
   cfg = config.microsoft-surface.ipts;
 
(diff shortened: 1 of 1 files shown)


microsoft/surface/common/ipts/default.nix
```diff
@@ -15,13 +18,33 @@ in {
...
+
+      # TODO:
+      # "system.packages" only seems to check for ${package}/lib/systemd/systemd-${type} which means
+      # it won't find the iptsd ones, anyway:
+      # - https://github.com/NixOS/nixpkgs/blob/10e51cdc0456f1d5c8a00f026c384f0e81126538/nixos/modules/system/boot/systemd.nix#L467-L473
+      systemd.packages = [ iptsd ];
+
+      # TODO:
+      # udev.packages does add files from ${package}/etc/udev/rules.d/ but maybe just adding to
+      # environment.systemPackages is good enough?
```
dotlambda, line 45: Only packages from `services.udev.packages ` are added: https://github.com/NixOS/nixpkgs/blob/10e51cdc0456f1d5c8a00f026c384f0e81126538/nixos/modules/services/hardware/udev.nix#L417

fdgefgd:
any progress on this?

Mic92:
Closing this, since progress seems stalled. Just re-open if you make progress on this.

PR #586, opened 2023-03-23
treewide: unset hardware.video.hidpi

K900:
###### Description of changes

Seems like we can't agree on what the "good" settings are.

For older NixOS versions this is largely a no-op anyway.

###### Things done

### DIFF
--- a/common/hidpi.nix
+++ b/common/hidpi.nix
@@ -1,0 +1,12 @@
+{ lib, pkgs, ... }:
+let
+  # This option is removed from NixOS 23.05 and up
+  nixosVersion = lib.versions.majorMinor lib.version;
+  config = if lib.versionOlder nixosVersion "23.05" then {
+    hardware.video.hidpi.enable = lib.mkDefault true;
+  } else {
+    # Just set the console font, don't mess with the font settings
+    console.font = lib.mkDefault "${pkgs.terminus_font}/share/consolefonts/ter-v32n.psf.gz";
+    console.earlySetup = lib.mkDefault true;
+  };
+in config
--- a/gpd/p2-max/default.nix
+++ b/gpd/p2-max/default.nix
@@ -1,11 +1,9 @@
-{lib, ...}: {
+{
   imports = [
     ../../common/pc/laptop
     ../../common/pc/laptop/ssd
     ../../common/cpu/intel
     ../../common/cpu/intel/kaby-lake
+    ../../common/hidpi.nix
   ];
-
-  # HiDPI settings
-  hardware.video.hidpi.enable = lib.mkDefault true;
 }
--- a/flake.nix
+++ b/flake.nix
@@ -180,6 +180,7 @@
       common-gpu-nvidia = import ./common/gpu/nvidia/prime.nix;
       common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
       common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
+      common-hidpi = import ./common/hidpi.nix;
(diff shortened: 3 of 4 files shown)


RaitoBezarius (approved):

gpd/p2-max/default.nix
```diff
@@ -1,11 +1,8 @@
...
+{
   imports = [
     ../../common/pc/laptop
     ../../common/pc/laptop/ssd
     ../../common/cpu/intel
     ../../common/cpu/intel/kaby-lake
   ];
-
-  # HiDPI settings
-  hardware.video.hidpi.enable = lib.mkDefault true;
```
Mic92, line 10: We need to keep this option for 22.11 and older. Please add this option back for those.

Mic92: `lib.versionOlder (lib.versions.majorMinor lib.version) "22.11"`

K900:
bors merge

?

common/hidpi.nix
```diff
@@ -0,0 +1,12 @@
+{ lib, pkgs, ... }:
+let
+  # This option is removed from NixOS 23.05 and up
+  nixosVersion = lib.versions.majorMinor lib.version;
+  config = if lib.versionOlder nixosVersion "23.05" then {
+    hardware.video.hidpi.enable = lib.mkDefault true;
+  } else {
+    # Just set the console font, don't mess with the font settings
+    console.font = lib.mkDefault "${pkgs.terminus_font}/share/consolefonts/ter-v32n.psf.gz";
```
Mic92, line 9: Do we still need this for the kernel that have a 32px builtin already?

K900: This module is currently only included on GPD handhelds that have small low resolution displays, and the kernel heuristic only uses the resolution, so it would pick a small font by default.

nbraud: @K900 If that was the intent, then it's likely incorrect on the GPD Pocket 3, which has a 2K builtin display (WUXGA resolution to be precise: 1920×1200)

What's the threshold the kernel's heuristic uses?

K900: It's basically calculating the screen area in characters, and picking the font that draws less characters than a certain threshold (around 200 wide on 16:9 screens IIRC).

nbraud: OK, so it's not about them being low-res but high-DPI ?

K900: It's about them being a "low" resolution, but high density, so the kernel will pick a smaller font than it probably should.

nbraud: That doesn't make sense to me (re: "low" resolution but high density, those two values are coupled anyhow)

It seems to me that the issue is the kernel trying to shove 200 characters on a screen whose *physical width* is simply insufficient to display those.  For instance, on the pocket computer I'm writing from, the screen is 172mm wide, which would make each character less than 1mm wide.

Since the kernel is limited by the smallest font available (say, something 6px wide) that's what it would pick (closest match to 200 chars per line)
... but if the builtin display had twice the resolution (and the same size) it would still pick the smallest font and each character would be half as wide (in terms of physical size) and even more unreadable.

*i.e.* any display that is small (in physical dimensions) but has sufficiently-high DPI will encounter the issue.

K900: > That doesn't make sense to me (re: "low" resolution but high density, those two values are coupled anyhow)

"Resolution" should probably be more accurately called "pixel count" here.

> i.e. any display that is small (in physical dimensions) but has sufficiently-high DPI will encounter the issue.

Yes, but those are somewhat rare.

PR #590, opened 2023-03-25
don't set bigger console fonts on newer nixos systems

Mic92:
###### Things done

### DIFF
--- a/common/hidpi.nix
+++ b/common/hidpi.nix
@@ -1,12 +1,5 @@
 { lib, pkgs, ... }:
-let
+lib.optionalAttrs (lib.versionOlder (lib.versions.majorMinor lib.version) "23.05") {
   # This option is removed from NixOS 23.05 and up
-  nixosVersion = lib.versions.majorMinor lib.version;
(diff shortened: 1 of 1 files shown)


common/hidpi.nix
```diff
@@ -1,12 +1,5 @@
 { lib, pkgs, ... }:
-let
+lib.optionalAttrs (lib.versionOlder (lib.versions.majorMinor lib.version) "23.05") {
   # This option is removed from NixOS 23.05 and up
-  nixosVersion = lib.versions.majorMinor lib.version;
-  config = if lib.versionOlder nixosVersion "23.05" then {
-    hardware.video.hidpi.enable = lib.mkDefault true;
-  } else {
-    # Just set the console font, don't mess with the font settings
-    console.font = lib.mkDefault "${pkgs.terminus_font}/share/consolefonts/ter-v32n.psf.gz";
```
Mic92, line 9: In my understanding this no longer required for current nixpkgs unstable?

K900: As I said on the other PR, the GPD handhelds probably still want it, but we can move the setting into their profiles instead.

PR #596, opened 2023-03-26
feat: add backlight support for rpi4

louib:
This commit adds support for controlling the backlight on the official touch display for the Raspberry Pi 4. Using this overlay, the brightness on the display can be controlled using the following snippet:
```
echo "55" > /sys/class/backlight/rpi_backlight/brightness
```

###### Things done

### DIFF
--- a/raspberry-pi/4/backlight.nix
+++ b/raspberry-pi/4/backlight.nix
@@ -1,0 +1,50 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.hardware.raspberry-pi."4".backlight;
+in
+{
+  options.hardware = {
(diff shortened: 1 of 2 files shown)


K900:
Is anyone testing these on mainline kernels? Because those don't have `rpi_backlight` _at all_.

louib:
@K900 good catch, I only tested with the RPi4 kernel. I'm not even sure the driver is required. I'll try to remove it and test again.

Mic92:
> Is anyone testing these on mainline kernels? Because those don't have `rpi_backlight` _at all_.

No, I don't have all this hardware for testing. That's why we ask contributor to test.

PR #598, opened 2023-03-26
Add settings for AMD Raphael iGPU

stffffn:
###### Description of changes

Adds settings that are necessary to be able to use the iGPU of AMD Raphael CPUs (Ryzen 7000-series).

Additional info:
- Ryzen 7000-series iGPU support came with 5.18 (afaik)
- Lowest, currently available LTS kernel with the needed support is 6.1, which works perfectly fine
- The latest LTS kernel is 6.2, which needs an additional kernel parameter to disable scatter/gather because it results in completely white or flashing screens (scatter/gather was introduced with 6.2)
- The current release candidates of 6.3 also need this parameter

Since I don't think it's sensible to use a specific version (in this case 6.1) I would rather just use the latest kernel and add the necessary parameter. Especially since 6.3 still needs it too.

###### Things done

### DIFF
--- a/common/cpu/amd/raphael/igpu.nix
+++ b/common/cpu/amd/raphael/igpu.nix
@@ -1,0 +1,15 @@
+{ lib, pkgs, ... }: 
+
+{
+  # Disables scatter/gather which was introduced with kernel version 6.2
+  # It produces completely white or flashing screens when enabled while using the iGPU of Ryzen 7000-series CPUs (Raphael)
+
+  imports = [ ../. ];
+
+  boot = lib.mkMerge [
+    (lib.mkIf (lib.versionOlder pkgs.linux.version "6.1") {
+      kernelPackages = pkgs.linuxPackages_latest;
+      kernelParams = lib.mkMerge [["amdgpu.sg_display=0"]];  
+    })
+  ];
+}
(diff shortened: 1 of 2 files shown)


Mic92:
Usually we also export those nixos module in the flake: https://github.com/NixOS/nixos-hardware/blob/f38f9a4c9b2b6f89a5778465e0afd166a8300680/flake.nix#L169

stffffn:
Oh, sorry, missed that. Added it now.

common/cpu/amd/raphael/igpu.nix
```diff
@@ -0,0 +1,15 @@
...
+{
+  # Disables scatter/gather which was introduced with kernel version 6.2
+  # It produces completely white or flashing screens when enabled while using the iGPU of Ryzen 7000-series CPUs (Raphael)
+
+  imports = [ ../. ];
+
+  boot = lib.mkMerge [
+    (lib.mkIf (lib.versionOlder pkgs.linux.version "6.1") {
+      kernelPackages = pkgs.linuxPackages_latest;
+      kernelParams = ["amdgpu.sg_display=0"];
```
Mic92, line 12: Is the kernelParams flag not required if the kernel is new enough?

stffffn: Yep. From a LTS perspective 5.15 is too old, 6.1 is new enough and doesn't need the flag, 6.2 introduced scatter/gather which makes this flag necessary, and 6.3 still needs the flag too.

Mic92:
bors merge

PR #601, opened 2023-03-30
Patch kernel for HP envy x360 ey0xxx speakers

yisraeldov:
###### Description of changes
The speakers on this laptop don't work out of the box, needs a kernel patch. 

###### Things done

### DIFF
--- a/hp/envy/x360/ey0xxx/default.nix
+++ b/hp/envy/x360/ey0xxx/default.nix
@@ -1,0 +1,12 @@
+{ config, lib, ... }:
+with lib;
+{
+  config = {
+      boot.kernelPatches = [
+        {
+          name = "hp-envy-x360-ey0xxx-speaker";
+          patch = ./speaker.patch;
+        }
+      ];
+  };
+}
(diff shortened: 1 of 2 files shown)


hp/envy/x360/ey0xxx/default.nix
```diff
@@ -0,0 +1,12 @@
+{ config, lib, ... }:
```
Mic92, line 1: Please also add this model to the flake.nix and the README table.

hp/envy/x360/ey0xxx/default.nix
```diff
@@ -0,0 +1,12 @@
...
+with lib;
+{
+  config = {
+      boot.kernelPatches = [
+        {
+          name = "hp-envy-x360-ey0xxx-speaker";
```
Mic92, line 7: Where do you got this patch from and what kernel version does it target?

yisraeldov:
Hello.

I no longer have access to the laptop.

The patch is from here
***@***.***/

I was using the kernel version from "Unstable" as far as I remember.

Mic92:
Okay. It would be better if we had at least a some user that can actually review future pull requests.

yisraeldov:
I agree, sadly I had to return the laptop to work, is there a good place to put this patch so that others with the same machine can find it?  Maybe make an issue that references this ?

Mic92:
Not quite sure. Hopefully they find this pull request?

PR #606, opened 2023-04-15
Raspberry Pi 3 model B+ definition

raboof:
Basically copied the configuration for 4 and updated `hardware.deviceTree.filter`, which was sufficient to boot to the console for me.


raboof:
Turns out the default `nixos-generate-config`-generated config is fine

PR #632, opened 2023-05-21
Framework 12th gen: add keys workaround

ggPeti:
###### Description of changes

Someone figured out a reliable workaround to make the brightness and airplane mode keys on the Framework 12th gen work. This is that change.

###### Things done

### DIFF
--- a/framework/12th-gen-intel/default.nix
+++ b/framework/12th-gen-intel/default.nix
@@ -17,9 +17,32 @@
     "i915.enable_psr=1"
   ];
 
-  # This enables the brightness keys to work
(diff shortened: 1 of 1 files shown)


Mic92:
That looks potentially a bit racy in case the device is not present yet. I think you should make it also waiting for the device. `requires = [ "xxx.device"];` in the service or by using udev rules.

ggPeti:
@Mic92 The device doesn't show up in `systemctl -t device --all`. I might try to find a way to do it but I'm not even sure it's necessary - the device probably gets mounted way before network.target.

Mic92:
bors merge

PR #639, opened 2023-05-31
lenovo-thinkpad-x1-6th-gen: swap throttled with thermald

dawidd6:
###### Description of changes

According to https://wiki.archlinux.org/title/Lenovo_ThinkPad_X1_Carbon_(Gen_6)#Using_thermald, using newer thermald + newer kernel (which we have since 22.11) should be enough to workaround the throttling issue. `thermald` comes from intel and it seems to be better maintained than `throttled`, so I propose we switch to that.

###### Things done


Mic92:
bors merge

PR #669, opened 2023-07-10
ga402: remove useless kernel flags

aacebedo:
###### Description of changes

I previously added those flags but it was an error. The `acpi.prefer_micosoft_dsm_guide` was handled correctly by the kernel when they fixed the fan behaviour and the `pcie_aspm.policy` shall be handled by scripts or tools like TLP.  

###### Things done

### DIFF
--- a/asus/zephyrus/ga402/default.nix
+++ b/asus/zephyrus/ga402/default.nix
@@ -20,6 +20,6 @@
   };
 
   boot = {
-    kernelParams = [ "pcie_aspm.policy=powersupersave" "acpi.prefer_microsoft_dsm_guid=1" ];
+    kernelParams = [ "pcie_aspm.policy=powersupersave" ];
   };
 }


aacebedo:
@K900

K900:
`prefer_microsoft_dsm_guid` is definitely unnecessary, forcing ASPM maybe we should leave because it's a setting that's known-good on the hardware so it's basically an unconditional improvement.

aacebedo:
Ok. I noticed that tlp was changing this parameter depending on the AC state so I thought it would be better to leave it by default.

K900:
You should absolutely not run TLP on these machines anyway.

Mic92:
> You should absolutely not run TLP on these machines anyway.

Maybe add this to the README along with a reason why?

Mic92:
> power-profiles-daemon

Since I am reviewing quite a few nixos hardware PRs, is there a rule of thumb which laptops work well with power-profiles-daemon?

K900:
Not really, you just have to check if the firmware exposes support for it.

aacebedo:
corrected the flags

PR #671, opened 2023-07-14
framework: switch to power-profiles-daemon

Mic92:
###### Description of changes

Apparently a good idea these days instead of using tlp: https://github.com/NixOS/nixos-hardware/pull/669#issuecomment-1632099013

I am mostly confident that all of the frameworks should be supported: https://gitlab.gnome.org/Infrastructure/Mirrors/lorry-mirrors/gitlab_freedesktop_org/hadess/power-profiles-daemon#operations-on-intel-based-machines

cc @K900 @lheckemann

###### Things done

### DIFF
--- a/framework/default.nix
+++ b/framework/default.nix
@@ -24,6 +24,9 @@
     options snd-hda-intel model=dell-headset-multi
   '';
 
+  # Less invasive than tlp and should work for frameworks
+  services.power-profiles-daemon.enable = lib.mkDefault true;
+
   # For fingerprint support
   services.fprintd.enable = lib.mkDefault true;
 


K900:
Enabling it will basically never be an issue - if the firmware doesn't support performance profiles, it will propagate that to userspace.

Mic92:
Only thing to consider though is that enabling it has a side effect: https://github.com/NixOS/nixos-hardware/blob/d4ea64f2063820120c05f6ba93ee02e6d4671d6b/common/pc/laptop/default.nix#L9

Mic92:
One question. When I am on battery it shows me this:

```
╭─[~]
╰─ % powerprofilesctl list
  performance:
    Driver:     intel_pstate
    Degraded:   no

* balanced:
    Driver:     intel_pstate

  power-saver:
    Driver:     intel_pstate
```

Is this expected or show it switch to power-saver?

K900:
> Only thing to consider though is that enabling it has a side effect:
> 
> https://github.com/NixOS/nixos-hardware/blob/d4ea64f2063820120c05f6ba93ee02e6d4671d6b/common/pc/laptop/default.nix#L9

Oof this is bad. Enabling TLP by default is very very very VERY questionable.

> Is this expected or show it switch to power-saver?

It's not supposed to switch profiles automatically. All it does is indicate your preferences to the firmware.

accelbread:
This seems contrary to the advice from the Framework battery life thread: https://community.frame.work/t/guide-linux-battery-life-tuning/6665

Mic92:
I actually ended up testing both now and it seems that tlp even without configuring a lot gives me 1h more battery. So I am leaving things as they are.

Mic92:
> > Only thing to consider though is that enabling it has a side effect:
> > https://github.com/NixOS/nixos-hardware/blob/d4ea64f2063820120c05f6ba93ee02e6d4671d6b/common/pc/laptop/default.nix#L9
> 
> Oof this is bad. Enabling TLP by default is very very very VERY questionable.

Now I want data to back this up. I just saw the opposite being true.

> 
> > Is this expected or show it switch to power-saver?
> 
> It's not supposed to switch profiles automatically. All it does is indicate your preferences to the firmware.

PR #675, opened 2023-07-25
framework: Clarify 13th Gen Intel Core support

JohnAZoidberg:
###### Description of changes

Clarify 13th Gen Intel Core support. The chipsets and mainboard are similar
enough that no separate configuration is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,8 +124,9 @@ See code for all available configurations.
 | [Dell XPS 17 9700, nvidia](dell/xps/17-9700/nvidia)                 | `<nixos-hardware/dell/xps/17-9700/nvidia>`         |
 | [Dell XPS 17 9710, intel only](dell/xps/17-9710/intel)              | `<nixos-hardware/dell/xps/17-9710/intel>`          |
 | [Dell XPS E7240](dell/e7240)                                        | `<nixos-hardware/dell/e7240>`                      |
-| [Framework](framework)                                              | `<nixos-hardware/framework>`                       |
+| [Framework 11th Gen Intel Core](framework)                          | `<nixos-hardware/framework>`                       |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -124,8 +124,9 @@ See code for all available configurations.
 | [Dell XPS 17 9700, nvidia](dell/xps/17-9700/nvidia)                 | `<nixos-hardware/dell/xps/17-9700/nvidia>`         |
 | [Dell XPS 17 9710, intel only](dell/xps/17-9710/intel)              | `<nixos-hardware/dell/xps/17-9710/intel>`          |
 | [Dell XPS E7240](dell/e7240)                                        | `<nixos-hardware/dell/e7240>`                      |
-| [Framework](framework)                                              | `<nixos-hardware/framework>`                       |
+| [Framework 11th Gen Intel Core](framework)                          | `<nixos-hardware/framework>`                       |
 | [Framework 12th Gen Intel Core](framework/12th-gen-intel)           | `<nixos-hardware/framework/12th-gen-intel>`        |
+| [Framework 13th Gen Intel Core](framework/12th-gen-intel)           | `<nixos-hardware/framework/12th-gen-intel>`        |
```
Mic92, line 129: I would rather make a new directory even if it imports the 12th gen one.
Otherwise we have no way in the future to change anything for the new generation and it also looks very confusing to import the 12th gen configuration for a 13th gen laptop.

JohnAZoidberg: Sure, make sense.

Mic92:
```suggestion
| [Framework 13th Gen Intel Core](framework/13th-gen-intel)           | `<nixos-hardware/framework/13th-gen-intel>`        |
```

Mic92: Please also add this to the flake.nix

Issue #678, opened 2023-07-27
linux-firmware guard fails probably due to new versioning


chaoflow:
https://github.com/NixOS/nixos-hardware/blob/ba9650b14e83b365fb9e731f7d7c803f22d2aecf/lenovo/thinkpad/p14s/amd/gen2/default.nix#L15C97-L15C111

Current linux-firmware version is "2023-07-24"

SuperSandro2000:
https://github.com/NixOS/nixpkgs/pull/245305

MangoIV:
is there a solution for this?

samueldr:
This would be intrinsically fixed the next time this package gets a version tag, but break again once it switches to another untagged release.

The conditional could check for `lib.hasPrefix "unstable" linux-firmware.version` and assume any untagged release is fresh enough.

Also, is this still required? Are there any supported NixOS releases matching `lib.versionOlder pkgs.linux-firmware.version "20230210"`? The simpler solution could be to just delete the check.

MangoIV:
The latter would also be my suggestion.

PR #680, opened 2023-07-29
p14s, t14s: fix linux firmware version check

dadada:
###### Things done

### DIFF
--- a/lenovo/thinkpad/p14s/amd/gen2/default.nix
+++ b/lenovo/thinkpad/p14s/amd/gen2/default.nix
@@ -10,10 +10,6 @@
   # acpi_backlight=none allows the backlight save/load systemd service to work on older kernel versions
   boot.kernelParams = [ "amdgpu.backlight=0" ] ++ lib.optional (lib.versionOlder config.boot.kernelPackages.kernel.version "6.1.6") "acpi_backlight=none";
 
-
-  # Wifi support
-  hardware.firmware = lib.mkIf (lib.versionOlder pkgs.linux-firmware.version "20230210") [ pkgs.rtw89-firmware ];
-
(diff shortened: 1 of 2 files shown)


samueldr:
@dadada I haven't really looked at the problem more than [in this comment](https://github.com/NixOS/nixos-hardware/issues/678#issuecomment-1656105938), but would it be okay to just remove the check?

In other words, is the earliest version available since 23.05 was released already passing the check? If so, we probably should drop it, as I don't think supporting EOL'd releases is a goal here.

lenovo/thinkpad/p14s/amd/gen2/default.nix
```diff
@@ -1,4 +1,8 @@
 { lib, pkgs, config, ... }:
+let
+  fwVersion = pkgs.linux-firmware.version;
+  oldLinuxFirmware = if (lib.hasPrefix "unstable" fwVersion) then (lib.versionOlder fwVersion "unstable-2023-02-10") else (lib.versionOlder fwVersion "20230210");
```
dadada, line 4: I'm not sure which unstable version to match for exactly, but matching just `unstable` would probably already be enough. I'd be against removing the check. Even if it is no-longer supported, some people will still depend on 21.11 form some reason, and I'd hate to break their builds so soon.

dadada:
> @dadada I haven't really looked at the problem more than [in this comment](https://github.com/NixOS/nixos-hardware/issues/678#issuecomment-1656105938), but would it be okay to just remove the check?
> 
> In other words, is the earliest version available since 23.05 was released already passing the check? If so, we probably should drop it, as I don't think supporting EOL'd releases is a goal here.

I suppose removing it is alright. If people use EOL NixOS, they can pin an old version of nixos-hardware.

PR #686, opened 2023-08-04
Adding Dell XPS 13 9333

ttamttam:
###### Things done

Note1: In the `default.nix` file, I [imported](https://github.com/ttamttam/nixos-hardware/blob/df6673231f19e091b3202b1815b3531150a49fa8/dell/xps/13-9333/default.nix#L5) 
`../../../common/pc/ssd` because my laptop has currently an SSD, but
I'm not sure if it was the case when I bought it or if I installed it later.

Note2: Not sure about this [line](https://github.com/ttamttam/nixos-hardware/blob/df6673231f19e091b3202b1815b3531150a49fa8/dell/xps/13-9333/default.nix#L14) (`services.fwupd.enable = true;`).


bagelwaffle:
Hi there, I’m not a reviewer but just like to say that the device isn’t found in the FWUPD [supported device list](https://fwupd.org/lvfs/search?value=Xps+13+9333). I think running `fwupdmgr upgrade` will give you a list of components that is supported.

PR #693, opened 2023-08-09
raspberry-pi/4: fix modesetting on 6.1 kernels

### LINKED ISSUE #631: [RPI4] modesetting overlay fails to build with latest kernels
I have `hardware.raspberry-pi."4".fkms-3d.enable = true;` and found that the deviceTree failed to build with recent `unstable` Nixpkgs. The error was the generic "FDT_..." one which wasn't very helpful.

This build failure happened both before and after #627, a recent PR from @gador which updated some of the device tree overlays for compatibility with the 6.x kernel.

`modesetting.nix` still looks reasonable w.r.t. the latest [upstream definition of the overlay](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/vc4-fkms-v3d-overlay.dts) (though I note there's also a [neighbouring version of this](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts) labelled `-pi4`, which looks quite different) so I tried changing the device tree filter in `raspberry-pi/4/default.nix` too, as was done in #627 (adding the `-4*` in the filter pattern). However, the resulting system wasn't bootable, for reasons I couldn't fathom. Debugging it all has been a bit beyond me, and I resorted to painstakingly restoring an older working boot configuration. Any pointers welcome!

yu-re-ka:
###### Description of changes

Fixes #631

This should work with both nixos-23.05 and nixos-unstable

###### Things done

### DIFF
--- a/raspberry-pi/4/modesetting.nix
+++ b/raspberry-pi/4/modesetting.nix
@@ -27,6 +27,13 @@ in
   };
 
   config = lib.mkIf cfg.enable {
+    # fixes a crash: https://github.com/raspberrypi/linux/issues/5568
+    # can be removed for >= nixos 23.11: https://github.com/NixOS/nixpkgs/pull/247826
+    boot.kernelParams = [ "kunit.enable=0" ];
+
+    # doesn't work for the CM module, so we exclude e.g. bcm2711-rpi-cm4.dts
+    hardware.deviceTree.filter = "bcm2711-rpi-4*.dtb";
+
     # Configure for modesetting in the device tree
     hardware.deviceTree = {
       overlays = [


raspberry-pi/4/modesetting.nix
```diff
@@ -27,6 +27,13 @@ in
   };
 
   config = lib.mkIf cfg.enable {
+    # fixes a crash: https://github.com/raspberrypi/linux/issues/5568
+    # can be removed for >= nixos 23.11: https://github.com/NixOS/nixpkgs/pull/247826
+    boot.kernelParams = [ "kunit.enable=0" ];
+
+    # doesn't work for the CM module, so we exclude e.g. bcm2711-rpi-cm4.dts
```
purcell, line 34: The glob below doesn't match this filename, and might exclude other modules, so the comment is a bit misleading.

yu-re-ka: what do you mean? it _excludes_ the said device tree, so that is exactly the point

purcell: The comment says we will exclude `bcm2711-rpi-cm4.dts`.

The glob says we will exclude `bcm2711-rpi-4*.dtb`.

The glob `bcm2711-rpi-4*.dtb` does not match the filename `bcm2711-rpi-cm4.dts`, so the comment is confusing at best.

purcell: Am I right in thinking that we don’t have a Sentry project for production?

purcell: Oh hang on, I was assuming the `.filter` _removes_ the specified files — does it instead specify which to include?

yu-re-ka: Yes, it is an include

purcell: Doh! Got it, thank you!

purcell (approved):
Confirmed on my hardware that this resolves the issue, thanks!

purcell:
Might an admin (@Mic92?) be willing to merge this please?

purcell:
Thanks @Mic92!

colonelpanic8:
@yu-re-ka @Mic92 @samueldr Does anyone actually understand WHY we fail to apply to cm4? I'm totally on board with merging this as a short term fix because normal 4b users shouldn't encounter errors when trying to use these settings, but I wonder whether this is breaking things for cm4 users.

Why is               compatible = "brcm,bcm2711";

not enough to make this apply appropriately, basically?

PR #696, opened 2023-08-09
Dell XPS 9560: Add info to README and adjust configuration

Lyndeno:
###### Description of changes

Added information about hardware functionality and solutions I have discovered:
- Firmware updates with `fwupd`
- System control with `libsmbios`
- TPM 2.0 functions not working (with solutions)
- Note on secure boot functionality
- Note on using the TPM to unlock root drive

And some configuration changes:
- Do not use `lib.mkDefault` for the GPU IDs. User should not need to change this and this will prevent accidentally changing it.
- Import the `common/cpu/intel/kaby-lake` configuration for intel and bumblebee configurations

###### Things done

### DIFF
--- a/dell/xps/15-9560/nvidia/default.nix
+++ b/dell/xps/15-9560/nvidia/default.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ ... }:
 
 {
   imports = [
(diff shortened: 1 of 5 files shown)


dell/xps/15-9560/nvidia/default.nix
```diff
@@ -14,9 +14,9 @@
   ##### disable intel, run nvidia only and as default
   hardware.nvidia.prime = {
     # Bus ID of the Intel GPU.
-    intelBusId = lib.mkDefault "PCI:0:2:0";
+    intelBusId = "PCI:0:2:0";
```
Mic92, line 17: What's the reasoning behind removing mkDefault?

Lyndeno: I do not believe that ID changes unless one is using a completely different laptop model. In that case, the user should be warned if they or another module tries to change the value without explicitly forcing it.

Lyndeno:
Is there anything else needed from my end for this PR to be merged?

PR #709, opened 2023-08-31
default value for `services.hardware.openrgb.motherboard`

lucasew:
###### Description of changes
With this PR if something references the common processor module so that OpenRGB works out of the box if enabled.

###### Things done


Mic92:
Since this is a bit of a niche application, what do you think about inverting the control and set it in the openrbg module instead? https://github.com/NixOS/nixpkgs/pull/253048

lucasew:
> Since this is a bit of a niche application, what do you think about inverting the control and set it in the openrbg module instead? [NixOS/nixpkgs#253048](https://github.com/NixOS/nixpkgs/pull/253048)

This idea is much better.

PR #717, opened 2023-09-10
framework laptop 11th gen: Improve sleep power efficiency

obj-obj:
###### Description of changes
Change `mem_sleep_default=deep` kernel parameter to `acpi_osi="!Windows 2020"` (fixes a regression in s2idle, making it more power efficient than deep sleep). It's also a good idea to get rid of deep sleep on the 11th gen anyway, as it doesn't function correctly and takes 10-15 seconds to wake back up.

###### Things done

I haven't tested it using my own fork, as I just started using NixOS yesterday, but I've been using this kernel parameter on my Arch Linux install for a while now.

### DIFF
--- a/framework/default.nix
+++ b/framework/default.nix
@@ -6,9 +6,8 @@
   ];
 
   boot.kernelParams = [
-    # For Power consumption
-    # https://kvark.github.io/linux/framework/2021/10/17/framework-nixos.html
-    "mem_sleep_default=deep"
+    # Fixes a regression in s2idle, making it more power efficient than deep sleep
+    "acpi_osi=\"!Windows 2020\""
     # For Power consumption
     # https://community.frame.work/t/linux-battery-life-tuning/6665/156
     "nvme.noacpi=1"


Mic92:
@Pacman99 maybe. Don't have the hardware but I can ask around for people to test.

NinjaTrappeur:
I'm testing that on my gen12 frawework. 

After applying this PR, the powertop/cpupower idle output looks pretty similar to what I had before. That being said, this only test the idle power draw, not the suspend one.

I don't really know how to test the suspend power efficiency. I'll daily drive this for a couple of days and see if I see a massive change.

obj-obj:
> I'm testing that on my gen12 frawework.
> 
> After applying this PR, the powertop/cpupower idle output looks pretty similar to what I had before. That being said, this only test the idle power draw, not the suspend one.
> 
> I don't really know how to test the suspend power efficiency. I'll daily drive this for a couple of days and see if I see a massive change.

This PR won't affect power draw. To test suspend power efficiency, I usually leave the laptop unplugged overnight (in suspend), writing down the battery % and time before and after. Then I do hours ÷ % lost to get % drain/hour

NinjaTrappeur:
:+1: Will do that for the next two nights (with and without this patch). I'll post the results here.

Pacman99:
Well regardless of hardware, rn i think your change affects both laptops profiles because it's in default.nix. but the 12th Gen also has the deep sleep kernel param which could conflict with s2idle. I don't fully know how all this works but I don't think those two should be set at the same time.

Mic92:
> Well regardless of hardware, rn i think your change affects both laptops profiles because it's in default.nix. but the 12th Gen also has the deep sleep kernel param which could conflict with s2idle. I don't fully know how all this works but I don't think those two should be set at the same time.

Currently the 12th Gen profile does not import the other file.

obj-obj:
> 👍 Will do that for the next two nights (with and without this patch). I'll post the results here.

This PR only affects 11th gen, so you'll have to enter the kernel parameters manually. You should test deep sleep one night and s2idle with these two kernel parameters the other night, to see which one is more efficient

NinjaTrappeur:
In the lack of a better test, here's what I got after suspending the laptop with and without the windows 2020 interface disabled:

```
Windows 2020 disabled: 21:30 -> 8h05: 100% -> 82%
Without patch : 22:20 -> 9:20: 98% -> 76% 
```

Looking at this dubious test, it seems like the 12th gen implements the Windows 2020 operating system interface just fine. Or at least, disabling it doesn't significantly improves the sleep power efficiency.

I don't think we need to apply this patch to the 12th gen profile.

obj-obj:
> In the lack of a better test, here's what I got after suspending the laptop with and without the windows 2020 interface disabled:
> 
> ```
> Windows 2020 disabled: 21:30 -> 8h05: 100% -> 82%
> Without patch : 22:20 -> 9:20: 98% -> 76% 
> ```
> 
> Looking at this dubious test, it seems like the 12th gen implements the Windows 2020 operating system interface just fine. Or at least, disabling it doesn't significantly improves the sleep power efficiency.
> 
> I don't think we need to apply this patch to the 12th gen profile.
> 
> Edit: forgot to note that I tested this with the unpatched HDMI adapter attached to the laptop.
> 
> Dongles connected during the test:
> 
>     * USB-C
> 
>     * USB-A
> 
>     * Ethernet gen 1
> 
>     * HDMI gen 1

I think it's just an 11th gen bug, as without Windows 2020 kernel parameter the power button and keyboard backlight stay on during s2idle

NinjaTrappeur:
On 12th gen, the power button turns off and pulsates, keyboard backlight turns off when in s2idle w/ ACPI windows 2020 interface enabled.

obj-obj:
> On 12th gen, the power button turns off and pulsates, keyboard backlight turns off when in s2idle w/ ACPI windows 2020 interface enabled.

Does it still do so without it enabled?

PR #731, opened 2023-09-18
build-profile: also accept nvidia license

Mic92:
There seems to be recursive dependencies between adding video drivers and checking if the unfree flag is set.

###### Things done


Mic92:
@mergify queue

PR #743, opened 2023-09-28
pine64-pinebook-pro: improvements and simplifications

tomfitzhenry:
###### Description of changes

Improvements to the Pine64 Pinebook Pro config:

* Support boot-from-NVMe
* Remove quirks that have been upstreamed.
* Remove ap6526-firmware. Upstream (linux-firmware) has WiFI drivers.

See commit messages for more details.

I have been running this config for >6 months.

###### Things done


samueldr:
Looks plausibly okay from just reading the diff. (I haven't been using mine for much other than Tow-Boot dev...)

So tentatively half-approved, because it sure *looks* like it's fine.

pine64/pinebook-pro/default.nix
```diff
@@ -8,6 +8,10 @@
 
   boot.kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
 
+  boot.kernelParams = [
+    "console=tty0"
```
Mic92, line 12: @samueldr what is your opinion on this?

Related: https://github.com/NixOS/nixos-hardware/pull/745#discussion_r1341922235

Mic92: I guess by default it will try some serial device instead?

Mic92: I have some understanding of the option in case you have questions.

samueldr: I think I misspoke in the previous comment in what I think the best option is for this.

A default might be right, and it would depend on the device type.

We can't rely on the "unconfigured default", since in DT-land Linux defaults it relies on `/chosen/stdout-path`, which sometimes doesn't make sense given a target:

https://github.com/torvalds/linux/blob/1c8b86a3799f7e5be903c3f49fcdaee29fd385b5/arch/arm64/boot/dts/rockchip/rk3399-pinebook-pro.dts#L28-L30

But setting `console=` without qualifying it is kinda rude as it prevents the user from "simply" choosing another one.

So the workaround (with our semantics) is to put it at the leftmost location. In other words:

```
boot.kernelParams = mkBefore [ "console=tty0" ];
```

This would be for devices we can ascertain that most users would expect a given default console choice.

Mic92:
@mergify queue

PR #745, opened 2023-09-29
pine64-rockpro64: init

tomfitzhenry:
###### Description of changes

Adds the https://wiki.pine64.org/wiki/ROCKPro64

###### Things done


pine64/rockpro64/default.nix
```diff
@@ -0,0 +1,14 @@
+{ pkgs, lib, ... }:
+{
+  boot.kernelParams = [
+    # Set 115200 baud, per https://tow-boot.org/getting-started.html
+    "console=ttyS2,115200n8"
+  ];
```
samueldr, line 6: Setting any `console=` parameter is a strong choice. The last (right-most) valid one will be used as stdout/stdin for the kernel, and thus the NixOS scripted stage-1.

This setting will make it so serial works, but using a display won't allow interacting with the OS during early boot. In fact, I think it might even prevent stage-2 messages from showing up? Not 100% sure how systemd acts with only one `console= ` param, assuming NixOS doesn't add its own here. (systemd will do console multiplexing to any added `console=` params.)

pine64/rockpro64/default.nix
```diff
@@ -0,0 +1,9 @@
+{ pkgs, lib, ... }:
```
Mic92, line 1: Please reference this in the flake.nix and in README.md so that it's tested and discoverable.

Mic92:
@mergify queue

PR #747, opened 2023-09-30
star64: set governor to schedutil

fgaz:
###### Description of changes

By default it's "performance", which is not ideal.

Not sure if this belongs to nixos-hardware though.

###### Things done


Mic92:
If it makes the system more stable, that should be fine.

fgaz:
The stability is unchanged, it just runs a bit colder

fgaz:
I'll merge because I don't see any disadvantage in setting it to schedutil. The kernel will use the maximum frequency for real time tasks anyway, and we can't really measure responsiveness in interactive tasks because we don't have video output

PR #765, opened 2023-10-25
Add Framework 13 Ryzen 7040 Series

kjhoerr:
~~**Note:** It may be prudent to reorganize the structure of these changes slightly since the [Framework Laptop 16 7040 Series](https://frame.work/products/laptop16-diy-amd-7040) is supposed to ship by the end of the year, and there will be new unique features to the FW16 with the expansion bay and input modules. Let me know if I should place these files in a different folder structure with that in mind.~~ See #773 

###### Description of changes

- Added new configuration for Framework Laptop 13 AMD Ryzen 7040 Series. This shares some common configuration with the other Framework laptops around core shared components, though since this is the first AMD based Framework laptop there are other common modules for import and other options that are slightly different. Namely:
  - Prefer PPD over TLP for compatibility
  - `nvme.noacpi=1` kernel param causes issues with standby (cannot log back in after suspending)
  - Do not include the fw-ectool package - the ectool will likely have to be packaged differently somehow (?) since it's using a different code branch of ChromeOS EC than the Intel based Framework laptops: https://community.frame.work/t/what-ec-is-used/38574/2
- Best kernel version to select is not locked in as of yet, but likely prefer kernel_latest due to amdgpu driver updates. Other recommended distros at launch are Fedora 39 Beta and Ubuntu 23.04 with some changes required.

###### Things done


framework/7040-amd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    ../../common/cpu/amd
+    ../../common/cpu/amd/pstate.nix
+    ../../common/gpu/amd
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+  ];
+
+  boot.kernelParams = [
+    # Fixes a regression in s2idle, making it more power efficient than deep sleep
+    "acpi_osi=\"!Windows 2020\""
```
superm1, line 12: Don't do this. There is no need and it will just confuse things.

framework/7040-amd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  ];
+
+  # Newer kernel is better for amdgpu driver updates
+  # Requires at least 5.16 for working wi-fi and bluetooth (RZ616, kmod mt7922):
+  # https://wireless.wiki.kernel.org/en/users/drivers/mediatek
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "5.16") (lib.mkDefault pkgs.linuxPackages_latest);
```
superm1, line 18: You want at least 6.1.

framework/7040-amd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    # https://community.frame.work/t/headphone-jack-intermittent-noise/5246/55
+    SUBSYSTEM=="pci", ATTR{vendor}=="0x8086", ATTR{device}=="0xa0e0", ATTR{power/control}="on"
+    # Ethernet expansion card support
+    ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="0bda", ATTR{idProduct}=="8156", ATTR{power/autosuspend}="20"
+  '';
+
+  # Mis-detected by nixos-generate-config
+  # https://github.com/NixOS/nixpkgs/issues/171093
+  # https://wiki.archlinux.org/title/Framework_Laptop#Changing_the_brightness_of_the_monitor_does_not_work
+  hardware.acpilight.enable = lib.mkDefault true;
```
superm1, line 45: This shouldn't be needed. Amdgpu controls brightness.

framework/7040-amd/README.md
```diff
@@ -0,0 +1,17 @@
...
+First put enable `fwupd`
+
+```nix
+services.fwupd.enable = true;
+```
+
+Then run
+
+```sh
+ $ sudo fwupdmgr update
```
superm1, line 14: With policy kit do you really need sudo with nix?

framework/7040-amd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  '';
+
+  # For fingerprint support
+  services.fprintd.enable = lib.mkDefault true;
+
+  # Custom udev rules
+  services.udev.extraRules = ''
+    # Fix headphone noise when on powersave
+    # https://community.frame.work/t/headphone-jack-intermittent-noise/5246/55
+    SUBSYSTEM=="pci", ATTR{vendor}=="0x8086", ATTR{device}=="0xa0e0", ATTR{power/control}="on"
```
superm1, line 37: Does this really do anything on AMD model? I wouldn't expect it to

kjhoerr: Looking at the base issue it was an Intel + TLP issue. I tried without and it seems to work fine. Good catch!

leona-ya:
Maybe something along the line of `framework/13-inch/amd-7040` would make sense as a future file structure?

framework/7040-amd/README.md
```diff
@@ -0,0 +1,17 @@
+# [Framework Laptop 13](https://frame.work/)
```
Mic92, line 1: 16 here?

README.md
```diff
@@ -132,6 +132,7 @@ See code for all available configurations.
 | [Framework 11th Gen Intel Core](framework)                          | `<nixos-hardware/framework>`                       |
 | [Framework 12th Gen Intel Core](framework/12th-gen-intel)           | `<nixos-hardware/framework/12th-gen-intel>`        |
 | [Framework 13th Gen Intel Core](framework/13th-gen-intel)           | `<nixos-hardware/framework/13th-gen-intel>`        |
+| [Framework AMD Ryzen 7040 Series](framework/7040-amd)               | `<nixos-hardware/framework/7040-amd>`              |
```
Mic92, line 135: I thought about this a bit and I think it's better if distingish between Framework Laptop 16 and Framework Laptop 13. Even it turns out to be identical its better to have different models in the exposed API. We can always internally point to the same configuration. Does this make sense? I made this mistake with the 11th gen, and now we have no way to separate it from the generic configuration.

leona-ya: I totally agree. Are there any possibilities to restructure the directory structure (I think there is not "the solution" so far)?
Maybe backwards comptability for some time and add a warning?

I think it would be confusing if only the 13 inch AMD model has a 13 inch path and the others only have the old paths.

nodu:
Keep the old paths/exposures and recreate all in the new structure? Then
the old ones are still available and then new users can use the newly
exposed paths?

README.md
Mic92: Do we need a 13-inch part. Is there an overlap in the model numbers?

alexshpilkin: There are no real model numbers as you can see, only a descriptive name for the generation. The 16-inch version currently preparing to ship (aka the only one so far) is officially known as “Framework Laptop *16* (AMD Ryzen 7040 Series)”, while one of the 13-inch ones listed here is “Framework Laptop *13* (AMD Ryzen 7040 Series)”. So unless we want to go with motherboard part numbers instead (and those seem too granular unfortunately), I’d say we do. I mean, the 16-inch does have 7x40HSs where the 13-inch has 7x40Us, but identifying them with that feels wonky.

kjhoerr: I totally agree with all of the above - I made changes in line with @leona-ya 's original suggestion to put it in a 13-inch folder and updated the linking modules to reflect that.

I took the liberty of creating an issue #773 to hopefully spin off the conversation, assuming the new folder is acceptable? I don't want to leave this PR in a murky state implementing both changes and prolong the review process if I can help it

framework/13-inch/7040-amd/default.nix
```diff
@@ -0,0 +1,36 @@
...
+  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "6.1") (lib.mkDefault pkgs.linuxPackages_latest);
+
+  # AMD has better battery life with PPD over TLP:
+  # https://community.frame.work/t/responded-amd-7040-sleep-states/38101/13
+  services.power-profiles-daemon.enable = lib.mkDefault true;
+
+  # Fix TRRS headphones missing a mic
+  # https://community.frame.work/t/headset-microphone-on-linux/12387/3
+  boot.extraModprobeConfig = ''
+    options snd-hda-intel model=dell-headset-multi
```
superm1, line 22: JFYI; this sounds like it should be a kernel patch.  After someone writes up a patch and it lands somewhere you can later remove it.

reedriley: Yeah, probably - it's also been in the config since https://github.com/NixOS/nixos-hardware/commit/34485f1807befc9b104e59e0716e7a5cd0967fcb:
https://github.com/NixOS/nixos-hardware/blob/master/framework/default.nix#L20

(And might be an indication that there would be a benefit to refactoring these configs to share some kind of common base?)

reedriley:
Is there anything blocking landing this?  I've been running NixOS with it on my 7840U from Framework successfully for about a week now.

framework/13-inch/7040-amd/default.nix
kjhoerr: Added a note about it being a temporary configuration - as for common modules for framework, I plan to make a separate PR for that

superm1 (approved):

Mic92:
@mergify queue

PR #780, opened 2023-11-14
raspberry-pi/4: add LED-disable overlay

Stunkymonkey:
###### Description of changes
because the `boot.loader.raspberryPi.firmwareConfig` will be deprecated, others might want to disable the LEDs. Found in:
https://gist.github.com/SFrijters/206d2c09656affb04284f076c75a1969

###### Things done

### DIFF
--- a/raspberry-pi/4/leds.nix
+++ b/raspberry-pi/4/leds.nix
@@ -2,25 +2,25 @@
 
 let
   cfg = config.hardware.raspberry-pi."4".leds;
+  mkDisableOption = name: lib.mkOption {
+    default = false;
+    example = true;
+    description = "Whether to disable ${name}.";
+    type = lib.types.bool;
+  };
 in
 {
   options.hardware = {
     raspberry-pi."4".leds = {
-      disable-eth = lib.mkEnableOption ''
-        disable ethernet LEDs.
-      '';
-      disable-act = lib.mkEnableOption ''
-        disable activity LED.
-      '';
-      disable-pwr = lib.mkEnableOption ''
-        disable power LED.
-      '';
+      eth.disable = mkDisableOption ''ethernet LEDs.'';
+      act.disable = mkDisableOption ''activity LED.'';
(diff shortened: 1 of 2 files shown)


Stunkymonkey:
I am not sure if these options are OK. Please make suggestions.

raspberry-pi/4/leds.nix
```diff
@@ -0,0 +1,116 @@
...
+  options.hardware = {
+    raspberry-pi."4".leds = {
+      disable-eth = lib.mkEnableOption ''
+        disable ethernet LEDs.
+      '';
+      disable-act = lib.mkEnableOption ''
+        disable activity LED.
+      '';
+      disable-pwr = lib.mkEnableOption ''
+        disable power LED.
```
Mic92, line 16: mkEnableOption will add `Whether to enable X` as a description:

You could potentially just invert it by defining:

```
  mkDisableOption = name: mkOption {
    default = false;
    example = true;
    description = "Whether to disable ${name}.";
    type = lib.types.bool;
  };
```

```suggestion
      disable-eth = mkDisableOption ''ethernet LEDs.'';
      disable-act = mkDisableOption ''activity LED.'';
      disable-pwr = mkDisableOption ''power LED.'';
```

Mic92:
Options look fine to me. Cannot test on actual hardware.

raspberry-pi/4/leds.nix
Stunkymonkey: should we use `eth.disable` instead of `disable-eth`? what do you think? Just got the idea. Feels more like nix to me.

Stunkymonkey: Applied your suggestions.

please resolve this thread, if you agree.

Stunkymonkey:
the changes in [89e6821](https://github.com/NixOS/nixos-hardware/pull/780/commits/89e68213709d78509994178dc3e56643cc2af629) are not yet tested, but the are worked.

PR #784, opened 2023-11-20
framework/13-inch/12th-gen-intel: ec crash fix

cab404:
###### Description of changes

Fixing EC crashes during sleep for FW13 12th gen Intel by blacklisting cros_ec_lpcs. Not sure about other generations. 

###### Things done

Tested it on my hardware for a month.

### DIFF
--- a/framework/13-inch/12th-gen-intel/default.nix
+++ b/framework/13-inch/12th-gen-intel/default.nix
@@ -13,9 +13,14 @@
     "i915.enable_psr=1"
   ];
 
-  # This enables the brightness and airplane mode keys to work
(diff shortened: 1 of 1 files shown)


cab404:
@ricci can you review?

ricci:
This is not a problem I've had personally, so I can't confirm whether it fixed the problem reported but:

I reviewed the link provided in the comment, and this is a workaround that's been suggested by the vendor's official Linux support folks; I also applied this patch to my own laptop, and it doesn't cause any obvious problems.

So: looks good to me :+1: !

Mic92:
@mergify queue

PR #787, opened 2023-11-24
Bump linux-surface to a6eafcad32dc789ae92f42636b11e9aae6e7c879.

iwanders:
###### Description of changes
(First contribution to this repo) 

Bump the [linux-surface](https://github.com/linux-surface/linux-surface) version to the most recently [tagged release](https://github.com/linux-surface/linux-surface/releases/tag/arch-6.6.1-1).

###### Things done

- [x] Tested the changes in your own NixOS Configuration
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

Confirmed system comes up and without any noticable regressions using [this NixOS](https://github.com/iwanders/nixos-surface/blob/67fe8d0b35c2b04b1cc17b84b7c1c506940456dd/flake.nix#L4) configuration on an Surface Pro 9.

PR #788, opened 2023-11-26
Linux surface give iptsd configurable options.

iwanders:
###### Description of changes
Allow writing the [iptsd.conf](https://github.com/linux-surface/iptsd/blob/v1.4.0/etc/iptsd.conf) file from the nixos configuration.

So this allows specifying properties like [this](https://github.com/iwanders/nixos-surface/blob/f746cb6e58f81c9357aea6b4946e95c4016e8129/configuration.nix#L56-L61);
```nix
  microsoft-surface.ipts.enable = true;
  microsoft-surface.ipts.config = {
    DFT = {
      PositionMinMag = 500;
      ButtonMinMag = 20000;
    };
  };
```

Which then results in:
```
[root@papyrus:/etc/nixos]# cat /etc/iptsd/iptsd.conf 
[DFT]
ButtonMinMag=20000
PositionMinMag=500
```

Which is then used by iptsd.

Also changed the rev from #787 to a tag.

###### Things done

Tested with [this config](https://github.com/iwanders/nixos-surface/blob/f746cb6e58f81c9357aea6b4946e95c4016e8129/configuration.nix#L56-L61).

### DIFF
--- a/microsoft/surface/common/ipts/default.nix
+++ b/microsoft/surface/common/ipts/default.nix
@@ -1,18 +1,39 @@
 { config, lib, pkgs, ... }:
 
 let
-  inherit (lib) mkDefault mkEnableOption mkIf mkMerge;
+  inherit (lib) mkDefault mkEnableOption mkIf mkMerge mkOption types;
 
   cfg = config.microsoft-surface.ipts;
 
-in {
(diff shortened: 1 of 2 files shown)


Mic92:
@mergify queue

PR #789, opened 2023-11-28
FW13 7040: workaround for SuspendThenHibernate bug

reedriley:
This fixes a bug that affects Framework AMD Ryzen laptops, also documented on the Arch wiki here: https://wiki.archlinux.org/title/Framework_Laptop_13#Suspend-then-hibernate_on_AMD_version

It looks like the patch has been merged to rtc-next, so should be present starting in Linux 6.8.


Mic92:
@mergify queue

PR #791, opened 2023-12-03
framework intel: Disable cros-usbpd-charger

accelbread:
###### Description of changes

Disables an unused module that otherwise outputs an error log on bootup.

See https://community.frame.work/t/responded-linux-kernel-errors-about-cros-usbpd-charger/26264

I've had this disabled on 11th gen, and when upgraded it to 13th gen, still had log without it. I have not tested 12th gen but I assume it is the same.

###### Things done


Mic92:
@mergify queue

PR #794, opened 2023-12-04
Add support for lenovo idapad s145 15api laptop

Wa1t5:
###### Description of changes
Blacklist ideapad-laptop kernel module because it keeps messes with rfkill devices 
Specify which model snd-hda-intel should use, because without it external microphones won't work

###### Things done

- [ X ] Tested the changes in your own NixOS Configuration
- [ X ] Tested the changes end-to-end by using your fork of `nixos-hardware` and


Mic92:
@mergify queue

PR #814, opened 2023-12-29
framework amd: apply headset mic fix on older kernels

tlvince:
###### Description of changes

A kernel fix landed in https://github.com/torvalds/linux/commit/33038efb64f7576bac635164021f5c984d4c755f, which is in >= v6.7-rc5 and CC’ed to stable, so part of 6.6.8 too.

###### Things done

### DIFF
--- a/framework/13-inch/common/default.nix
+++ b/framework/13-inch/common/default.nix
@@ -6,9 +6,7 @@
 
   # Fix TRRS headphones missing a mic
   # https://community.frame.work/t/headset-microphone-on-linux/12387/3
-  #
-  # This is temporary until a kernel patch is submitted
-  boot.extraModprobeConfig = ''
+  boot.extraModprobeConfig = lib.mkIf (lib.versionOlder pkgs.linux.version "6.6.8") ''
     options snd-hda-intel model=dell-headset-multi
   '';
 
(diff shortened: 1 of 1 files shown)


framework/13-inch/common/default.nix
```diff
@@ -6,9 +6,7 @@
 
   # Fix TRRS headphones missing a mic
   # https://community.frame.work/t/headset-microphone-on-linux/12387/3
-  #
-  # This is temporary until a kernel patch is submitted
-  boot.extraModprobeConfig = ''
+  boot.extraModprobeConfig = lib.mkIf (lib.versionOlder pkgs.linux.version "6.6.8") ''
```
Mic92, line 9: pkgs also needs to be passed in the module header of this file.

tlvince: Done, thanks

Mic92:
@mergify queue

leana8959:
Hello,

I was reading the option set by this PR, and I noticed that the `mkIf` is on `pkgs.linux.version` and not `config.boot.kernelPackages.kernel.version`.
Hypothetically if one were to set their `config.boot.kernelPackages` to a prior version, this will not catch it.

Concretely it wouldn't make a difference since 6.6 is the lowest supported linux kernel in nixpkgs, I'm just curious.

Thank you for your time.

PR #815, opened 2023-12-29
Update config for xps-13-9010

Septias:
Add support for fingerprint.

### DIFF
--- a/dell/xps/13-9310/default.nix
+++ b/dell/xps/13-9310/default.nix
@@ -17,7 +17,7 @@
   boot.blacklistedKernelModules = [ "psmouse" ];
 
   # enable finger print sensor.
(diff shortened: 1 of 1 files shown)


dell/xps/13-9310/default.nix
```diff
@@ -16,6 +16,12 @@
   # Without this we get errors in dmesg on boot and hangs when shutting down.
   boot.blacklistedKernelModules = [ "psmouse" ];
 
+  # enable finger print sensor.
+  # this has to be configured with `sudo fprintd-enroll septias`
```
Septias, line 20:
```suggestion
  # this has to be configured with `sudo fprintd-enroll <username>`.
```

Mic92:
@mergify queue

PR #826, opened 2024-01-05
microchip icicle-kit: Fixes to kernel

mikatammi:
###### Description of changes

This change was mostly done, because in my NixOS configuration, I want to be able to use `boot.kernelPatches` to enable kernel options and patch the kernel.

* Improve by changing from legacy extraConfig to the structuredExtraConfig style kerenl configuration.
* Remove few unneeded kernel modules from default configuration.
* Make kernelPatches, structuredExtraConfig and extraMeta overrideable.
* Change callPackage-style function to actually take the individual attributes from pkgs.
* Get rid of recursive syntax.

###### Things done


Mic92:
@mergify queue

PR #831, opened 2024-01-07
surface: linux 6.6.8 -> 6.6.10, remove linux 6.1, set default 6.6 for surface-go

stigtsp:
###### Description of changes

- surface: linux 6.6.8 -> 6.6.10
- surface: set default kernel to major version 6.6
- surface: remove linux 6.1

This PR changes the default kernel for surface-go from 6.1 to the default 6.6, and removes linux 6.1 to reduce maintenance overhead.

It should also be tested on surface-go devices before being merged.

###### Things done

- [X] Tested on Surface Laptop 4 AMD
- [ ] Tested on Surface Go

- [X] Tested the changes in your own NixOS Configuration
- [X] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

Issue #825, opened 2024-01-04
Surface Go 3 - unsupported kernel version - 6.1.68 required


roukydesbois:
Hello!

As I have been informed in [this issue](https://github.com/linux-surface/linux-surface/issues/1339), the reason why my rebuild of nixos doesn't work should be related to the version of the kernel that the patches are applied to.

I don't think I'm yet knowledgeable enough to suggest a pull request myself, so could you please:
- tell me if I'm right assuming that the current state of the surface-go hardware configuration is 6.1.62 as it looks like [here](https://github.com/NixOS/nixos-hardware/blob/8e34f33464d77bea2d5cf7dc1066647b1ad2b324/microsoft/surface/common/kernel/linux-6.1.x/default.nix#L11C22-L11C22)
- help me in modifying the hardware configuration so that the kernel that is used is 6.1.68 (it is the one supported by the linux-surface team)

Thanks!

Mic92:
You probably just need to change this string? https://github.com/NixOS/nixos-hardware/blob/7fd5585e07386e8cf1f335b3950e52dc1f7d4b5a/microsoft/surface/surface-go/default.nix#L20

Mic92:
Ah. Here we actually build this version: https://github.com/NixOS/nixos-hardware/blob/7fd5585e07386e8cf1f335b3950e52dc1f7d4b5a/microsoft/surface/common/kernel/linux-6.1.x/default.nix

Mic92:
maybe @stigtsp has time for a bump?

roukydesbois:
> You probably just need to change this string?
> 
> https://github.com/NixOS/nixos-hardware/blob/7fd5585e07386e8cf1f335b3950e52dc1f7d4b5a/microsoft/surface/surface-go/default.nix#L20

I had indeed tried, but the value needs to be from a list, and you answered too fast already 😄

Mic92:
In the best case you just change the hash to an empty string and bump the version:

https://github.com/NixOS/nixos-hardware/blob/7fd5585e07386e8cf1f335b3950e52dc1f7d4b5a/microsoft/surface/common/kernel/linux-6.1.x/default.nix#L24

Than nix will complain with the correct hash that you can put in this file.

stigtsp:
_

> maybe @stigtsp has time for a bump?

Sure, I'll bump the kernels later today

stigtsp:
[...]

PR #832, opened 2024-01-07
surface: linux 6.1.62 -> 6.1.71

Cryolitia:
###### Description of changes

Update surface linux kernel from 6.1.62 to 6.1.71.

---

It fixes a build fault below:

```log
cpupower> applying patch /nix/store/lf6nr283p1vwm6gi5930sqmkmj614197-source/patches/6.1/0010-surface-shutdown.patch
cpupower> patching file drivers/pci/pci-driver.c
cpupower> patching file drivers/pci/quirks.c
cpupower> Hunk #1 FAILED at 6070.
cpupower> 1 out of 1 hunk FAILED -- saving rejects to file drivers/pci/quirks.c.rej
cpupower> patching file include/linux/pci.h
error: builder for '/nix/store/9j0h3djqqjzpvhinqfz675yk8szrkz5d-cpupower-6.1.62.drv' failed with exit code 1
```

https://github.com/NixOS/nixos-hardware/commit/249a94e715a3d6cc0217b37b0172dea1edd4afd2 updated `linux-surface` from `arch-6.6.4-1` to `arch-6.6.6-1`, which also introduced the commit https://github.com/linux-surface/linux-surface/commit/944cabb9a01b3069d1f016d021f0b6b8aa78f69d for linux 6.1.

This commit has a change https://github.com/linux-surface/linux-surface/commit/944cabb9a01b3069d1f016d021f0b6b8aa78f69d#diff-87dba197d613f882797cd43bc87e0ca180ecefbc3c28bac379f1d1d9b4c1ef47 , which updated the patch files' base commit.

https://github.com/torvalds/linux/commit/c9260693aa0c1e029ed23693cfd4d7814eee6624 was introduced in linux 6.1 by https://github.com/torvalds/linux/commit/1c8f75ee92334d89f1ddada26d47f9caa955f1a4 , which is the newer `0010-surface-shutdown.patch` based on, but we didn't update linux 6.1 at the same time.

Finally, boom——

---

What I would like to suggest is that maybe we should update and test more carefully, or an unexpected ripple effect could cause a 6.6 update to break a 6.1 build.

###### Things done

### DIFF
--- a/microsoft/surface/common/kernel/linux-6.1.x/default.nix
+++ b/microsoft/surface/common/kernel/linux-6.1.x/default.nix
@@ -8,7 +8,7 @@ let
 
   cfg = config.microsoft-surface;
 
-  version = "6.1.62";
+  version = "6.1.71";
   majorVersion = "6.1";
   patchDir = repos.linux-surface + "/patches/${majorVersion}";
   kernelPatches = pkgs.callPackage ./patches.nix {
@@ -21,7 +21,7 @@ let
     extraMeta.branch = majorVersion;
     src = fetchurl {
       url = "mirror://kernel/linux/kernel/v6.x/linux-${version}.tar.xz";
-      sha256 = "sha256-uf1hb6zWvs/O74i5vnGNDxZiXKs/6B0ROEgCpwkehew=";
+      sha256 = "sha256-Lfd03VP5/9Tlfr+ATPWXcJKV32owT+Jh0lIgoTS38EE=";
     };
   };
 
--- a/microsoft/surface/surface-go/default.nix
+++ b/microsoft/surface/surface-go/default.nix
@@ -17,7 +17,7 @@ in {
     ../../../common/cpu/intel/kaby-lake
   ];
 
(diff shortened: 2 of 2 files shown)


Mic92:
we were about to replace this kernel version in https://github.com/NixOS/nixos-hardware/pull/831

Cryolitia:
> we were about to replace this kernel version in #831

I agree the replace, and am building https://github.com/NixOS/nixos-hardware/pull/831 on my surface go to test. And I have absolutely no objection if this PR is closed directly.

I just think a bug should be logged and fixed rather than swept under the rug.

Mic92:
merged https://github.com/NixOS/nixos-hardware/pull/831 instead

stigtsp:
Hi @Cryolitia thanks for testing on Surface Go

> I just think a fault should be logged and fixed rather than swept under the rug.

The problem was logged in https://github.com/NixOS/nixos-hardware/issues/825 and fixed by replacing 6.1 with 6.6 in https://github.com/NixOS/nixos-hardware/pull/831

I hope this works for you and that I didn't miss anything here?

Issue #795, opened 2023-12-07
Config build fails due to boot.initrd.extraFiles “permission denied”


afreakk:
This error:
```
       > mkdir: cannot create directory 'root/lib/firmware/edid': Permission denied
       For full logs, run 'nix log /nix/store/bbrs8l9hz8gac3y63pkrs8wywq41cr54-initrd-linux-6.6.4.drv
```
Now happens because of this line https://github.com/NixOS/nixos-hardware/blob/c3abafb01cd7045dba522af29b625bd1e170c2fb/lenovo/legion/16ach6h/edid/default.nix#L14

After maybe this commit https://github.com/NixOS/nixpkgs/commit/d9105c28c8ca979fc1e55ba2f30511cbc36efa5a

I couldnt see that this issue was posted here.
More context here:
https://discourse.nixos.org/t/config-build-fails-due-to-boot-initrd-extrafiles-permission-denied/35976

Issue #786, opened 2023-11-23
Support for Lenovo Legion 7 16achg6


bratfizyk:
I'm using Lenovo Legion 7 16achg6.
`nixos-hardware` repo already contains an entry for a very similar module 16ach6h [here](https://github.com/NixOS/nixos-hardware/blob/master/lenovo/legion/16ach6h/hybrid/default.nix)

I naively tried using it on my machine but after a reboot the X service failed to start.

After a quick investigation I realized that for 16ach6h AMD GPU is connected to PCI:6:0:0.

https://github.com/NixOS/nixos-hardware/blob/c3abafb01cd7045dba522af29b625bd1e170c2fb/lenovo/legion/16ach6h/hybrid/default.nix#L22

However, on my machine it's PCI:5:0:0.
Here's a link to my dot file:

https://github.com/bratfizyk/dotFiles/blob/6c1125614f0d3de3b747447f574c695820f610b2/legion/configuration.nix#L113

If I'd like to create a PR adding support for my machine, shall I copy all the files required for 16ach6h and change the PCI for AMD GPU, or shall I link the files for the existing configuration and only override the fields that differ?

4JX:
Should be pretty similar looking config wise, though you probably don't want the `16ach6h`'s edid part, as that's a workaround to get higher refresh rates working. And there's also some other bits that might not affect your device, such as https://github.com/NixOS/nixos-hardware/pull/550/files

bratfizyk:
OK, thanks. I'll experiment with it and will let you know.
I guess I can create a branch in this repo and then refer to it using e.g. a flake like this, right?

```
inputs.nixos-hardware.url = "github:NixOS/nixos-hardware/myNewBranch";
```

4JX:
I don't think you'd have repo perms to do so, would be better to just point to your own fork in the meantime. I'd also advise to first try with the simplest of modules, ie just "common" `imports` and `hardware.nvidia.prime` before diving into having separate dedicated and hybrid setups.

bratfizyk:
Thanks, doing this now.

bratfizyk:
Ok, here it is:
https://github.com/bratfizyk/nixos-hardware/blob/master/lenovo/legion/16achg6/hybrid/default.nix

I copied `hybrid` config from the existing model and it almost worked. I had to comment out `nvidia-prime` common module, as I was unable to start X with it. Do you have an idea how can I check the exact error message my machien complained at?

[...]

PR #841, opened 2024-01-19
16ach6h: Various fixes

### LINKED ISSUE #628: Lenovo Legion 5 Pro 16ach6h breaks when using modesetting with current config
Opening this up to see how to go about fixing it.

The commit https://github.com/NixOS/nixos-hardware/commit/630a8e3e4eea61d35524699f2f59dbc64886357d switched the default AMD driver from `amdgpu` to `modesetting`. When updating past this commit Xorg does not boot up correctly on my machine in hybrid/offload mode: https://gist.github.com/4JX/37d35bf95970f96ad341658da7af0ecd, with the Xorg config that is created being https://gist.github.com/4JX/ec3b37198fdc5900ac258ca99ce0e068.

If I manually remove the layout/device/screen referencing `amdgpu` (Created due to the presence of `hardware.nvidia.prime.amdgpuBusId`. Where any duplicate entries for drivers, from both `amdgpu` or `modesetting`, that will eventually end up controlling the iGPU trigger this issue if at least one of them uses the `modesetting` driver.) and move the `BusID` to the modesetting section (not necessary, there for redundancy) it boots up just fine.

I'm not sure if a proper fix would be to just tell it to use the `amdgpu` driver for my device, as it seems that the NixOS option is also to blame (it only considers modesetting for [intel cards](https://github.com/NixOS/nixpkgs/blob/0b530a72bd17b3cb83923f3d3e1165726902a78b/nixos/modules/hardware/video/nvidia.nix#L247) despite the move towards ditching the other drivers https://github.com/NixOS/nixpkgs/pull/218437).

4JX:
###### Description of changes

A few commits to fix various annoyances I had with the config.

Fixes #628 and #795

Before merging I'd like to discuss the following the various people using this config the fact that the currently selected `amdgpuBusId = "PCI:06:00:0"` will only work for people who have two drives installed, while those that have one need to select `PCI:05:00:0` (see also https://github.com/NixOS/nixos-hardware/issues/786#issuecomment-1890439075 for a similar case, this was also brought up to my attention a long time ago on discord with the same issue)

If I had to choose a "sane default" that would be to do `lib.mkDefault "PCI:05:00:0"` with an accompanying note to set this manually for 2 drive users, but I wanted to hear your thoughts about it.

^ cc @LostAttractor @afreakk @piousdeer 

###### Things done

### DIFF
--- a/lenovo/legion/16ach6h/README.md
+++ b/lenovo/legion/16ach6h/README.md
@@ -3,6 +3,12 @@ Due to the introduction of DDG feature, you may toggle DDG frequently, so for th
 **But It will slow down NixOS evaluation by factor 2 and increase memory usage.**  
 So if you don't need specialization feature, you can just use hybrid only configuration or nvidia only (DDG only) configuration
 
+## Using multiple drives with this configuration
+
+When using more than one drive, the value of `hardware.nvidia.prime.amdgpuBusId` will change from the default of `PCI:5:0:0`.
+
+Make sure you override this default in your personal configuration. For two drives, it should be `PCI:6:0:0`.
+
 ## Setup at the time of testing
 ```
 $ nix-info -m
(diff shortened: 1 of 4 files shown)


piousdeer:
Indeed, the correct default value for unmodified laptops is `PCI:05:00:0`.

Mic92:
I think adding a hint in a README file for this model will be the best way forward.

4JX:
That should do it, works just fine with my my multiple drive laptop and a simple [config update](https://github.com/4JX/nixos-config/commit/2c7aebedc8c09bbc4aab77373da290ce492d4807).

Mic92:
@mergify queue

PR #852, opened 2024-02-02
Add specialisation for Lenovo Yoga Slim 7 Pro-X 14ARH7

mexisme:
###### Description of changes
Add specialisation for Lenovo Yoga Slim 7 Pro-X 14ARH7

###### Things done


mexisme:
At the moment, to support booting with only AMD-GPU _or_ NVidia support, two variations are provided.

I'm not convinced this is the best way, as opposed to (say) adjusting the `//common/gpu/amd/` and `//common/gpu/nvidia/` files to make them selectable via options rather than just from different imports.

However, the "two variations" way doesn't touch these files, that quite a few other hardware targets also rely on.

Mic92:
@mergify queue

PR #868, opened 2024-02-20
dell-xps-15-9570: init

Vinetos:
###### Description of changes
Adding configuration for Dell XPS 15 9570.

###### Things done

### DIFF
--- a/dell/xps/15-9570/README.wiki
+++ b/dell/xps/15-9570/README.wiki
@@ -1,0 +1,30 @@
+= Dell XPS 15 9570 =
+
+== Tested Hardware ==
+
+* CPU: Intel(R) Core(TM) i7-8750H CPU @ 4.10GHz
+* RAM: 32 GB
+* HDD: 512 GiB SSD
+* Screen: 15.6" FHD(1920x1080) non-touchscreen
+* Graphics: NVIDIA Corporation GTX1050 Ti Mobile 4GB GDDR5, with Intel Graphics too.
+* Input: trackpad, bluetooth mouse and keyboard
+
+
+== Firmware Configuration ==
+
+Not much tweaking of NixOS itself was needed. But we currently cannot automate the firmware setup, so this must be done by hand.
+
+== Installation ==
+
+=== Before installation ===
+
+These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thankfully doing so is as easy as changing any other simple setting.
+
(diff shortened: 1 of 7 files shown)


Mic92:
Can you have a look at the eval error?

dell/xps/15-9570/README.wiki
```diff
@@ -0,0 +1,30 @@
...
+
+Not much tweaking of NixOS itself was needed. But we currently cannot automate the firmware setup, so this must be done by hand.
+
+== Installation ==
+
+=== Before installation ===
+
+These settings are needed both for booting the final install, and installer itself. Therefore, they must be done first.
+
+* ''Disable Secure Boot (but keep UEFI Boot).'' Thakfully doing so is as easy as changing any other simple setting.
```
UrGuardian4ngel, line 23: > Thankfully* doing so

dell/xps/15-9570/default.nix
```diff
@@ -0,0 +1,12 @@
+{ lib, ... }:
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/cpu/intel/kaby-lake
```
UrGuardian4ngel, line 5: Any particular reason `kaby-lake` is imported here, but not in `intel/` or `nvidia/`?

flake.nix
```diff
@@ -64,6 +64,8 @@
       dell-xps-15-9560 = import ./dell/xps/15-9560;
       dell-xps-15-9560-intel = import ./dell/xps/15-9560/intel;
       dell-xps-15-9560-nvidia = import ./dell/xps/15-9560/nvidia;
+      dell-xps-15-9570-intel = import ./dell/xps/15-9570/intel;
+      dell-xps-15-9570-nvidia = import ./dell/xps/15-9570/nvidia;
```
UrGuardian4ngel, line 68: Do we also need to expose `./dell/xps/15-9570` as an option?

Mic92: I think having intel or nvidia should be fine. People have to make a choice between the two, no?

Vinetos: I agree with you. On these laptops, NVIDIA consumption is quite high. To prevent battery drain, I have added these options

dell/xps/15-9570/default.nix
Vinetos: Because here, `intel/` is the configuration for intel only and `nvidia/` means intel+nvidia
I could have imported kaby-lake in both files but as it is a common configuration, import it one time in the common file

Mic92:
@mergify queue

PR #884, opened 2024-03-06
Gigabyte b550 suspend fix

NireBryce:
I'm new to nixos and github collaboration in general so I'm not quite sure what I should fill out here. 

this fixes the b550 suspend bug by setting up systemd services that disable GPP0 and GPP8 in /proc/acpi/wakeup

###### Description of changes
- changed the README.md entry 
- created b550-fix-suspend.nix
  - which creates two systemd services that 

###### Things done

I also ran `./tests/run.py './gigabyte/b550/b550-fix-suspend.nix' ` as the contributing docs said to.  
Unfortunately, I cannot test this on boards outside of the B550m-D3SH which is my current motherboard; the README contains reports of others in the card family it also helps.


NireBryce:
I fixed a very embarassing typo, and then to make sure it sticks later if that ever happens in the future, added checks from another reply to the fix thread's checks for if it's enabled.

I'm sure there's more efficient ways to do this, but I'm not that deep into nix yet.

gigabyte/b550/b550-fix-suspend.nix
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+}
+
+
+
+
+
+
+
```
Mic92, line 42:
```suggestion
```

Mic92:
@mergify queue

PR #893, opened 2024-03-18
ASUS TUF FA507NV added.

samemrecebi:
###### Things done

### DIFF
--- a/asus/fa507nv/default.nix
+++ b/asus/fa507nv/default.nix
@@ -1,0 +1,21 @@
+{ lib, pkgs, ... }:
+
+{
+    imports = [
+      ../../common/cpu/amd
+      ../../common/cpu/amd/raphael/igpu.nix
+      ../../common/cpu/amd/pstate.nix
+      ../../common/gpu/nvidia
+      ../../common/gpu/nvidia/prime-sync.nix
+      ../../common/hidpi.nix
+      ../../common/pc/laptop
+      ../../common/pc/ssd
+    ];
+
+    boot.kernelPackages = pkgs.linuxPackages_latest;
+
+    hardware.nvidia.prime = {
+      amdgpuBusId = "PCI:54:0:0";
+      nvidiaBusId = "PCI:1:0:0";
+    };
+  }
--- a/common/gpu/nvidia/prime-sync.nix
+++ b/common/gpu/nvidia/prime-sync.nix
@@ -1,0 +1,10 @@
+{ lib, config, ... }:
+
+{
+  imports = [ ./. ];
+
+  hardware.nvidia.prime = {
+    # For people who want to use sync insted of offload. Esipecially for AMD CPU users
+    sync.enable = lib.mkOverride 990 true;
+  };
+}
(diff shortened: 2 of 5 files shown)


common/gpu/nvidia/prime-sync.nix
```diff
@@ -0,0 +1,10 @@
+{ lib, config, ... }:
+
+{
+  imports = [ ./. ];
+
+  hardware.nvidia.prime = {
+    # For people who want to use sync insted of offload. Esipecially for AMD CPU users
```
Mic92, line 7:
```suggestion
    # For people who want to use sync instead of offload. Especially for AMD CPU users
```

asus/fa507nv/default.nix
```diff
@@ -0,0 +1,21 @@
...
+      ../../common/cpu/amd/raphael/igpu.nix
+      ../../common/cpu/amd/pstate.nix
+      ../../common/gpu/nvidia
+      ../../common/gpu/nvidia/prime-sync.nix
+      ../../common/hidpi.nix
+      ../../common/pc/laptop
+      ../../common/pc/ssd
+    ];
+
+    boot.kernelPackages = pkgs.linuxPackages_latest;
```
Mic92, line 15:
```suggestion
    boot.kernelPackages =  lib.mkIf (lib.versionOlder pkgs.linux.version "6.8") pkgs.linuxPackages_latest;
```

Mic92: We usually don't set a kernel except if we depend on a new kernel to unlock features.
Please also clarify why we set the kernel to _latest here.

samemrecebi: Sorry for the late reply, Asus/TUF laptops patches where added after kernel 6.1 and for 2023+ model laptops for wireless and the keyboard to work kernel 6.6 was needed. As the default for NixOS is the LTS kernel these patches are not included. [https://asus-linux.org/guides/asusctl-install/](Here) is the guide I used for learning about the kernel issue. 

"pkgs.linuxPackages_latest" makes NixOS to use the latest stable release instead of the LTS so your suggestion here I believe is unnecessary. If necessary I can remove the kernel package line but I believe it is important for ideal use of this laptop model.

samemrecebi:
Made it more clear why I chose the latest kernel in the read me and fixed an issue related to kernel selection. Also synced the repo with the main. All inputs for changes and recommendations are appreciated.

asus/fa507nv/default.nix
Mic92: Your statement is true today but once we have linux 6.6 kernels by default, we no longer need to force users onto the latest kernel. That's why we instead have these conditional kernels in NixOS:

Using this option we only set kernelPackages if the current LTS kernel is too old.
```
boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "6.6") pkgs.linuxPackages_latest;
```

Mic92: Does this make sense?

samemrecebi: Ok I understand what you say, I will do the recomended change. Thanks for your input.

Mic92:
@mergify queue

PR #897, opened 2024-03-25
framework/12th-gen-intel: use s2idle

huantianad:
###### Description of changes

88348cb adds a fix that describes s2idle as being better than S3 deep sleep, thus we remove this option to take advantage of improved s2idle

###### Things done

### DIFF
--- a/framework/13-inch/12th-gen-intel/default.nix
+++ b/framework/13-inch/12th-gen-intel/default.nix
@@ -5,9 +5,6 @@
   ];
 
   boot.kernelParams = [
-    # For Power consumption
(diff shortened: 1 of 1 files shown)


SebTM:
Hey, refactoring the framework/12th config atm. and found your PR - do you still use it and do you have some logs/benchmarks that validates it's also true for the 12th/13th (as they use same config internal)?

huantianad:
I haven't had the time to test this unfortunately, not sure what the best way to test either. Still using it though, I honestly can't tell if it's better or worse haha

SebTM:
I did some tests and included the results/this change in my general-refactoring #903 - so I would go ahead and close this for now.

PR #909, opened 2024-04-12
apple/t2: update to kernel 6.7.7 and add tiny-dfr service

MarcusWagberg:
###### Description of changes
- Update the linux kernel with t2 patches to 6.7.7
- Removed fix-keyboard-backlight-and-touchbar.sh as this should no longer be needed.
- Added the package and service tiny-dfr as the new patches removed this functionality from a kernal module and placed it in userland. Kekrby's fork is used since the official one does not support t2 macs.

###### Things done

Any testing would be appreciated especially anyone that uses apple-set-os-loader as i do not use this and have not tested with it.


Mic92:
@mergify queue

PR #933, opened 2024-05-06
Lenovo P14s, T14 updates

tadfisher:
###### Description of changes

Add hardware configs for the following Lenovo models:

- ThinkPad P14s Gen 1 (AMD)
- ThinkPad P14s Gen 3 (AMD)
- ThinkPad P14s Gen 4 (AMD)
- ThinkPad T14 Gen 4 (AMD)

The only model I don't own and haven't tested is the T14 Gen 4 (AMD), which should actually be identical to the P14s Gen 4 except it can't be configured with more than 32GB of RAM.

The P14s base config differs slightly from the T14 base config, so align them.

In addition, enable the `amd_pstate` driver on P14s/T14 Gen 2 and above, which run Zen 3+ and thus are supported by the driver. This is needed for proper power profiles support.

###### Things done


Mic92:
@mergify queue

PR #941, opened 2024-05-12
New hardware config: Surface Pro 9

8bitbuddhist:
###### Description of changes

This is a fairly minor change to disable an Intel GPU feature called Panel Self Refresh (PSR). This option ostensibly increases battery life by letting the display refresh independently, but it causes a lot of display lag and stuttering, even on Windows. This PR proposes to disable this feature by default so folks have a usable out-of-the-box experience at the cost of a little battery life.

###### Things done

      importing it via `<nixos-hardware>` or Flake input: https://github.com/8bitbuddhist/nix-configuration/blob/main/flake.nix#L110


tracteurblinde:
@iwanders I recall you have an SP9. Do you have thoughts on disabling PSR by default?

iwanders:
Yeah, I have an SP9, I [also have this option disabled](https://github.com/iwanders/nixos-surface/blob/c5d9432fcf2ce9eadeef709d163e17c12be60935/configuration-base.nix#L16-L20), for me it mostly appeared as if the screen was flickering (Wayland, Gnome 45). So yeah, good change imo.

tracteurblinde (approved):

flake.nix
```diff
@@ -194,6 +194,7 @@
       microsoft-surface-laptop-amd = import ./microsoft/surface/surface-laptop-amd;
       microsoft-surface-common = import ./microsoft/surface/common;
       microsoft-surface-pro-3 = import ./microsoft/surface-pro/3;
+      microsoft-surface-pro-9 = import ./microsoft/surface-pro/9;
```
Mic92, line 197: Please also add this model to the table in the README. This is where the ci picks up model for evaluation.

Mic92:
@mergify queue

PR #944, opened 2024-05-14
asus/zenbook/ux371: init

fidgetingbits:
###### Description of changes

This laptop uses the tiger-lake chipset which didn't have a chipset folder, so I added it and added handling for loading the new Intel Iris Xe graphics driver over i915 if the kernel is new enough.

###### Things done

### DIFF
--- a/asus/zenbook/ux371/default.nix
+++ b/asus/zenbook/ux371/default.nix
@@ -1,0 +1,24 @@
+{
+  config,
+  lib,
+  pkgs,
+  inputs,
+  ...
+}:
+{
+  imports = [
+    ../../../common/cpu/intel/tiger-lake
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../battery.nix
+  ];
+
+  boot.kernelParams = lib.mkIf (config.hardware.intelgpu.driver == "xe") [
+    "i915.force_probe=!9a49"
+    "xe.force_probe=9a49"
+  ];
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+  services.thermald.enable = lib.mkDefault true;
+  powerManagement.cpuFreqGovernor = lib.mkDefault "powersave";
+}
--- a/common/cpu/intel/tiger-lake/default.nix
+++ b/common/cpu/intel/tiger-lake/default.nix
@@ -1,0 +1,12 @@
+{ config, lib, ... }:
+{
+  imports = [ ../. ];
+  config = lib.mkMerge [
+    (lib.mkIf (lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8") {
+      hardware.intelgpu.driver = "xe";
+    })
+    (lib.mkIf (config.hardware.intelgpu.driver == "i915") {
+      boot.kernelParams = [ "i915.enable_guc=3" ];
+    })
+  ];
+}
--- a/common/gpu/intel/default.nix
+++ b/common/gpu/intel/default.nix
@@ -1,32 +1,64 @@
-{ config, lib, pkgs, ... }:
-
 {
-  options.hardware.intelgpu.loadInInitrd = lib.mkEnableOption (lib.mdDoc
-    "loading `i195` kernelModule at stage 1. (Add `i915` to `boot.initrd.kernelModules`)"
-  ) // {
-    default = true;
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  options.hardware.intelgpu.driver = lib.mkOption {
+    description = "Intel GPU driver to use";
+    type = lib.types.enum [
+      "i915"
+      "xe"
+    ];
+    default = "i915";
   };
 
-  config = lib.mkMerge [
-    (lib.mkIf config.hardware.intelgpu.loadInInitrd { 
-      boot.initrd.kernelModules = [ "i915" ]; 
-    })
-    {
-      environment.variables = {
-        VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkDefault "va_gl");
-      };
+  options.hardware.intelgpu.loadInInitrd =
+    lib.mkEnableOption (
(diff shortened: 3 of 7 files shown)


asus/zenbook/ux371/default.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  boot.initrd.availableKernelModules = [
+    "xhci_pci"
+    "thunderbolt"
+    "vmd"
+    "nvme"
+    "usb_storage"
+    "sd_mod"
+  ];
+  boot.kernelModules = [ "kvm-intel" ];
```
Mic92, line 28: We are currently not duplicating code generated by nixos-generate-config:

```suggestion
```

common/gpu/intel/default.nix
```diff
@@ -1,23 +1,42 @@
...
+    // {
+      default = false;
+    };
 
   config = lib.mkMerge [
-    (lib.mkIf config.hardware.intelgpu.loadInInitrd { 
-      boot.initrd.kernelModules = [ "i915" ]; 
-    })
+    (lib.mkIf config.hardware.intelgpu.loadInInitrd { boot.initrd.kernelModules = [ "i915" ]; })
+    (lib.mkIf config.hardware.intelgpu.loadXeInInitrd { boot.initrd.kernelModules = [ "xe" ]; })
```
Mic92, line 27: so xe is for newer devices only, right?

common/gpu/intel/default.nix
```diff
@@ -1,23 +1,42 @@
...
-  };
+  options.hardware.intelgpu.loadInInitrd =
+    lib.mkEnableOption (
+      lib.mdDoc "loading `i195` kernelModule at stage 1. (Add `i915` to `boot.initrd.kernelModules`)"
+    )
+    // {
+      default = true;
+    };
+
+  options.hardware.intelgpu.loadXeInInitrd =
```
Mic92, line 17: Rather than duplicating the option above, I would rather add
`hardware.intelgpu.driver = "i915";` as an option and make new tiger lake + devices using it.

common/cpu/intel/tiger-lake/default.nix
```diff
@@ -0,0 +1,15 @@
...
+in
+{
+  imports = [ ../. ];
+  config = lib.mkMerge [
+    (lib.mkIf hasXeDriver {
+      # Favor xe driver over i915 if it exists
+      hardware.intelgpu.loadXeInInitrd = true;
+      hardware.intelgpu.loadInInitrd = false;
+    })
+    (lib.mkIf (!hasXeDriver) { boot.kernelParams = [ "i915.enable_guc=3" ]; })
```
Mic92, line 13: This I would move also to the gpu module and do a `hardware.intelgpu.driver == "xe"` comparison.

common/cpu/intel/tiger-lake/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  hasXeDriver = lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8";
+in
+{
+  imports = [ ../. ];
+  config = lib.mkMerge [
+    (lib.mkIf hasXeDriver {
+      # Favor xe driver over i915 if it exists
+      hardware.intelgpu.loadXeInInitrd = true;
+      hardware.intelgpu.loadInInitrd = false;
+    })
```
Mic92, line 12:
```suggestion
    {
      hardware.intelgpu.driver = "xe";
    }
```

common/gpu/intel/default.nix
fidgetingbits: ya it was introduced in the tiger lake chipset, so only it, alder lake, and raptor lake will have it afaik. I don't have an alder or raptor lake system to test so didn't bother adding those.

common/cpu/intel/tiger-lake/default.nix
fidgetingbits: I think we can't actually do that because "i915.enable_guc=3" is only applicable to tiger-lake, so it has to stay here. You can see other chipsets will have other `enable_guc` settings, and don't support the value 3. Also only 6.8 or newer kernels support the xe driver, but otherwise people will still use i915 on the tiger-lake. I think I've addressed this now though in the latest commit, but switching the tests around a bit and being more clear about the kernel version requirements.

fidgetingbits:
The assert I added fails, but I'm not exactly sure why yet. I see a similar kernel check in `asus/zephyrus/ga402/default.nix`. I'm guessing maybe the testing configuration uses an older kernel, but I'll have to investigate later.

common/gpu/intel/default.nix
```diff
@@ -1,26 +1,56 @@
...
+    ];
+
+    assertions = [
+      {
+        assertion = (
+          config.hardware.intelgpu.driver == "i915"
+          || (
+            config.hardware.intelgpu.driver == "xe"
+            && lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8"
+          )
```
Mic92, line 50: Minor simplification nit:

```suggestion
          config.hardware.intelgpu.driver != "xe"
          || lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8"
```

Mic92:
@mergify squash

flake.nix
```diff
@@ -23,6 +23,7 @@
       asus-pro-ws-x570-ace = import ./asus/pro-ws-x570-ace;
       asus-rog-strix-g513im = import ./asus/rog-strix/g513im;
       asus-rog-strix-g733qs = import ./asus/rog-strix/g733qs;
+      asus-zenbook-ux371 = import ./asus/zenbook/ux371;
```
Mic92, line 26: Ok. Now all that is missing is adding the model also to the markdown table in the README.md.

fidgetingbits: Added

fidgetingbits:
Fixed a merge conflict with [f0cf568](https://github.com/NixOS/nixos-hardware/commit/f0cf56878046c42ec2096a2ade89203e7348917b), but should be good now.

Mic92:
@mergify queue

Mic92:
@mergify rebase

Mic92:
@mergify queue

PR #948, opened 2024-05-16
Revision for Asus fa507nv

samemrecebi:
###### Description of changes
Some of the changes can be considered opinionated but I think all of them are needed to achieve full functionality for the device.

I added the battery module and changed the NVIDIA module to the normal NVIDIA offload instead of sync. After looking at the Arch wiki turns out AMD CPU's are also supported, unlike how its stated in the NixOS wiki. So I changed it to Offload. 

Secondly I copied 2 lines from the module of Framework 16-inch laptop. The first line fixes a issue related to AMD CPU's (Its details are included in the URL in the module) and the other one is setting up PPD for power management.

Additionally I enabled asusd service. This might be considered opinionated but without this tool functionalities like keyboard color control, fan curves etc. are not easily doable.

Finally I expanded the NVIDIA configuration to include some of the required and reccomended settings like modesetting.enable and nvidiaSettings.
###### Things done

### DIFF
--- a/asus/fa507nv/default.nix
+++ b/asus/fa507nv/default.nix
@@ -1,21 +1,45 @@
-{ lib, pkgs, ... }:
-
 {
-    imports = [
-      ../../common/cpu/amd
-      ../../common/cpu/amd/raphael/igpu.nix
-      ../../common/cpu/amd/pstate.nix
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

PR #951, opened 2024-05-19
Add config for ASUS TUF FX506HM

Andy3153:
###### Description of changes
I was already doing something like this in my configuration for a while and I decided it'd be best to publish it to this repo as well, so that other people can make use of it.

###### Things done

###### Others
Sorry for going off-topic but, I was wondering, do these configs get into the hardware scan NixOS does when generating a config, or is it up to the user to manually use these?


Mic92:
@mergify queue

PR #954, opened 2024-05-24
Add kernel version check for Lenovo 16ARHA7 speaker fix

8bitbuddhist:
###### Description of changes

This adds a check to prevent installing the speaker fix if the kernel is later than 6.9. Kernel 6.9-rc3 introduced the fix for the speaker issue, so this patch shouldn't be needed.

###### Things done


Mic92:
@mergify queue

PR #965, opened 2024-06-04
Move battery options to non-asus specific folder

fidgetingbits:
###### Description of changes

The current `asus/battery.nix` file isn't actually asus-specific, and seems useful to be used by other non-asus laptops. This moves it into a generic folder `common/pc/laptop/` and updates the current users.

Since there was also an option `hardware.asus.battery`, this deprecates the old option and changes to new `hardware.battery` instead.

I tested that the deprecation warning works if `hardware.asus.battery.chargeUpTo = 50;` is set somewhere. I also ran the test script against the three updated asus laptops to ensure they pass. 

I have not actually tested the changes on a live machine.

###### Things done


common/pc/laptop/battery.nix
```diff
@@ -19,11 +24,33 @@ in
...
+        "local-fs.target"
+        "suspend.target"
+      ];
+      after = [
+        "local-fs.target"
+        "suspend.target"
+      ];
       description = "Set the battery charge threshold to ${toString cfg.chargeUpto}%";
       startLimitBurst = 5;
       startLimitIntervalSec = 1;
```
Mic92, line 56: The script below currently only works with one battery, right?
I know that some thinkpads have a soldered battered and a changeable one.

Mic92: Are you sure this interface is actually portable? I see quite a few different approaches here: https://linrunner.de/tlp/settings/bc-vendors.html
There don't seem to be a standard.

Mic92: Ok. Looks like `charge_control_end_threshold` is the most common interface though.

fidgetingbits: Right ya not completely portable. I'm happy to update the script to test if that interface exists and then print a warning and exit if not. Then I (or others as needed over time) could add support for other battery interfaces. Can also put a note in the option about the supported interfaces. But then there is still the issue that inevitably some people will import/set it and not get their expected behavior I guess. 

Open to other suggestions, or if you don't think it's worth pursuing, feel free to close.

Mic92: Probably failing the service as it does now, is probably the best course of action.
However it would be cool if you could link to https://linrunner.de/tlp/settings/bc-vendors.html in the code so that people can figure out themselves if there battery is not supported.

Mic92: Also I would change it to iterate over all batteries instead of using just one.

fidgetingbits: > Also I would change it to iterate over all batteries instead of using just one.

The original command already does echo into a folder with `/BAT?/` so will update all of the batteries.

I did update the code to only declare the interface path once, mention that it applies to all batteries, and link the vendor page and also suggest tlp package if the script isn't applicable. 

I didn't get a chance to test the latest commit on a real asus laptop yet, but it does pass the test runs.

Mic92:
Let me know when you have tested this:

Currently we have an eval error for some laptops.

```
error: path '/home/runner/work/nixos-hardware/nixos-hardware/asus/battery.nix' does not exist
```

ambroisie:
@fidgetingbits there's a [module (and its relevant `settings` option for it](https://github.com/NixOS/nixpkgs/blob/57d6973abba7ea108bac64ae7629e7431e0199b6/nixos/modules/services/hardware/tlp.nix#L26-L35).

Mic92:
I was thinking the same. I don't think tlp is a particular big dependency either. Only thing is that it might conflict with power-profile-daemon.

PR #967, opened 2024-06-07
add Dell Latitude 9430

jheidbrink:
###### Description of changes

add Dell Latitude 9430: Set Kernel Parameter `i915.enable_psr=0` as described in https://discourse.nixos.org/t/i915-driver-has-bug-for-iris-xe-graphics/25006/12

Before this the graphics were lagging so much that the desktop (Sway) was unusable. Using the consoles acecssible with ctrl+alt+Fx worked, but also with a lot of delay. It also seemed to completely stop updating the screen after a while in Wayland.

With this change it works fine for me.

###### Things done


dell/latitude/9430/default.nix
```diff
@@ -0,0 +1,14 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
```
jheidbrink, line 7: I copied these imports from the 7430 config. There were some more settings in 7430 which I didn't copy over because I don't know what they do. To be honest I don't know what `intel`, `laptop` and `ssd` do neither, but they sound generic enough that they should be applied here ;)

Mic92: If you have a specific intel generation, you can also include it: https://github.com/NixOS/nixos-hardware/tree/master/common/cpu/intel

dell/latitude/9430/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  ];
+
+  boot.kernelParams = [
+    # See https://discourse.nixos.org/t/i915-driver-has-bug-for-iris-xe-graphics/25006/12
+    # jheidbrink reports that without this setting there is a very high lag in Sway which makes it unusable
+    "i915.enable_psr=0"
```
Mic92, line 13: Is this still necessary if we set this?

```
hardware.intelgpu.driver = "xe";
```

dell/latitude/9430/default.nix
Mic92: SSD mainly enables fstrim.

jheidbrink: The generation is Alder Lake which currently does not exist below `cpu/intel` - do you think it's best to create it, add the kernel param there and then import it here?

dell/latitude/9430/default.nix
jheidbrink: I tested with that config (and Kernel 6.8 as it is required for "xe") with `i915.enable_psr=0` commented out, and the lag appeared again. Setting `enable_psr=0` again made it disappear again.

Mic92:
@mergify queue

Issue #926, opened 2024-04-28
Gigabyte b550 as nixosModule to use in flake installation


gbrls:
Hi, I want to use the `gigabyte/b550` module from a flakes install. If the `flake.nix` file in this repo updates to include this module inside the `nixosModules` output, will it work or is there extra work needed?

I'm glad to help any way that I can, just new to nix and not sure how to do that.

matdibu:
Yeah, just open a PR and add it to this repo's `flake.nix`, then you can use it normally.

The fact that it's missing might have been an oversight, but I'm not sure.

Mic92:
Done https://github.com/NixOS/nixos-hardware/pull/978/files

PR #978, opened 2024-06-14
Fix issues with exisiting Gigabyte B550 patches

xunuwu:
###### Description of changes

This adds gigabyte-b550 to the flake outputs, it is currently missing (https://github.com/NixOS/nixos-hardware/issues/926)
This also fixes a quoting error in one of the systemd unit scripts that was causing it to fail

###### Things done


Mic92:
@mergify queue

ahydronous:
@xunuwu didn't know for sure where to post this, the Aorus B560m Aorus Pro is also affected by this. In this case the fix is `RP05`. Not sure how to get that added..

PR #984, opened 2024-06-16
System76 Gazelle Gaze 18

matteobongio:
###### Description of changes
trying to add support for the System76 Gazelle model Gaze18

###### Things done
added the start of a functioning nvidia configuration. Vulkan does not seem to be working.


system76/gaze18/default.nix
```diff
@@ -0,0 +1,40 @@
...
+      offload = {
+        enable = true;
+        enableOffloadCmd = lib.mkDefault true;
+      };
+    };
+  };
+}
+
+
+
```
Mic92, line 40:
```suggestion
}
```

Mic92, line 40:
```suggestion
}
```

flake.nix
```diff
@@ -250,6 +250,7 @@
       supermicro-x10sll-f = import ./supermicro/x10sll-f;
       supermicro-x12scz-tln4f = import ./supermicro/x12scz-tln4f;
       system76 = import ./system76;
+      system76-gaze18 = import ./system76/gaze18;
```
Mic92, line 253: Please also add the model to the README table.

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
+{ config, lib, pkgs, ... }:
+{
+  imports = [
+    ../.
```
Mic92, line 4:
```suggestion
    ../.
    ../../common/gpu/nvidia/prime.nix
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  ];
+  boot.initrd.kernelModules = [ "nvidia" ];
```
Mic92, line 6: Part of the nvidia module.

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+
+  hardware.opengl = {
+    enable = lib.mkDefault true;
+    driSupport = lib.mkDefault true;
+    driSupport32Bit = lib.mkDefault true;
+    extraPackages = with pkgs; [
+      vaapiVdpau
+    ];
```
Mic92, line 14: same.

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  services.xserver.videoDrivers = ["nvidia"];
```
Mic92, line 17:
```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+
+  hardware.nvidia = {
+
+    modesetting.enable = lib.mkDefault true;
+
+    powerManagement.finegrained = lib.mkDefault true;
+
+    nvidiaSettings = lib.mkDefault true;
+
+    package = config.boot.kernelPackages.nvidiaPackages.stable;
```
Mic92, line 27: I believe this is the default.

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+    prime = {
+      intelBusId = "PCI:0:2:0";
+      nvidiaBusId = "PCI:1:0:0";
+      offload = {
+        enable = true;
+        enableOffloadCmd = lib.mkDefault true;
+      };
```
Mic92, line 34: Part of the prime module:

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+
+  services.xserver.videoDrivers = ["nvidia"];
+
+  hardware.nvidia = {
+
+    modesetting.enable = lib.mkDefault true;
+
+    powerManagement.finegrained = lib.mkDefault true;
+
+    nvidiaSettings = lib.mkDefault true;
```
Mic92, line 25: This is the default.

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+    extraPackages = with pkgs; [
+      vaapiVdpau
+    ];
+  };
+
+  services.xserver.videoDrivers = ["nvidia"];
+
+  hardware.nvidia = {
+
+    modesetting.enable = lib.mkDefault true;
```
Mic92, line 21: From the documentation option:

> This is not enabled by default because it is not officially supported by NVIDIA and would not work with SLI

Can you add a comment why we are enabling it here?

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+{
+  imports = [
+    ../.
+  ];
+  boot.initrd.kernelModules = [ "nvidia" ];
+
+  hardware.opengl = {
+    enable = lib.mkDefault true;
+    driSupport = lib.mkDefault true;
+    driSupport32Bit = lib.mkDefault true;
```
Mic92, line 11: Enabled by default:

```suggestion
```

system76/gaze18/default.nix
```diff
@@ -0,0 +1,37 @@
...
+    ];
+  };
+
+  services.xserver.videoDrivers = ["nvidia"];
+
+  hardware.nvidia = {
+
+    modesetting.enable = lib.mkDefault true;
+
+    powerManagement.finegrained = lib.mkDefault true;
```
Mic92, line 23: Some laptops have issues with suspend and this. I assume this is not the case here?

system76/gaze18/default.nix
matteobongio: I am enabling it because the [NixOS Nvidia wiki](https://wiki.nixos.org/wiki/Nvidia) says to.
```
  hardware.nvidia = {

    # Modesetting is required.
    modesetting.enable = true;
```

If it's not required I can remove it.

system76/gaze18/default.nix
matteobongio: Nope, I haven't encountered any issues with suspend.

system76/gaze18/default.nix
Mic92: It further says:

```
Enabling this fixes screen tearing when using Optimus via PRIME (see hardware.nvidia.prime.sync.enable.
```

Did you notice this when enabling/disabling this option?

matteobongio: prime sync didn't work at all, when I enabled it and rebooted I couldn't get a graphical desktop, I had to open a tty and disable it.

I haven't noticed any screen tearing, but I also haven't really used the nvidia card, I assume KDE has been running on the integrated graphics, and I haven't done anything graphically intensive because I can't get vulkan to work, so I haven't really been able to put anything to the test, also I've had it enabled since the start because I copied and pasted it when adding nvidia drivers to my system configuration. I can try without it later today.

matteobongio: Just tried it and now the nvidia driver isn't being loaded, one of the changes from this morning probably caused this.

system76/gaze18/default.nix
matteobongio: When I remove this, the nvidia drivers do not get loaded.

system76/gaze18/default.nix
matteobongio: I'm confused, should I remove driSupport32Bit or leave it?

system76/gaze18/default.nix
matteobongio: removing modesetting had no visible impact so far.

system76/gaze18/default.nix
Mic92: I personally probably wouldn't add `driSupport32Bit` these days by default because it's only used for older 32-bit games, but I leave it to you if you want to keep it.

matteobongio: I don't t think that there are any downsides from having it enabled.

Mic92: It adds something like 100M of drivers including dependencies.

system76/gaze18/default.nix
Mic92: It probably gets loaded, but later after the system is mounted. For intel drivers we also add them to the initrd by default. I don't know why we don't do this for nvidia as well by default.

system76/gaze18/default.nix
```diff
@@ -0,0 +1,26 @@
...
+  imports = [
+    ../.
+    ../../common/gpu/nvidia/prime.nix
+  ];
+
+  boot.initrd.kernelModules = [ "nvidia" ];
+
+  hardware.opengl = {
+    enable = lib.mkDefault true;
+    driSupport32Bit = lib.mkDefault true;
```
Mic92, line 12:
```suggestion
    # adds ~100MB of 32-bit mesa drivers.
    driSupport32Bit = lib.mkDefault true;
```

Mic92 (approved):
Does look good on my side. Same for you? @matteobongio

matteobongio:
looks good to me, but I still haven't managed to make vulkan work.

Mic92:
@mergify queue

PR #985, opened 2024-06-17
refactor Microsoft directory

### LINKED ISSUE #936: iptsd failing on suspension
`iptsd` fails when suspended, but can be manually restarted just fine. reported this to main repo [linux-surface/iptsd#162](https://github.com/linux-surface/iptsd/issues/162), but they noted that it functions well without the alterations made by this flake. why, then, does the flake not use the implementation of `iptsd` that is in `nixpkgs` and instead its own?

Flameopathic:
###### Description of changes
I have made the structure of the `microsoft` directory standard and better organized. I also removed some things that were no longer necessary and some that were causing bugs.

- removed SP3 `default.nix`. the folder is now mostly useless, but I am not going to remove the guide that is in the folder that someone put the effort into creating. the SP3 module now just points to the same place as the intel module.
- moved all surface pro things into the same folder as the surface pro intel and removed "surface" from the start of some directories that were already in the `surface` directory and thus did not need this extra word.
- removed custom `iptsd` implementation and replaced with the one that is a part of standard NixOS. fixes #936
- removed custom `surface-control` implementation. kinda weird that it was there because it did nothing more than the standard implementation does.
- used `lib.mkDefault` for `iptsd`. it was not something that needed to be required.  

###### Things done

I have only tested on the SP7; however, I have taken care to do exactly what I did to the intel config to all other configs that need it.


README.md
```diff
@@ -270,12 +270,12 @@ See code for all available configurations.
 | [MSI GS60 2QE](msi/gs60)                                               | `<nixos-hardware/msi/gs60>`                             |
 | [MSI GL62/CX62](msi/gl62)                                              | `<nixos-hardware/msi/gl62>`                             |
 | [Microchip Icicle Kit](microchip/icicle-kit)                           | `<nixos-hardware/microchip/icicle-kit>`                 |
-| [Microsoft Surface Go](microsoft/surface/surface-go)                   | `<nixos-hardware/microsoft/surface/surface-go>`         |
```
Mic92, line 273: When we remove old models, we put deprecation warnings in place so that existing users now where there model went: https://github.com/NixOS/nixos-hardware/blob/master/microsoft/surface/default.nix

Mic92: There is also flake.nix that needs to be updated.

Flameopathic: @Mic92 gotchya, I believe it is fixed now. thank you!

Mic92: There are still references missing here: https://github.com/NixOS/nixos-hardware/blob/cde8f7e11f036160b0fd6a9e07dc4c8e4061cf06/flake.nix#L208
Also `/microsoft/surface/surface-go` needs warning.

Flameopathic: okay, so all changed directories need deprecation messages, not just module references?

Flameopathic: @Mic92 i only ask because i changed a good number of paths, but ensured that all modules in the flake point to the right place, including the surface go. i made just about the same changes to the surface go as i did the surface laptop amd, and the paths to everything in the pro series was changed. only model removed was sp3.

Mic92: Everything we had before in flake.nix and the README should still work or throw a warning. I personally would not do to many refactoring that just provide minor cosmetic improvements. We usually do these rename if we split up hardware profiles into multiple devices or if we unify things or if models where simply named incorrectly.

Flameopathic: > Everything we had before in flake.nix and the README should still work or throw a warning.

That is true right now, Surface Pro 3 configuration is the only one that should throw an error.

> I personally would not do to many refactoring that just provide minor cosmetic improvements. We usually do these rename if we split up hardware profiles into multiple devices or if we unify things or if models where simply named incorrectly.

understood. would it be better if i just close this?

Flameopathic: @Mic92 I'm willing to do whatever, i just wanna understand what you'd like.

Mic92: Sorry. I am busy with other things just now. If it can wait a bit, I will have a look later. Apart from that I invite @mexisme to make judgment calls.

Mic92: If you want to speed up the process, please separate fixes from refactorings. Than I can already merge fixes itself.

Flameopathic: don't worry about it, i'll close this and do the proper things separately. thanks for working through this with me, i apologize for my poor form as this is my first contribution. :pray:

Flameopathic:
separating fixes and refactoring

PR #995, opened 2024-06-20
gpu/amd: drop hardware.amdgpu.opencl option

Mic92:
This now conflicts with nixpkgs. See https://nixpk.gs/pr-tracker.html?pr=319865

You will need to update from `hardware.amdgpu.opencl = true;` to `hardware.amdgpu.opencl.enable = true;`

###### Description of changes

See https://github.com/NixOS/nixpkgs/pull/319865

The other options we will migrate too, once they have landed in stable/unstable channels

###### Things done


Mic92:
A few people seem to use this:
https://github.com/search?q=hardware.amdgpu.opencl+lang%3Anix&type=code

cc @blitz @necrophcodr @jocelynthode @felipecaninnovaes @jpaali @nightconcept @EnzioKam

Mic92:
You will need to update nixpkgs to the version that has this option or pin nixos-hardware to the old version.
@nevivurn @tupakkatapa @DrymarchonShaun @arvigeus @anna328p @Mange @wochap

Mic92:
cc @jzbor @CallumTarttelin @knightpp @spagnoloG

common/gpu/amd/default.nix
```diff
@@ -6,11 +6,6 @@
   ) // {
     default = true;
   };
-  options.hardware.amdgpu.opencl = lib.mkEnableOption (lib.mdDoc
-    "rocm opencl runtime (Install rocmPackages.clr and rocmPackages.clr.icd)"
-  ) // {
-    default = true;
```
Mic92, line 12: Once we have the option, we maybe should default this to true?

Mic92:
An alternative hotfix is to just import the packages yourself:

```
{
  hardware.opengl.extraPackages = with pkgs.rocmPackages; [ clr clr.icd ];
}
```

Mic92:
@mergify queue

common/gpu/amd/default.nix
JohnRTitor: For this module? Yeah I think it should use 
```nix
hardware.amdgpu.rocm.enable = lib.mkDefault true;
```

Mic92: Wasn't this called `hardware.amdgpu.opencl.enable = lib.mkDefault true;` now?

Mic92:
@mergify queue

common/gpu/amd/default.nix
JohnRTitor: yeah, sorry, that was a mistake on my part.

Mic92:
@mergify dequeue

Mic92:
@mergify queue

blitz:
So this means that nixos-hardware master will not be compatible with 24.05 wrt to amdgpu settings? If so would it make sense to just create a branch/tag for the last working version?

Mic92:
> So this means that nixos-hardware master will not be compatible with 24.05 wrt to amdgpu settings? If so would it make sense to just create a branch/tag for the last working version?

The module has been backported to 24.05, so branching won't change anything here.

PR #998, opened 2024-06-21
apple/t2: update to kernel 6.9.4

zachcoyle:
###### Description of changes

Update kernel to 6.9.4
Updates patches

It's still kind of flaky, but much more stable than 6.9.3 for me 

###### Things done


Mic92:
Their used to be some opponents in nixpkgs that didn't want to have all vendor kernels in nixpkgs... But maybe things have changed now - I believe samueldr was one of them. However as an alternative we could also add nixos-hardware to hydra to get kernel like this one and surface kernels build.

zachcoyle:
oh man that would save me so much time if that were possible

PR #999, opened 2024-06-21
Fix and test 24.05 evaluation

Mic92:
###### Things done


Mic92:
@mergify queue

Mic92:
It was also likely broken before this commit because nixos-hardware was using hardware.graphics.enable, which doesn't exist in 24.05.

maralorn:
Yeah, I had just bad luck that I bumped the dependency in the short time window where it wasn’t producing an eval error but a run time behavior change.

PR #1008, opened 2024-06-24
doc: Improve doc for Tuxedo Pulse 14 Gen3 power issues

gabyx:
###### Description of changes

Documentation for Tuxedo Pulse 14 Gen3

###### Things done


tuxedo/pulse/14/gen3/README.md
```diff
@@ -5,3 +5,22 @@
...
+With the Linux Kernel version `6.6.33` (NixOS 24.05) there are shutdown issues resulting in the battery not turning off
+completely. Apparently a newer Kernel (tested with `6.8.12`) fixes this (the exact version where this problem is fixed is unknown).
+
+You can use
+
+```nix
+  boot.kernelPackages =
+    if (config.boot.zfs.enabled)
+    then pkgs.zfs.latestCompatibleLinuxPackages
+    else pkgs.linuxPackages_latest;
```
Mic92, line 22: We can also set a newer kernel by default in this profile, if it fixes issues.

gabyx: Jeah, I was just unsure if its really good to do that. Because you are on stable and then the hardware configs just intrusively replace the linux kernel  :-). lol.

Not sure, what you think?

Mic92: The issues, that you are describing here, seems worth the upgrade.

gabyx: Jeah thats true, but is it safe to just add the fix?. Or would it be good to make an assertion if the kernel version is smaller then 6.8 etc and t let the user decide what to do?

gabyx: @britter This is the PR: How would you go about dissecting this?

Read the linux kernel changelof and search for something as apmd or power etc? 
Or just git bisect and test: Does anybody know how to cycle through the kernels in Nix, update Nixpkgs in flake always to another commit? or just look into linuxPackages_latest.kernel.version

May be we can just build a boot entry "test" with all the generations with incresing kernel version and the bisect through till we find one that does not work anymore...

britter: I find upgrading to the latest kernel unconditionally a little bit drastic. People on the release channel expect some kind of stability and might not want to roll with the latest and greatest kernel. Personally I'm planning to downgrade again once the nixpkgs release channel pics up the fixed kernel version. Probably when 24.11 is released.

I think the best thing would be a condition like `if kernel_version >= 6.1.91 %% kernel_version <= 6.?` and only then upgrade to the latest kernel. But that required looking more closely into the kernel changelog and they tend to be very long. I think looking through the changelog is faster then booting into individual kernel version. That said, I don't see myself having the time to go through the kernel changelog anytime soon...

@Mic92 I would merge this because it's already an improvement over what we have. Putting this into the code can be a follow up PR.

Mic92: We never upgrade to the latest kernel in nixos-hardware. This is what we do instead:

```
 boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "6.8") (lib.mkDefault pkgs.linuxPackages_latest);
```

tuxedo/pulse/14/gen3/default.nix
```diff
@@ -1,9 +1,24 @@
...
     ../../../../common/cpu/amd
     ../../../../common/cpu/amd/pstate.nix
     ../../../../common/cpu/amd/raphael/igpu.nix
     ../../../../common/pc/laptop
     ../../../../common/pc/laptop/ssd
   ];
+
+  # Fixing a power-issue with older kernels.
+  # When powered off, the battery does not turn off completely.
+  # Kernel 6.8.12 fixes this,
```
britter, line 17: @gabyx Amazing you found the exact version. I love this!

britter: @Mic92 WDYT?

Mic92:
@mergify queue

Mic92:
@mergify queue

tuxedo/pulse/14/gen3/default.nix
Mic92: Looks good.

PR #1017, opened 2024-06-26
gpu/amd: Remove `loadInInitrd` option

Lyndeno:
###### Description of changes

Option is now in nixpkgs under `hardware.amdgpu.initrd.enable` as of NixOS/nixpkgs@6a0b6a6b74a71be7c37b9d8bc4472cf087aff709

###### Things done

Tests completed at eval time only.

### DIFF
--- a/common/gpu/amd/default.nix
+++ b/common/gpu/amd/default.nix
@@ -1,24 +1,15 @@
-{ config, lib, ... }:
+{ lib, ... }:
 
 {
-  options.hardware.amdgpu.loadInInitrd = lib.mkEnableOption (lib.mdDoc
-    "loading `amdgpu` kernelModule at stage 1. (Add `amdgpu` to `boot.initrd.kernelModules`)"
-  ) // {
-    default = true;
-  };
-
   imports = [ ../24.05-compat.nix ];
-  config = lib.mkMerge [
-    {
-      services.xserver.videoDrivers = lib.mkDefault [ "modesetting" ];
+  config = {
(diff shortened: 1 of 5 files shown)


Mic92:
We would still need to have it for 24.05 for the time being.

Mic92:
> We would still need to have it for 24.05 for the time being.

Ah. Wait, we have it already backported here: https://github.com/NixOS/nixpkgs/pull/321245

common/gpu/amd/default.nix
```diff
@@ -1,24 +1,13 @@
-{ config, lib, ... }:
+{ lib, ... }:
 
 {
-  options.hardware.amdgpu.loadInInitrd = lib.mkEnableOption (lib.mdDoc
```
Mic92, line 4: We also need to fix this in all our modules.

Lyndeno: I will get that fixed

Lyndeno: Fixed this and reenabled it by default when the gpu/amd/default is imported.

Lyndeno:
> > We would still need to have it for 24.05 for the time being.
> 
> Ah. Wait, we have it already backported here: [NixOS/nixpkgs#321245](https://github.com/NixOS/nixpkgs/pull/321245)



> > We would still need to have it for 24.05 for the time being.
> 
> Ah. Wait, we have it already backported here: [NixOS/nixpkgs#321245](https://github.com/NixOS/nixpkgs/pull/321245)

Thanks for checking though, I did not think of that.

Mic92:
@mergify queue

Issue #992, opened 2024-06-20
Let's upstream all profiles in `common/` to nixos


Mic92:
Having generic hardware modules profiles available allows us to include them automatically
in tools like nixos-generate-config and co.

Hence I am proposing to upstream all our modules in `common` and remove them from this repository:

```
- [ ] ./cpu/intel/tiger-lake/default.nix
- [ ] ./cpu/intel/sandy-bridge/default.nix
- [ ] ./cpu/intel/kaby-lake/default.nix
- [ ] ./cpu/intel/cpu-only.nix
- [ ] ./cpu/intel/default.nix
  * Note: We are currently doing igpu  configuration in our CPU configuration. We should probably separate these two things when upstreaming, since not all Intel CPUs have a GPU.
- [ ] ./cpu/intel/jasper-lake/default.nix
- [ ] ./cpu/intel/elkhart-lake/default.nix
- [ ] ./cpu/intel/comet-lake/default.nix
- [ ] ./cpu/amd/raphael/igpu.nix
- [ ] ./cpu/amd/zenpower.nix
- [ ] ./cpu/amd/default.nix
- [ ] ./cpu/amd/pstate.nix
- [ ] ./pc/default.nix
- [ ] ./pc/ssd/default.nix
- [ ] ./pc/hdd/default.nix
- [ ] ./pc/laptop/acpi_call.nix
- [ ] ./pc/laptop/hdd/default.nix
- [ ] ./pc/laptop/default.nix
- [ ] ./pc/laptop/ssd
- [ ] ./gpu/nvidia/prime-sync.nix
- [ ] ./gpu/nvidia/prime.nix
- [ ] ./gpu/nvidia/disable.nix
- [ ] ./gpu/nvidia/default.nix
- [ ] ./gpu/amd/default.nix
- [ ] ./gpu/amd/sea-islands
- [ ] ./gpu/amd/sea-islands/default.nix
- [ ] ./gpu/amd/southern-islands
- [ ] ./gpu/amd/southern-islands/default.nix
- [ ] ./gpu/intel/default.nix
- [ ] ./gpu/intel/disable.nix
- [ ] ./gpu/intel.nix
- [ ] ./gpu/nvidia.nix
- [ ] ./gpu/nvidia-disable.nix
```

Open questions:

- Do we upstream them as module options or as importable profiles?

SuperSandro2000:
[...]

PR #1022, opened 2024-06-29
common: Move Intel generation-specific config from cpu to gpu

Lyndeno:
###### Description of changes

These modules only included gpu configuration, so they belong in that location.

###### Things done


apple/macbook-air/4/default.nix
```diff
@@ -3,7 +3,7 @@
 {
   imports = [
     ../.
-    ../../../common/cpu/intel/sandy-bridge
+    ../../../common/gpu/intel/sandy-bridge
```
Mic92, line 6: Sounds good, we have to think about what to do with our flake.nix that also references this file: https://github.com/NixOS/nixos-hardware/blob/3980e7816c99d9e4da7a7b762e5b294055b73b2f/flake.nix#L267

in https://github.com/NixOS/nixos-hardware/pull/991 someone wanted to expand the list of exports, but I would actually suggest to deprecate the old modules so that we can make progress on: https://github.com/NixOS/nixos-hardware/issues/992

Lyndeno: I am in favour of removing the common exports. They can be used by the existing model exports.

If a system needs the common modules, perhaps that system can be upstreamed to nixos-hardware.

Removing these exports does not block usage, one can reference the source file directly if needed.

Mic92: Ok. Could you remove those exports that are affected with warning? I believe there at least some users out there.

Lyndeno: I added a trace directly to the flake.nix. That way modules within nixos-hardware do not trigger any warning or trace.

Mic92:
@mergify queue

PR #1024, opened 2024-07-01
Framework 16 keyboard-wake fix

betalars:
###### Description of changes
Due to a firmware issue, the framework 16 will wake up with its lid closed when the display flexes upon the keyboard. This is a suggested workaround.

These questions should be answered before merging:

1. Are these device IDs consistent across different framework 16 notebooks *and different keyboard modules*? 
      - I have tested this on one more fw16, and it worked there. I think we used the same layout tho.
3. Do we want to also disable the Trackpad, and if so: how?
      - will create a review for that ...
4. This might be unexpected behavior, as it also prevents the device form waking up from keyboard input *when the lid is open*. Is that a worthwhile trade-off, and if not so: is there another way?
      -  The framework will become very hot inside a backpack, to the point where I am concerned about battery and storage health. Users on the forum reported the 82 degree warning being exceeded on the SSD. And as I pretty much never use wake by keyboard anyways, I would strongly support this workaround, while Framework figures out a firmware patch.

###### Things done

### DIFF
--- a/framework/16-inch/common/default.nix
+++ b/framework/16-inch/common/default.nix
@@ -19,6 +19,9 @@
   services.udev.extraRules = ''
     # Ethernet expansion card support
     ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="0bda", ATTR{idProduct}=="8156", ATTR{power/autosuspend}="20"
+    # Disabling keyboard waking up from suspend to prevent wakeups while lid is closed due to keyboard being presssed by display flexing. 
+    ACTION=="add", SUBSYSTEM=="usb", DRIVERS=="usb", ATTRS{idVendor}=="32ac", ATTRS{idProduct}=="0012", ATTR{power/wakeup}="disabled", ATTR{driver/1-1.1.1.4/power/wakeup}="disabled"
+    ACTION=="add", SUBSYSTEM=="usb", DRIVERS=="usb", ATTRS{idVendor}=="32ac", ATTRS{idProduct}=="0014", ATTR{power/wakeup}="disabled", ATTR{driver/1-1.1.1.4/power/wakeup}="disabled"
   '';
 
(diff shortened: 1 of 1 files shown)


xeniarose:
i'm a FW16 user, so for my input on the questions:

1. i have the same device ID for my keyboard module (the US ANSI keyboad with generic super key). maybe a more robust match could be anything that exposes an HID keyboard interface that has the framework device ID, but i'm not sure off the top of my head how to do that in udev
2. the trackpad appears to be i2c, present at /sys/bus/i2c/devices/i2c-PIXA3854:00/ and has a wakeup node that can be set to `disabled`. on my machine, this causes the touchpad to no longer wake
3. personally i don't intentionally use the keyboard/trackpad to wake, i only use the power button and imo that's the intuitive behavior. i don't think it's a huge deal to disable the keyboard/trackpad wake

xeniarose:
these two udev rules cover [all official keyboard variants, with product IDs mined from info.json](https://github.com/FrameworkComputer/qmk_firmware/tree/framework16-keyboards/keyboards/framework), and the trackpad

```
ACTION=="add", SUBSYSTEM=="usb", DRIVER=="usb", ATTR{idVendor}=="32ac", ATTR{idProduct}=="0012|0013|0014|0018|0019", ATTR{power/wakeup}="disabled"
ACTION=="add", SUBSYSTEM=="i2c", DRIVER=="i2c_hid_acpi", ATTR{name}=="PIXA3854:00", ATTR{power/wakeup}="disabled"
```

one issue is that sometimes the keyboard backlight will not turn off during sleep if its disabled for wakeup. i'm not sure how big of a deal this is besides wasting a little bit of power, and maybe it's completely a firmware thing that we can't do anything about, but i'm hoping there's some workaround possible

nyabinary:
Is this still needed?

Mic92:
Please re-open if you think this is still needed.

PR #1027, opened 2024-07-03
Fix Lenovo Thinkpad T14s not powering off

tobias-kuendig:
###### Description of changes

My Lenovo Thinkpad T14s Gen 4 does not propery power off after shutdown. The display goes blank but the CPU runs very hot and the fan starts to spin up. In this state, the battery is drained within around an hour.

Adding the `apm` parameter fixes this behaviour.

Maybe someone else can chime in if they had similar problems.

###### Things done

### DIFF
--- a/lenovo/thinkpad/t14s/amd/gen4/default.nix
+++ b/lenovo/thinkpad/t14s/amd/gen4/default.nix
@@ -1,0 +1,9 @@
+{ lib, pkgs, ... }:
+{
+  imports = [
+    ../.
+  ];
+
+  # Fix laptop not properly powering off during shutdown.
+  boot.kernelParams = [ "apm=power_off" ];
+}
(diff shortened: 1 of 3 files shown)


pinage404:
sometimes, i have the same issue on Framework 16

lenovo/thinkpad/t14s/amd/gen4/default.nix
```diff
@@ -0,0 +1,9 @@
+{ lib, pkgs, ... }:
```
Mic92, line 1: Can you add this model to flake.nix and the table in the README?

tobias-kuendig: Done!

Mic92:
@mergify queue

PR #1029, opened 2024-07-04
framework: refactor kmod for improved AMD support

### LINKED ISSUE #1028: Framework 13/16 AMD: hardware.framework.enableKmod requires kernel patch for kernel < 6.10
The [nix source for enabling this option](https://github.com/NixOS/nixos-hardware/blob/6e253f12b1009053eff5344be5e835f604bb64cd/framework/kmod.nix#L14) seems to reference [the README for the kernel module](https://github.com/DHowett/framework-laptop-kmod?tab=readme-ov-file#usage), which states that an additional patch series is required for Framework 13/16 AMD

Here's [the latest version of the patch series](https://lore.kernel.org/chrome-platform/20240403004713.130365-1-dustin@howett.net/), which i've applied to my kernel (6.6.36, nixos 24.05) via `boot.kernelPatches` and it appears to work for all functionality (edit: that is supported on FW16). The patch is currently [in the 6.10 RCs](https://elixir.bootlin.com/linux/v6.10-rc6/source/drivers/platform/chrome/cros_ec_lpc.c#L589) but it's not available in any current stable or LTS kernels afaict.

Perhaps adding an additional option which enables including the patch (via recompiling) on kernel < 6.10, or barring that, a note in the framework README, would be a good idea. One reason that having an actually functional framework kmod is convenient is that here on KDE plasma 6, it exposes the battery charge limit to the KDE settings GUI, which is pretty nice.

xeniarose:
###### Description of changes

Fixes #1028 

Changes several aspects of [framework-laptop-kmod](https://github.com/DHowett/framework-laptop-kmod) support:
- default value has been changed to true when NixOS >= 24.05 *and* linux kernel >=6.10; false otherwise
- adds required patch for framework AMD enablement when kernel <6.10
- add README section describing usage

Also: this may be an upstream nixpkgs issue, but the lack of a good lore.kernel.org fetcher is kind of surprising. I tried using [this sample code with `fetchpatch`](https://discourse.nixos.org/t/fetching-patches-from-lore-kernel-org/45594) initially, but got weird errors with it. I [implemented](https://git.lain.faith/haskal/dragnpkgs/src/commit/ecdf449848aaa8fe368922fb3cb28d2ce8b09be7/lib/fetchb4) a `fetchb4` fetcher for my own personal usage, which uses `b4` to download and format patches from <https://lore.kernel.org> and the skeleton of that is present in this PR using `runCommand` for simplicity

###### Things done

Edit: I've been running this on my own framework 16 for a bit now and everything seems to be working fine

### DIFF
--- a/framework/kmod.nix
+++ b/framework/kmod.nix
@@ -1,17 +1,42 @@
-{ config, lib, ... }:
-{
-  options.hardware.framework.enableKmod = lib.mkEnableOption
+{ config, lib, pkgs, ... }:
+let
+  kernel_version_compatible = lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.10";
+in {
+  options.hardware.framework.enableKmod = (lib.mkEnableOption
     "Enable the community created Framework kernel module that allows interacting with the embedded controller from sysfs."
-  // {
-    # Enable by default if on new enough version of NixOS
-    default = (lib.versionAtLeast (lib.versions.majorMinor lib.version) "24.05");
+  ) // {
+    # enable by default on NixOS >= 24.05 and kernel >= 6.10
+    default = lib.and
+      (lib.versionAtLeast (lib.versions.majorMinor lib.version) "24.05")
+      kernel_version_compatible;
   };
 
-  config = lib.mkIf config.hardware.framework.enableKmod {
-    boot.extraModulePackages = with config.boot.kernelPackages; [
+
+  config.boot = lib.mkIf config.hardware.framework.enableKmod {
+    extraModulePackages = with config.boot.kernelPackages; [
       framework-laptop-kmod
     ];
+
     # https://github.com/DHowett/framework-laptop-kmod?tab=readme-ov-file#usage
-    boot.kernelModules = [ "cros_ec" "cros_ec_lpcs" ];
+    kernelModules = [ "cros_ec" "cros_ec_lpcs" ];
+
+    # add required patch if enabled on kernel <6.10
+    kernelPatches = lib.mkIf (!kernel_version_compatible) [
+      rec {
+        name = "platform/chrome: cros_ec_lpc: add support for AMD Framework Laptops";
+        msgid = "20240403004713.130365-1-dustin@howett.net";
+        version = "3";
+        hash = "sha256-aQSyys8CMzlj9EdNhg8vtp76fg1qEwUVeJL0E+8w5HU=";
+        patch = pkgs.runCommandLocal "patch-${msgid}" {
+          nativeBuildInputs = with pkgs; [ b4 git cacert ];
+          SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
+
+          outputHash = hash;
+        } ''
+          export HOME="$TMP"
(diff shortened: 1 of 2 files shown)


framework/kmod.nix
```diff
@@ -1,17 +1,42 @@
-{ config, lib, ... }:
-{
-  options.hardware.framework.enableKmod = lib.mkEnableOption (lib.mdDoc
+{ config, lib, pkgs, ... }:
+let
+  kernel_version_compatible = lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.10";
```
Mic92, line 3: One can access config.boot.kernelPackages when setting kernelPatches? I always thought this would result in infinite recursion.

framework/kmod.nix
```diff
@@ -1,17 +1,42 @@
...
+    extraModulePackages = with config.boot.kernelPackages; [
       framework-laptop-kmod
     ];
+
     # https://github.com/DHowett/framework-laptop-kmod?tab=readme-ov-file#usage
-    boot.kernelModules = [ "cros_ec" "cros_ec_lpcs" ];
+    kernelModules = [ "cros_ec" "cros_ec_lpcs" ];
+
+    # add required patch if enabled on kernel <6.10
+    kernelPatches = lib.mkIf (!kernel_version_compatible) [
```
Mic92, line 24: How important is this kernel module? Because many users might not want to re-compile their kernel. Can we maybe default to a later kernel instead and disable this module if the kernel is too old?

nyabinary:
Bump :p

framework/kmod.nix
xeniarose: it works for me so i guess not. i'm not sure if there's theoretically an issue here, but if this is undefined behavior then i can refactor the patch

framework/kmod.nix
xeniarose: this line is also gated by `config.hardware.framework.enableKmod` above, which under this change defaults to `false` if the kernel version is too old. so i think it addresses what you want already

nyabinary:
image
kmod is not needed anymore in the latest kernels btw

paperdigits:
The kernel now fails to build with 6.11.

quatquatt:
> The kernel now fails to build with 6.11.

There was a patch yesterday that landed on stable and unstable that makes it work on 6.11. Should be fine now, it's working on my machine.

paperdigits:
Awesome I will give it another go today. Thanks!

framework/kmod.nix
Mic92: Maybe patches are applied after the fact.

Mic92:
What is the reason this is marked as draft?

xeniarose:
> What is the reason this is marked as draft?

it looks like you marked it as a draft during the review comments. if you think those issues still need to be addressed in some way, i'm happy to make further changes. otherwise, it would be fine to remove the draft status -- it doesn't look like i can do it myself

my personal daily laptop is running nixos-hardware from my own branch here and at this point it's been working with no issues for >6 months. my use case was solely to get the battery charge limit showing up in KDE, so i could extend my battery lifespan

Mic92:
@mergify queue

PR #1032, opened 2024-07-09
XPS 15-9560 fixed disabling the gpu

liamwb:
###### Description of changes
The current default module for the XPS 15-9560 includes the following
```
    # FIXME: remove this when bumblebee works again
    ../../../common/gpu/nvidia/disable.nix
```
Which completely disables the nvidia gpu. However, it seems like prime offload works fine, so I have replaced the (not working) bumblebee options with prime.

Everything is basically taken from the following [nixos forum post](https://discourse.nixos.org/t/help-getting-nvidia-gpu-working-on-dell-xps-15-9560/46885)

###### Things done

- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via Flake input

###### Questions I have
Currently, the `nvidia` module looks like this:
```
  # This runs only nvidia, great for games or heavy use of render applications

  ##### disable intel, run nvidia only and as default
  hardware.nvidia.prime = {
    # Bus ID of the Intel GPU.
    intelBusId = "PCI:0:2:0";

    # Bus ID of the NVIDIA GPU.
    nvidiaBusId = "PCI:1:0:0";
  };
```
As far as I can tell, the comment saying it runs "only nvidia" is a bit misleading? Rather, the imported `prime.nix` module seems to enable a specialization that does that.

Please let me know if I've done anything wrong, this is my first PR!


dell/xps/15-9560/default.nix
```diff
@@ -4,36 +4,22 @@
     ../../../common/cpu/intel/kaby-lake
     ../../../common/pc/laptop
     ./xps-common.nix
-
-    # FIXME: remove this when bumblebee works again
-    ../../../common/gpu/nvidia/disable.nix
+    ../../../common/gpu/nvidia
   ];
 
+  hardware.opengl = {
```
Mic92, line 10:
```suggestion
  hardware.graphics = {
```

dell/xps/15-9560/default.nix
```diff
@@ -4,36 +4,22 @@
...
-    # FIXME: remove this when bumblebee works again
-    ../../../common/gpu/nvidia/disable.nix
+    ../../../common/gpu/nvidia
   ];
 
+  hardware.graphics = {
+    enable = true;
+    driSupport = true;
+    driSupport32Bit = true;
+  };
```
Mic92, line 14:
```suggestion
  hardware.grapics.enable = true;
```

Mic92, line 14: driSupport is no longer and 32-bit drivers is probably a bit obsolete outside of older games:

```suggestion
  hardware.graphics.enable = true;
```

dell/xps/15-9560/default.nix
```diff
@@ -4,36 +4,18 @@
     ../../../common/cpu/intel/kaby-lake
     ../../../common/pc/laptop
     ./xps-common.nix
-
-    # FIXME: remove this when bumblebee works again
-    ../../../common/gpu/nvidia/disable.nix
+    ../../../common/gpu/nvidia
```
Mic92, line 7: Shouldn't be that prime?

```suggestion
    ../../../common/gpu/nvidia/prime
```

Mic92: Ah from reading the code it looks that if offload is not enabled, it doesn't use prime offload but prime sync? https://github.com/NixOS/nixpkgs/blob/bf99590fba1405cbc8c24154a2563a1f81b908d9/nixos/modules/hardware/video/nvidia.nix#L335

Lyndeno: Prime sync needs to be enabled explicitly, at least the NixOS option does.

I believe if neither is enabled, the igpu is used and the nvidia gpu is just available for compute.

Lyndeno: Should we import nvidia/offload such that the offload config and command is available?

Lyndeno:
The code you added already exists in 15-9560/nvidia/default.nix (minus the modesetting part)

Maybe we should just have the 15-9560/default.nix import the nvidia folder by default, and do nothing else.

Then the only change I think you need to make is to add the modesetting.enable line to 15-9560/nvidia/default.nix.

> As far as I can tell, the comment saying it runs "only nvidia" is a bit misleading? Rather, the imported prime.nix module seems to enable a specialization that does that.

No specialization, just makes offload and the offload command enable. In this case, the igpu is used by default, but the Nvidia gpu is configured to be available when desired via `nvidia-offload` or other methods like `switcherooctl` (if enabled)

liamwb:
My mistake! I think that just importing the nvidia folder is probably sensible then

Mic92:
@mergify queue

PR #1035, opened 2024-07-11
common-gpu-nvidia: drop libva-vdpau-driver, enable modesetting by default

getchoo:
#### Removing libva-vdpau-driver

libva-vdpau-driver has not seen a new release in over a decade and is a common cause for bugs in some applications. NixOS has also been using the now preferred nvidia-vaapi-driver since https://github.com/NixOS/nixpkgs/pull/162660, so there isn't much reason to keep this

#### Enabling modesetting by default

This is commonly required for many applications. As of https://github.com/NixOS/nixpkgs/pull/324921 this will also enable `nvidia-drm.fbdev=1`, possibly fixing issues such as https://github.com/NixOS/nixpkgs/issues/302059. This will hopefully be done upstream in NixOS as well with https://github.com/NixOS/nixpkgs/pull/326369

CC @Mic92 based on [this](https://github.com/NixOS/nixpkgs/issues/302059#issuecomment-2222143794) comment

###### Things done

### DIFF
--- a/common/gpu/nvidia/default.nix
+++ b/common/gpu/nvidia/default.nix
@@ -4,5 +4,5 @@
   imports = [ ../24.05-compat.nix ];
   services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
   # TODO: this will be a default after https://github.com/NixOS/nixpkgs/pull/326369
-  hardware.nvidia.modesetting.enable = true;
+  hardware.nvidia.modesetting.enable = lib.mkDefault true;
 }
--- a/common/gpu/nvidia/default.nix
+++ b/common/gpu/nvidia/default.nix
@@ -3,4 +3,6 @@
 {
   imports = [ ../24.05-compat.nix ];
   services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
+  # TODO: this will be a default after https://github.com/NixOS/nixpkgs/pull/326369
+  hardware.nvidia.modesetting.enable = true;
 }
--- a/common/gpu/nvidia/default.nix
+++ b/common/gpu/nvidia/default.nix
@@ -1,13 +1,6 @@
-{ lib, pkgs, ... }:
+{ lib, ... }:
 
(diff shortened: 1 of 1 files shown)


Mic92:
I will give @K900 a bit to review this change.

K900:
I don't really know what the situation is like on Nvidia.

getchoo:
The NixOS module PR gives some more context to this if you'd like to look it over, but to keep it brief: modesetting has been enabled OOTB on distros like Fedora for a while now and doesn't have any major known breakages. libva-vdpau-driver has also been know to cause many bugs (notoriously on [OBS](https://github.com/obsproject/obs-studio/issues/8056)) or simply not work at all

...I also don't think it's even used currently. At least on my system, `vainfo` reported nvidia-vaapi-driver was being used before and after I removed the vdpau translation layer here [here](https://github.com/getchoo/flake/commit/04f816863407c20f72e17d322abae4c284e81fb7). It probably requires [`LIBVA_DRIVER_NAME` to be set](https://wiki.archlinux.org/title/Hardware_video_acceleration#Configuring_VA-API) to take precedence over former; I can ofc double check this later today or tomorrow to confirm, though

common/gpu/nvidia/default.nix
```diff
@@ -1,13 +1,8 @@
...
   services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
-  hardware.graphics.extraPackages = [
-    (
-      if pkgs ? libva-vdpau-driver
-      then pkgs.libva-vdpau-driver
-      else pkgs.vaapiVdpau
-    )
-  ];
+  # TODO: this will be a default after https://github.com/NixOS/nixpkgs/pull/326369
+  hardware.nvidia.modesetting.enable = true;
```
Mic92, line 7:
```suggestion
  hardware.nvidia.modesetting.enable = lib.mkDefault true;
```

Mic92:
@mergify queue

PR #1042, opened 2024-07-16
lenovo-thinkpad-z13-gen2: move to asound.conf

toastal:
###### Description of changes

`sound.*` was removed upstream in Nixpkgs

###### Things done

- [x] Tested the changes in your own NixOS Configuration
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

PR #1048, opened 2024-07-18
Fix CI by disabling sound at lenovo-thinkpad-z13-gen2

superherointj:
lenovo-thinkpad-z13-gen2: disable sound to fix CI

Fixes CI


superherointj:
ping @toastal

superherointj:
> #1042
> 
> This is already opened. Rather than calling me out as a FIXME, can we just merge my open request?

Obviously I did not see your fix. Then, I would not do a duplicated lesser work.

Issue #1052, opened 2024-07-21
Unexpected / new "error: value is a path while a set was expected" when importing "nixos-hardware.nixosModules.asus-zephyrus-ga402x.amdgpu" or "nixos-hardware.nixosModules.asus-zephyrus-ga402x.nvidia"


mexisme:
I think PR #1046 has broken my importing of `nixos-hardware.nixosModules.asus-zephyrus-ga402x.amdgpu`, which was working beforehand, but now gives the error:
```
error: value is a path while a set was expected
```

This really confused me until (after some digging) I realised that `nixos-hardware.nixosModules.asus-zephyrus-ga402x` now pointed to a separate source path, rather than a direct import of the `default.nix` at [`//asus/zephyrus/ga402x/default.nix`](https://github.com/NixOS/nixos-hardware/blob/master/asus/zephyrus/ga402x/default.nix), which does export a `amdgpu` and `nvidia` attr.

This made it easy to switch between AMD only and NVidia boot specialisations, because disabling NVidia completely saved some battery — let alone for those times where the AMD GPU driver _or_ the NVidia driver have a regression.

NOTE:
Also breaks `lenovo-yoga-7-14ARH7.amdgpu` and `lenovo-yoga-7-14ARH7.nvidia`

mexisme:
// @NiklasGollenstede

mexisme:
I suspect it's not useful, but the full error message (for me) looks like:
```
error:
       … while calling the 'seq' builtin

         at /nix/store/qszplw617r895nbcprgyj139c9a3r0xs-source/lib/modules.nix:322:18:

          321|         options = checked options;
          322|         config = checked (removeAttrs config [ "_module" ]);
             |                  ^
          323|         _module = checked (config._module);

       … while evaluating a branch condition

         at /nix/store/qszplw617r895nbcprgyj139c9a3r0xs-source/lib/modules.nix:261:9:

          260|       checkUnmatched =
          261|         if config._module.check && config._module.freeformType == null && merged.unmatchedDefns != [] then
             |         ^
          262|           let

       … in the left operand of the AND (&&) operator

         at /nix/store/qszplw617r895nbcprgyj139c9a3r0xs-source/lib/modules.nix:261:72:

          260|       checkUnmatched =
          261|         if config._module.check && config._module.freeformType == null && merged.unmatchedDefns != [] then
             |                                                                        ^
          262|           let

       … in the left operand of the AND (&&) operator

[...]

PR #1053, opened 2024-07-21
Create asus-zephyrus-ga402x-amdgpu and asus-zephyrus-ga402x-nvidia entries

mexisme:
###### Description of changes

This PR:
- Replaces `asus-zephyrus-ga402x.amdgpu` with a `asus-zephyrus-ga402x-amdgpu` entry.
- Replaces `asus-zephyrus-ga402x.nvidia` with a `asus-zephyrus-ga402x-nvidia` entry.
- Makes `asus-zephyrus-ga402x` throw a deprecation warning.

FIXES: #1052 

Background:
The `asus-zephyrus-ga402x` provides an attr-set with `amdgpu` and `nvidia` entries, to allow users to choose whether to enable only the AMD-GPU driver, or also enable the NVidia driver with (by default) Prime enabled.

However, this attr-set style seems to be broken by PR #1046, which exports modules as paths, instead.
That change seems to cause an error of "value is a path while a set was expected".

###### Things done

### DIFF
--- a/asus/zephyrus/ga402x/ATTR-SET-DEPRECATION.md
+++ b/asus/zephyrus/ga402x/ATTR-SET-DEPRECATION.md
@@ -1,0 +1,16 @@
+# Deprecation of //asus/zephyrus/ga402x/default.nix
+
+Background:
+The `asus-zephyrus-ga402x` provides an attr-set with `amdgpu` and `nvidia` entries, to allow users
+to choose whether to enable only the AMD-GPU driver, or also enable the NVidia driver with (by
+default) Prime enabled.
+
+However, this attr-set style seems to be broken by [PR #1046](https://github.com/NixOS/nixos-hardware/pull/1046),
+which exports modules as paths, instead.
+That change seems to cause an error of "value is a path while a set was expected".
+
+[PR #1053](https://github.com/NixOS/nixos-hardware/pull/1053):
+- Replaced `asus-zephyrus-ga402x.amdgpu` with a `asus-zephyrus-ga402x-amdgpu` entry.
+- Replaced `asus-zephyrus-ga402x.nvidia` with a `asus-zephyrus-ga402x-nvidia` entry.
+- Made `asus-zephyrus-ga402x` throw a deprecation error.
+- [FIXES: #1052](https://github.com/NixOS/nixos-hardware/issues/1052)
--- a/asus/zephyrus/ga402x/default.nix
+++ b/asus/zephyrus/ga402x/default.nix
@@ -1,15 +1,10 @@
-## When using from a Flake, you can access these via imports of the attr key, e.g:
-#
-# imports = [
-#   nixos-hardware.nixosModules.asus-zephyrus-ga402x.amdgpu
-# ];
-#
-## or:
-# imports = [
-#   nixos-hardware.nixosModules.asus-zephyrus-ga402x.nvidia
-# ];
+{ ... }:
 
 {
-  amdgpu = import ./amdgpu;
-  nvidia = import ./nvidia;
+  assertions = [
+    {
+      assertion = false;
+      message = "Importing asus/zephyrus/ga402x/ (default.nix) directly is deprecated! #TODO: More details";
+    }
+  ];
 }
--- a/asus/zephyrus/ga402x/default.nix
+++ b/asus/zephyrus/ga402x/default.nix
@@ -4,7 +4,7 @@
   assertions = [
     {
       assertion = false;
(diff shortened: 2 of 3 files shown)


NiklasGollenstede:
Cool! I was about to suggest the same fix (but without the fancy deprecation warning, so this is nicer).

The exact same also applies to `lenovo-yoga-7-14ARH7`/`lenovo/yoga/7/14ARH7`. Can you add that as well?

And sorry that my #1046 broke your import. It was pretty clear that that change was going to break some dependents (but only in cases where the export was wrong or the importer relied on unspecified behavior).
In this case, the previous export pattern was wrong. `outputs.nixosModules` is supposed to be an "attrsett of modules" (things you can put for example in `imports` of other modules, with includes paths to files that define modules). Before it was apparently partially "attrsets of attrsets of modules".

mexisme:
> The exact same also applies to `lenovo-yoga-7-14ARH7`/`lenovo/yoga/7/14ARH7`. Can you add that as well?

Yeah, I realised the same thing applies earlier this evening.
I'll try to raise a PR tomorrow.

It does feel like this is a useful-enough pattern that it would be nice to make it more reusable...?

> And sorry that my #1046 broke your import. It was pretty clear that that change was going to break some dependents (but only in cases where the export was wrong or the importer relied on unspecified behavior).
> In this case, the previous export pattern was wrong.
> `outputs.nixosModules` is supposed to be an "attrset of modules" (things you can put for example in `imports` of other modules, with includes paths to files that define modules). Before it was apparently partially "attrsets of attrsets of modules".

Gotcha.

Though, it's equally a shame that the tests couldn't catch for the use-cases that were going to be affected, because the error sent me down several rabbit-holes before I realised where the cause might actually lie!

NiklasGollenstede:
> It does feel like this is a useful-enough pattern that it would be nice to make it more reusable...?

I am not sure what you mean by pattern. IMHO, that `default.nix` should never have existed. All other `default.nix` files are modules, that one isn't. It would in general be nice if `nixos-hardware` was a bit more ... structured, though I am not sure how realistic that is, given the amount of (partially old) code and that ideally everything would be tested across several hundred different hardware devices.

> it's equally a shame that the tests couldn't catch for the use-cases that were going to be affected, because the error sent me down several rabbit-holes

Yeah. Unfortunately `nixosModule`s and `overlay`s are pretty hard to duck-type reliably.
As I sair in the other PR, a newer Nix version would probably saved you most of that rabbit-hunting.

mexisme:
> > It does feel like this is a useful-enough pattern that it would be nice to make it more reusable...?
> 
> I am not sure what you mean by pattern. IMHO, that `default.nix` should never have existed. All other `default.nix` files are modules, that one isn't.

Ah, I don't mean the attr-set pattern that was in `default.nix`, but more that repeatable config patterns like this should be reusable among multiple hardware targets.
I'll have a think about this, and maybe raise a draft PR for discussion.

> It would in general be nice if `nixos-hardware` was a bit more ... structured, though I am not sure how realistic that is, given the amount of (partially old) code and that ideally everything would be tested across several hundred different hardware devices.

I think @Mic92 has mentioned that he would prefer hardware definitions to be in `nixpkgs` instead, which suggests to me that this repo is more than a little for supporting and/or providing configs that might not fully stable, as the best config for the hardware is still being worked out?
I know it took a bit of effort to get a working config for AMD 6k processors on Lenovo   ;-)
 
> > it's equally a shame that the tests couldn't catch for the use-cases that were going to be affected, because the error sent me down several rabbit-holes
> 
> Yeah. Unfortunately `nixosModule`s and `overlay`s are pretty hard to duck-type reliably. As I said in the other PR, a newer Nix version would probably saved you most of that rabbit-hunting.

No worries.

mexisme:
> And sorry that my https://github.com/NixOS/nixos-hardware/pull/1046 broke your import. It was pretty clear that that change was going to break some dependents (but only in cases where the export was wrong or the importer relied on unspecified behavior).
> In this case, the previous export pattern was wrong. outputs.nixosModules is supposed to be an "attrsett of modules" (things you can put for example in imports of other modules, with includes paths to files that define modules). Before it was apparently partially "attrsets of attrsets of modules".

[...]

> Yeah. Unfortunately `nixosModule`s and `overlay`s are pretty hard to duck-type reliably. As I sair in the other PR, a newer Nix version would probably saved you most of that rabbit-hunting.

Oh, and I should've said earlier:  thanks for the detailed replies, @NiklasGollenstede.  They've been both helpful and enlightening.

Mic92:
@mergify queue

PR #1055, opened 2024-07-23
Create lenovo-yoga-7-14ARH7-amdgpu and lenovo-yoga-7-14ARH7-nvidia entries

mexisme:
###### Description of changes

This PR:
- Replaces `lenovo-yoga-7-14ARH7.amdgpu` with a `lenovo-yoga-7-14ARH7-amdgpu` entry.
- Replaces `lenovo-yoga-7-14ARH7.nvidia` with a `lenovo-yoga-7-14ARH7-nvidia` entry.
- Makes `lenovo-yoga-7-14ARH7` throw a deprecation warning.

FIXES: #1052 

Background:
The `lenovo-yoga-7-14ARH7` provides an attr-set with `amdgpu` and `nvidia` entries, to allow users to choose whether to enable only the AMD-GPU driver, or also enable the NVidia driver with (by default) Prime enabled.

However, this attr-set style seems to be broken by PR #1046, which exports modules as paths, instead.
That change seems to cause an error of "value is a path while a set was expected".

###### Things done


Mic92:
@mergify queue

PR #1066, opened 2024-07-31
lenovo/legion/16arha7: fix kernel check for speaker patch

8bitbuddhist:
###### Description of changes

This fixes an incorrect comparison between kernel versions, which resulted in a build error. Now, we'll only apply the patch if the kernel version is below 6.10 (i.e. 6.9.X or older).

###### Things done


Mic92:
@mergify queue

PR #1082, opened 2024-08-21
malibal/aon/s1: Add initial intel-only support

Dietr1ch:
###### Description of changes

Add initial support for the [MALIBAL Aon S1](https://www.malibal.com/shop/laptops/aon-s1/).

###### Things done

### DIFF
--- a/malibal/aon/s1/README.wiki
+++ b/malibal/aon/s1/README.wiki
@@ -1,0 +1,79 @@
+= Malibal Aon S1 =
+
+== Tested Hardware ==
+
+* CPU: Intel(R) Core(TM) i7-13700H CPU @ 2.4GHz
+* RAM: 32 GB
+* HDD: 2 TiB SSD
+* Screen: 14" WQXGA+ (2880x1800)
+* Graphics: Intel Raptor Lake-P [Iris Xe Graphics]. nVidia graphics disabled preemptively.
+* Display: Wayland.
+* Input: trackpad
+
+== Firmware Configuration ==
+
+No tweaking to get a useful system, but some is needed.
+- Sleeping doesn't work properly and overheats the system.
+- Fans spin even when the system is at almost no load and CPUs report 30°C-35°C
+
+== Installation ==
+
+=== Before installation ===
+
+I don't recall what I did very well, probably just disabled secure boot as
+there's no nice and meaningful way of using it yet.
+
+=== After installation ===
+
+=== Optional ===
+
+* Run fwupd
+
+== Troubleshooting ==
+
+
+== Annoyances ==
+
+=== It's easy to tap the power button ===
+
+It's better to ignore short tapping the power button,
+
+```nix
+  services = {
+    logind = {
+      if options.hardware.ignorePowerButtonTap
+      extraConfig = ''
+        # Don’t shutdown when power button is short-pressed
+        HandlePowerKey=ignore
+      '';
+    };
+  };
+```
+
+== Broken Functionality ==
+
+=== Sleeping works, but overheats the machine ===
+
+This drains battery faster than keeping the machine on and it might get you backpack really warm.
+
+Best workaround I found so far is to increase the time to take an action, and use shutdown as the action.
+This is *not* a reasonable workaround for most.
+
+
+```nix
+  services = {
+    logind = {
+      if options.hardware.ignorePowerButtonTap
+      extraConfig = ''
+        # Shutdown until sleep/hibernate works
+        IdleAction=poweroff
+
+        # Ideal case (hibernation onto / isn't working for me, but probably works with a swap partition)
+        # When idle, sleep, but hibernate if it's been too long.
+        # IdleAction=hybrid-sleep
+      '';
+    };
+  };
+```
+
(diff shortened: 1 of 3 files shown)


malibal/aon/s1/default.nix
```diff
@@ -0,0 +1,71 @@
...
+
+  powerManagement = {
+    powertop.enable = lib.mkDefault true;
+  };
+
+  services = {
+    logind = {
+      extraConfig = ''
+        # Don’t shutdown when power button is short-pressed
+        HandlePowerKey=ignore
```
Mic92, line 57: This is the same behavior on other systems. Is there something specific why we disable it here?

malibal/aon/s1/default.nix
```diff
@@ -0,0 +1,71 @@
...
+    enableRedistributableFirmware = lib.mkDefault true;
+
+    graphics = {
+      enable = lib.mkDefault true;
+      enable32Bit = lib.mkDefault true;
+	    extraPackages = with pkgs; [
+	    	(if (lib.versionOlder (lib.versions.majorMinor lib.version) "23.11")
+         then vpl-gpu-rt
+         else intel-vaapi-driver)
+	    ];
```
Mic92, line 23: Our intel module enable this.

```suggestion
```

malibal/aon/s1/default.nix
```diff
@@ -0,0 +1,71 @@
...
+  imports = [
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../../common/cpu/intel
+    ../../../common/gpu/intel
+    ../../../common/gpu/nvidia/disable.nix
+  ];
+
+  hardware = {
+    enableAllFirmware = lib.mkDefault true;
```
Mic92, line 13: Which of the hardware files from here are you enabling? https://github.com/NixOS/nixpkgs/blob/797f7dc49e0bc7fab4b57c021cdf68f595e47841/nixos/modules/hardware/all-firmware.nix#L42
Please add this as a comment.

malibal/aon/s1/README.wiki
```diff
@@ -0,0 +1,44 @@
...
+=== Optional ===
+
+* Run fwupd
+
+== Troubleshooting ==
+
+
+== Broken Functionality ==
+
+=== Sleeping works, but overheats the machine ===
```
Mic92, line 37: Maybe some hardware prevents deeper sleep states?

malibal/aon/s1/default.nix
```diff
@@ -0,0 +1,71 @@
...
+	  initrd = {
+	    kernelModules = [
+        "i915"  # Early loading so the passphrase prompt appears on external displays
+      ];
+	    availableKernelModules = [
+        "nvme"
+        "xhci_pci"
+        "thunderbolt"
+        "usbhid"
+      ];
```
Mic92, line 37: We don't include nixos-generate-config here:

```suggestion
```

malibal/aon/s1/default.nix
```diff
@@ -0,0 +1,71 @@
...
+         else intel-vaapi-driver)
+	    ];
+    };
+  };
+
+  boot = {
+	  initrd = {
+	    kernelModules = [
+        "i915"  # Early loading so the passphrase prompt appears on external displays
+      ];
```
Mic92, line 31: Should be part of intel:

```suggestion
```

malibal/aon/s1/README.wiki
Dietr1ch: I'm not sure if it's just prevention, or there's something really wrong because I tested and leaving the machine idle, but on produces less heat than making it sleep.

I suspect there's something broken with the BIOS firmware right now.

malibal/aon/s1/default.nix
Dietr1ch: I have it because it's easy to press the power key as it's right above the backspace key.

I remember that I got the shutdown prompt on KDE when I tapped it, but I can't recall if this is the only thing I changed. I can test this this weekend

malibal/aon/s1/default.nix
Dietr1ch: I think it's missing the video processing library (`vpl-gpu-rt`),
  - https://github.com/NixOS/nixos-hardware/blob/master/common/gpu/intel/default.nix

I guess the real fix is to add a specific entry for Intel Raptor Lake GPUs and then update the GPUs that are supported by [VPL](https://github.com/intel/vpl-gpu-rt)

Mic92: Yes. If you could do this, this would be the preferred solution.

malibal/aon/s1/default.nix
Mic92: Ok. If you think this is a an easy to accidental press to make issue, please add a comment why this is the default. Also one concern is that it's not easy to override this way... Maybe leave it comment it out for now and enable it in your own configuration.

malibal/aon/s1/default.nix
Dietr1ch: I'm not exactly sure, so I guess I'm better leaving it off by now.

Between cpu, bluetooth, wifi and multiple controllers I guess everyone wants it, but I guess it's better to keep that explicit.

PR #1093, opened 2024-08-30
XPS 15 7590 (nvidia) use open-source driver

eg-ayoub:
###### Description of changes
set `hardware.nvidia.open = true;` for xps 15 7590 
This fixes error induced intentionally, see https://github.com/NixOS/nixpkgs/pull/337289#issuecomment-2314851863

###### Things done

### DIFF
--- a/dell/xps/15-7590/nvidia/default.nix
+++ b/dell/xps/15-7590/nvidia/default.nix
@@ -9,7 +9,7 @@
     powerManagement = {
       # Enable NVIDIA power management.
       enable = lib.mkDefault true;
-      
+
       # Enable dynamic power management.
       finegrained = lib.mkDefault true;
     };
@@ -21,5 +21,7 @@
       # Bus ID of the NVIDIA GPU.
       nvidiaBusId = lib.mkDefault "PCI:1:0:0";
     };
+
+    open = lib.mkDefault true;
   };
 }


Mic92:
@mergify queue

dell/xps/15-7590/nvidia/default.nix
```diff
@@ -21,5 +21,7 @@
       # Bus ID of the NVIDIA GPU.
       nvidiaBusId = lib.mkDefault "PCI:1:0:0";
     };
+
+    open = lib.mkDefault true;
```
Mic92, line 25: To avoid the conflicting option:

```suggestion
    open = true;
```

BCNelson:
Should we have a way to encode compatibility in a more general way? For instance does the logic to enable and disable the opensource driver belong in the shared gpu code?

eg-ayoub:
@BCNelson it might be. I just faced this on another one of my machines. 
Personally, I think it might cause unexplained breakage for some people. 
IIRC, the option's default description shows a condition comparing the package.version to 560. Maybe something along those lines would cause minimal breakage.

BCNelson:
> the option's default description shows a condition comparing the package.version to 560.

@eg-ayoub Where is this I can't seem to find it? All I can find is [this](https://github.com/NixOS/nixpkgs/blob/6e99f2a27d600612004fbd2c3282d614bfee6421/nixos/modules/hardware/video/nvidia.nix#L355) line

BCNelson:
@eg-ayoub I opened another PR #1094 with a bit more of a generic solution

eg-ayoub:
@BCNelson alright, I'll be closing this then.

PR #1094, opened 2024-08-31
Add configurations for nvidia microarchitectures

BCNelson:
And add logic to enable the opensource drivers if the NVIDIA driver package and micro architecture supports it

###### Things done

### DIFF
--- a/common/gpu/nvidia/ada-lovelace/default.nix
+++ b/common/gpu/nvidia/ada-lovelace/default.nix
@@ -6,5 +6,5 @@ in
   imports = [ ../. ];
 
   # enable the opensorce drivers if the package supports it
-  hardware.nvidia.open = lib.mkDefault (nividiaPackage ? open && nividiaPackage ? firmware);
+  hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
 }
--- a/common/gpu/nvidia/ada-lovelace/default.nix
+++ b/common/gpu/nvidia/ada-lovelace/default.nix
@@ -5,6 +5,6 @@ in
 {
   imports = [ ../. ];
 
-  # enable the opensorce drivers if the package supports it
+  # enable the open source drivers if the package supports it
   hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
 }
--- a/common/gpu/nvidia/ada-lovelace/default.nix
+++ b/common/gpu/nvidia/ada-lovelace/default.nix
@@ -1,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
+  hardware.nvidia.open = lib.mkDefault (nividiaPackage ? open && nividiaPackage ? firmware);
+}
--- a/common/gpu/nvidia/ampere/default.nix
+++ b/common/gpu/nvidia/ampere/default.nix
@@ -6,5 +6,5 @@ in
   imports = [ ../. ];
 
   # enable the opensorce drivers if the package supports it
-  hardware.nvidia.open = lib.mkDefault (nividiaPackage ? open && nividiaPackage ? firmware);
+  hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
 }
--- a/common/gpu/nvidia/ampere/default.nix
+++ b/common/gpu/nvidia/ampere/default.nix
@@ -5,6 +5,6 @@ in
 {
   imports = [ ../. ];
 
-  # enable the opensorce drivers if the package supports it
+  # enable the open source drivers if the package supports it
   hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
 }
--- a/common/gpu/nvidia/ampere/default.nix
+++ b/common/gpu/nvidia/ampere/default.nix
@@ -1,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
+  hardware.nvidia.open = lib.mkDefault (nividiaPackage ? open && nividiaPackage ? firmware);
+}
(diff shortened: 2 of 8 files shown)


common/gpu/nvidia/ada-lovelace/default.nix
```diff
@@ -0,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
```
Mic92, line 8:
```suggestion
  # enable the opensource drivers if the package supports it
```

common/gpu/nvidia/ada-lovelace/default.nix
```diff
@@ -0,0 +1,10 @@
...
+  hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
```
Mic92, line 9: What is this priority? Is lib.mkDefault not also sufficient?

common/gpu/nvidia/ampere/default.nix
```diff
@@ -0,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
+  hardware.nvidia.open = lib.mkOverride 990 (nividiaPackage ? open && nividiaPackage ? firmware);
```
Mic92, line 9: Same question as above.

flake.nix
```diff
@@ -295,6 +295,11 @@
       common-gpu-nvidia-sync = import ./common/gpu/nvidia/prime-sync.nix;
       common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
       common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
+      common-gpu-nvidia-ada-lovelace = import ./common/gpu/nvidia/ada-lovelace;
+      common-gpu-nvidia-ampere = import ./common/gpu/nvidia/ampere;
+      common-gpu-nvidia-turing = import ./common/gpu/nvidia/turing;
+      common-gpu-nvidia-pascal = import ./common/gpu/nvidia/pascal;
+      common-gpu-nvidia-maxwell = import ./common/gpu/nvidia/maxwell;
```
Mic92, line 302: Making this a public api, would work against https://github.com/NixOS/nixos-hardware/issues/992 a bit. Progress on that has been slow, but we made good progress on https://github.com/numtide/nixos-facter so I would prefer to not expose stuff in the flake.nix, which we later on have to deprecate again. The alternative would be to make this part of NixOS as well...

With that we could than automatically detect what nvidia generation is installed in a machine.

flake.nix
```diff
@@ -295,6 +295,11 @@
       common-gpu-nvidia-sync = import ./common/gpu/nvidia/prime-sync.nix;
       common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
       common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
+      common-gpu-nvidia-ada-lovelace = import ./common/gpu/nvidia/ada-lovelace;
```
Mic92, line 298: Does nvidia has any notion of version number for like intel 13th gen?
Assuming that new cards will be supported by the opensource driver, we could than just have a single comparison that would check if the generation number is bigger than X.

common/gpu/nvidia/ada-lovelace/default.nix
Mic92: Ok. I think I answered my question myself: https://github.com/NixOS/nixos-hardware/pull/1093#pullrequestreview-2273752181

common/gpu/nvidia/ampere/default.nix
```diff
@@ -0,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
```
Mic92, line 8:
```suggestion
  # enable the opensource drivers if the package supports it
```

common/gpu/nvidia/turing/default.nix
```diff
@@ -0,0 +1,10 @@
+{lib, config, ...}:
+let
+  nividiaPackage = config.hardware.nvidia.package;
+in
+{
+  imports = [ ../. ];
+
+  # enable the opensorce drivers if the package supports it
```
Mic92, line 8:
```suggestion
  # enable the opensource drivers if the package supports it
```

flake.nix
BCNelson: Unfortunately most other numbers that user facing have been co-opted for marketing purposes and are not reliable to figure out the micro Architecture of the GPU. For the most part they follow the major version of the cards 10xx 20xx and so on. However there are exceptions and those will cause headaches.

eg-ayoub:
I have a small question about this, 
considering the configs for other devices would still show an error until they set the architecture,
would it be possible to have a meaningful error message in this case ? 
Even if one is integrated into nixpkgs, will it guarantee users will know how to fix the nixos-hardware part properly when making their PRs ?

Mic92:
It's difficult the way this was implemented in nixpkgs to have such an error message.

Lyndeno (changes requested):
Can we add the Pascal dependency to the XPS 9560 while we are at it?

flake.nix
Mic92, line 302:
```suggestion
```

Lyndeno:
> I have a small question about this, considering the configs for other devices would still show an error until they set the architecture, would it be possible to have a meaningful error message in this case ? Even if one is integrated into nixpkgs, will it guarantee users will know how to fix the nixos-hardware part properly when making their PRs ?

I suggest we add the architecture to each model that uses nvidia before this is merged

Mic92:
> > I have a small question about this, considering the configs for other devices would still show an error until they set the architecture, would it be possible to have a meaningful error message in this case ? Even if one is integrated into nixpkgs, will it guarantee users will know how to fix the nixos-hardware part properly when making their PRs ?
> 
> I suggest we add the architecture to each model that uses nvidia before this is merged

I think this might be a lot of work for a single person. This is better parallelized.

Lyndeno:
> I think this might be a lot of work for a single person. This is better parallelized after merging.

True, I can take a look at affected modules once this is merged.

Mic92:
here is a checklist: https://github.com/NixOS/nixos-hardware/issues/1100

PR #1122, opened 2024-09-12
dell-precision-5560: init

matdibu:
###### Description of changes

Added config for `Dell Precision 5560`

###### Things done

- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via `<nixos-hardware>` or Flake input: https://codeberg.org/mateidibu/nix-config/commit/00d5d6947d9e028c7cc7d1662f070635eeb84275

### DIFF
--- a/dell/precision/5560/default.nix
+++ b/dell/precision/5560/default.nix
@@ -1,40 +1,25 @@
+{ lib, ... }:
 {
-  config,
-  lib,
-  ...
-}: {
   imports = [
     ../../../common/pc/laptop
     ../../../common/pc/laptop/ssd
     ../../../common/cpu/intel
     ../../../common/gpu/intel/tiger-lake
-    ../../../common/gpu/nvidia/prime.nix
     ../../../common/gpu/nvidia/turing
   ];
 
   hardware.enableRedistributableFirmware = lib.mkDefault true;
 
   boot = {
-    blacklistedKernelModules = ["nouveau"];
-    kernelModules = ["kvm-intel"];
-    kernelParams = ["i915.modeset=1"];
+    blacklistedKernelModules = [ "nouveau" ];
+    kernelModules = [ "kvm-intel" ];
+    kernelParams = [ "i915.modeset=1" ];
   };
 
-  hardware = {
-    graphics = {
-      enable = lib.mkDefault true;
-      enable32Bit = lib.mkDefault true;
-    };
-    nvidia = {
-      package = config.boot.kernelPackages.nvidiaPackages.beta;
-      modesetting.enable = lib.mkDefault true;
-      open = lib.mkDefault false;
-      prime = {
-        intelBusId = "PCI:0:2:0";
-        nvidiaBusId = "PCI:1:0:0";
-      };
-    };
-  };
+  services.xserver.videoDrivers = [ "nvidia" ];
 
-  powerManagement.cpuFreqGovernor = lib.mkDefault "performance";
+  hardware.nvidia.prime = {
+    intelBusId = "PCI:0:2:0";
+    nvidiaBusId = "PCI:1:0:0";
+  };
 }
--- a/dell/precision/5560/default.nix
+++ b/dell/precision/5560/default.nix
@@ -1,0 +1,40 @@
+{
+  config,
+  lib,
+  ...
+}: {
+  imports = [
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../../common/cpu/intel
+    ../../../common/gpu/intel/tiger-lake
+    ../../../common/gpu/nvidia/prime.nix
+    ../../../common/gpu/nvidia/turing
+  ];
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  boot = {
+    blacklistedKernelModules = ["nouveau"];
+    kernelModules = ["kvm-intel"];
+    kernelParams = ["i915.modeset=1"];
+  };
+
+  hardware = {
+    graphics = {
+      enable = lib.mkDefault true;
+      enable32Bit = lib.mkDefault true;
+    };
+    nvidia = {
+      package = config.boot.kernelPackages.nvidiaPackages.beta;
+      modesetting.enable = lib.mkDefault true;
+      open = lib.mkDefault false;
+      prime = {
+        intelBusId = "PCI:0:2:0";
+        nvidiaBusId = "PCI:1:0:0";
+      };
+    };
+  };
+
+  powerManagement.cpuFreqGovernor = lib.mkDefault "performance";
+}
--- a/dell/precision/5560/default.nix
+++ b/dell/precision/5560/default.nix
@@ -11,12 +11,8 @@
   hardware.enableRedistributableFirmware = lib.mkDefault true;
 
   boot = {
-    blacklistedKernelModules = [ "nouveau" ];
-    kernelModules = [ "kvm-intel" ];
     kernelParams = [ "i915.modeset=1" ];
   };
-
-  services.xserver.videoDrivers = [ "nvidia" ];
 
   hardware.nvidia.prime = {
     intelBusId = "PCI:0:2:0";
--- a/dell/precision/5560/README.md
+++ b/dell/precision/5560/README.md
@@ -1,0 +1,20 @@
+## Dell Precision 5560
+
+- Intel i7-11800H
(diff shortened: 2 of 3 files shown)


dell/precision/5560/default.nix
```diff
@@ -0,0 +1,40 @@
...
+    kernelParams = ["i915.modeset=1"];
+  };
+
+  hardware = {
+    graphics = {
+      enable = lib.mkDefault true;
+      enable32Bit = lib.mkDefault true;
+    };
+    nvidia = {
+      package = config.boot.kernelPackages.nvidiaPackages.beta;
```
Mic92, line 29: can you put a version number here, so that we can remove this once the stable nvidia package works as well?

dell/precision/5560/default.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  boot = {
+    blacklistedKernelModules = ["nouveau"];
+    kernelModules = ["kvm-intel"];
+    kernelParams = ["i915.modeset=1"];
+  };
+
+  hardware = {
+    graphics = {
+      enable = lib.mkDefault true;
```
Mic92, line 25: Is this not enabled by the nvidia module?

dell/precision/5560/default.nix
```diff
@@ -0,0 +1,40 @@
...
+      enable32Bit = lib.mkDefault true;
```
Mic92, line 26: https://github.com/NixOS/nixos-hardware/pull/1126#discussion_r1764676074

```suggestion
```

dell/precision/5560/default.nix
matdibu: removed it entirely

dell/precision/5560/default.nix
matdibu: you're right, I've deleted it now

dell/precision/5560/default.nix
matdibu: done

matdibu:
I've also added a README

dell/precision/5560/default.nix
```diff
@@ -0,0 +1,25 @@
...
+    ../../../common/cpu/intel
+    ../../../common/gpu/intel/tiger-lake
+    ../../../common/gpu/nvidia/turing
+  ];
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  boot = {
+    blacklistedKernelModules = [ "nouveau" ];
+    kernelModules = [ "kvm-intel" ];
```
Mic92, line 15: nixos-generate-config already detect this:

```suggestion
```

dell/precision/5560/default.nix
```diff
@@ -0,0 +1,25 @@
...
+    kernelParams = [ "i915.modeset=1" ];
+  };
+
+  services.xserver.videoDrivers = [ "nvidia" ];
```
Mic92, line 19: Part of the nvidia module:

```suggestion
```

dell/precision/5560/default.nix
matdibu: fixed

dell/precision/5560/default.nix
matdibu: fixed

dell/precision/5560/default.nix
```diff
@@ -0,0 +1,21 @@
...
+    ../../../common/pc/laptop/ssd
+    ../../../common/cpu/intel
+    ../../../common/gpu/intel/tiger-lake
+    ../../../common/gpu/nvidia/turing
+  ];
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  boot = {
+    kernelParams = [ "i915.modeset=1" ];
```
matdibu, line 14: @Mic92 what about this? I think it should at least be wrapped in a
`lib.mkIf (config.hardware.intelgpu.driver == "i915")`
but at that point, maybe it should be in `common` ?

Mic92: You are probably right.

```suggestion
    kernelParams = lib.mkIf (config.hardware.intelgpu.driver == "i915") [ "i915.modeset=1" ];
```

Also is this laptop not new enough to use `xe` anyway? Also isn't modset=1 the default since along time also?

```
lsmod | grep 'xe'
xe                   2875392  0
drm_gpuvm              45056  1 xe
drm_exec               12288  2 drm_gpuvm,xe
gpu_sched              65536  1 xe
drm_suballoc_helper    12288  1 xe
drm_ttm_helper         12288  1 xe
drm_buddy              20480  2 xe,i915
ttm                   102400  3 drm_ttm_helper,xe,i915
drm_display_helper    258048  2 xe,i915
cec                    73728  3 drm_display_helper,xe,i915
i2c_algo_bit           20480  2 xe,i915
video                  81920  2 xe,i915
backlight              28672  3 video,xe,i915

```

matdibu: it is, I left a note in the readme about using xe instead of i915
I simply don't know if we should default to that or not, since it's pretty new and might be less stable than i915

matdibu: also, I'm not sure about `i915.modeset`, I can see this locally:
```
$ modinfo i915 | grep modeset:
parm:           modeset:Use kernel modesetting [KMS] (0=disable, 1=on, -1=force vga console preference [default]) (int)
```

matdibu: are you on a precision5560 right now? I had to use `force_probe` to swap to xe

matdibu: `common/gpu/intel/tiger-lake/default.nix` does seem to default to xe when the kernel has support for it

Mic92: No. I am on 12th gen intel framework laptop. However you will get the xe driver by default when you update to linux 6.8, I believe:

https://github.com/NixOS/nixos-hardware/blob/master/common/gpu/intel/tiger-lake/default.nix

Mic92: So maybe just adding the conditional is enough. Please also add the reason, why you did add this flag.

matdibu: I just checked, on 6.10.11 I get this:
```
[    0.829046] xe 0000:00:02.0: Your graphics device 9a60 is not officially supported
               by xe driver in this kernel version. To force Xe probe,
               use xe.force_probe='9a60' and i915.force_probe='!9a60'
               module parameters or CONFIG_DRM_XE_FORCE_PROBE='9a60' and
               CONFIG_DRM_I915_FORCE_PROBE='!9a60' configuration options.
```

matdibu: that being said, I've been using xe since I got this laptop and everything works fine
it _is_ the earliest supported igpu by Xe
should we add the force_probe config here? or in gpu/intel/tigerlake ?

Mic92: Mhm. Maybe limit this to your laptop for now, to reduce the blast radius.

matdibu: fair enough, and thanks for reviewing btw!
but now I'm not sure the conditional in common/gpu/intel/tiger-lake/default.nix makes sense
could there be other tigerlake igpu device IDs that _are_ whitelisted for xe? if not, maybe we should remove xe from that config

matdibu: I looked through the kernel sources for xe, found this struct:
```
static const struct xe_device_desc tgl_desc = {
        .graphics = &graphics_xelp,
        .media = &media_xem,
        PLATFORM(TIGERLAKE),
        .has_display = true,
        .has_llc = true,
        .require_force_probe = true,
};
```
so tigerlake has `require_force_probe = true`

Mic92: Seems like it, only LUNARLAKE and BATTLEMAGE will actually use the driver without force probe: 

```
static const struct xe_device_desc lnl_desc = {
	PLATFORM(LUNARLAKE),
	.has_display = true,
};

static const struct xe_device_desc bmg_desc = {
	DGFX_FEATURES,
	PLATFORM(BATTLEMAGE),
	.has_display = true,
	.has_heci_cscfi = 1,
};
```

So we should probably adapt it for nixos-hardware as well.

Mic92: So this was introduced by @fidgetingbits in https://github.com/NixOS/nixos-hardware/pull/944

Mic92: https://github.com/NixOS/nixos-hardware/pull/1149

PR #1126, opened 2024-09-15
Add system76 galp5-1650

ahoneybun:
###### Description of changes

Adds the system76 galp5 (1650 version in this case)

###### Things done

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,8 @@ See code for all available configurations.
 | [Supermicro X12SCZ-TLN4F](supermicro/x12scz-tln4f)                     | `<nixos-hardware/supermicro/x12scz-tln4f>`              |
 | [System76 (generic)](system76)                                         | `<nixos-hardware/system76>`                             |
 | [System76 Darter Pro 6](system76/darp6)                                | `<nixos-hardware/system76/darp6>`                       |
-| [System76 Gazelle Gaze18](system76/gaze18)                             | `<nixos-hardware/system76/gaze18>`                      |
+| [System76 Gazelle 18](system76/gaze18)                             | `<nixos-hardware/system76/gaze18>`                      |
+| [System76 Galago Pro 5](system76/galp5-1650)                           | `<nixos-hardware/system76/galp5-1650>`                  | 
 | [Toshiba Chromebook 2 `swanky`](toshiba/swanky)                        | `<nixos-hardware/toshiba/swanky>`                       |
 | [Tuxedo InfinityBook v4](tuxedo/infinitybook/v4)                       | `<nixos-hardware/tuxedo/infinitybook/v4>`               |
 | [TUXEDO InfinityBook Pro 14 - Gen7](tuxedo/infinitybook/pro14/gen7)    | `<nixos-hardware/tuxedo/infinitybook/pro14/gen7>`       |
--- a/system76/galp5-1650/default.nix
+++ b/system76/galp5-1650/default.nix
@@ -1,0 +1,27 @@
+{ config, lib, pkgs, ... }:
+{
+  imports = [
+    ../.
+    ../../common/gpu/nvidia/prime.nix
+    ../../common/gpu/nvidia/ampere
+  ];
+
+  boot.initrd.kernelModules = [ "nvidia" ];
+
+  hardware.graphics = {
+    enable = lib.mkDefault true;
+    enable32Bit = lib.mkDefault true;
+  };
+
+  hardware.nvidia = {
+
+    # modesetting.enable = lib.mkDefault true;
+
+    powerManagement.finegrained = lib.mkDefault true;
+
+    prime = {
+      intelBusId = "PCI:0:2:0";
+      nvidiaBusId = "PCI:23:0:0";
+    };
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -280,6 +280,7 @@
         supermicro-x12scz-tln4f = import ./supermicro/x12scz-tln4f;
         system76 = import ./system76;
         system76-gaze18 = import ./system76/gaze18;
+        system76-galp5-1650 = import ./system76/galp5-1650;
         system76-darp6 = import ./system76/darp6;
         toshiba-swanky = import ./toshiba/swanky;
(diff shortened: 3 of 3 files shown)


ahoneybun:
The NVIDIA driver is loaded with this configuration:

```
[sudo] password for aaronh: 
  *-display                 
       product: i915drmfb
       physical id: 2
       bus info: pci@0000:00:02.0
       logical name: /dev/fb0
       version: 01
       width: 64 bits
       clock: 33MHz
       capabilities: pciexpress msi pm bus_master cap_list rom fb
       configuration: depth=32 driver=i915 latency=0 resolution=1920,1080
       resources: irq:184 memory:83000000-83ffffff memory:90000000-9fffffff ioport:1000(size=64) memory:c0000-dffff memory:892000000-898ffffff memory:8a0000000-97fffffff
  *-display
       physical id: 0
       bus info: pci@0000:23:00.0
       version: a1
       width: 64 bits
       clock: 33MHz
       capabilities: pm msi pciexpress bus_master cap_list rom
       configuration: driver=nvidia latency=0
       resources: iomemory:80-7f iomemory:80-7f irq:183 memory:81000000-81ffffff memory:880000000-88fffffff memory:890000000-891ffffff ioport:4000(size=128) memory:82000000-8207ffff
```

Though `system76-power graphics` does not seem to work:

[aaronh@garrus:~]$ system76-power graphics
daemon returned an error message: "The name com.system76.PowerDaemon was not provided by any .service files"

I have the `hardware.system76.enableAll` enabled in my configuration:

https://gitlab.com/ahoneybun/nix-configs/-/blob/main/hosts/x86_64/garrus/configuration.nix?ref_type=heads

I'll add add the system76 generic part as well of nixos-hardware to test.

ahoneybun:
Yep I needed that for the `system76-power` service!

[aaronh@garrus:~]$ system76-power graphics
nvidia

ahoneybun:
I think this **might** be that NixOS has version `1.1.23` of `system76-power`. I don't think this not working would be a blocking to add this system since the `gaze18` was not blocked though.

README.md
```diff
@@ -328,7 +328,8 @@ See code for all available configurations.
 | [Supermicro X12SCZ-TLN4F](supermicro/x12scz-tln4f)                     | `<nixos-hardware/supermicro/x12scz-tln4f>`              |
 | [System76 (generic)](system76)                                         | `<nixos-hardware/system76>`                             |
 | [System76 Darter Pro 6](system76/darp6)                                | `<nixos-hardware/system76/darp6>`                       |
-| [System76 Gazelle Gaze18](system76/gaze18)                             | `<nixos-hardware/system76/gaze18>`                      |
+| [System76 Gazelle 18](system76/gaze18)                             | `<nixos-hardware/system76/gaze18>`                      |
```
Mic92, line 331:
```suggestion
| [System76 Gazelle 18](system76/gaze18)                                 | `<nixos-hardware/system76/gaze18>`                      |
```

system76/galp5-1650/default.nix
```diff
@@ -0,0 +1,27 @@
+{ config, lib, pkgs, ... }:
+{
+  imports = [
+    ../.
+    ../../common/gpu/nvidia/prime.nix
+    ../../common/gpu/nvidia/ampere
+  ];
+
+  boot.initrd.kernelModules = [ "nvidia" ];
```
Mic92, line 9: Afaik this is enabled?

```suggestion
```

system76/galp5-1650/default.nix
```diff
@@ -0,0 +1,27 @@
...
+
+  hardware.graphics = {
+    enable = lib.mkDefault true;
+    enable32Bit = lib.mkDefault true;
```
Mic92, line 13: I am bit unsure if this is actually hardware configuration and not just about playing old games, which not everyone does:

```suggestion
```

system76/galp5-1650/default.nix
```diff
@@ -0,0 +1,27 @@
...
+  imports = [
+    ../.
+    ../../common/gpu/nvidia/prime.nix
+    ../../common/gpu/nvidia/ampere
+  ];
+
+  boot.initrd.kernelModules = [ "nvidia" ];
+
+  hardware.graphics = {
+    enable = lib.mkDefault true;
```
Mic92, line 12: Is this not enabled by the nividia module?

ahoneybun:
@Mic92 I just copied the config from the gaze18 so if we need to address things here I can make a PR for that model as well.

system76/galp5-1650/default.nix
ahoneybun: I'm pretty sure Steam will need it in general as it needs 32bit even if you have modern games.

Mic92: Ok, but than if we enable the module everywhere, should we not just do it in one place? i.e. nvidia.nix?

Mic92:
> @Mic92 I just copied the config from the gaze18 so if we need to address things here I can make a PR for that model as well.

Likely yes.

ahoneybun:
If I remove those options then rebuild and reboot `nvidia-smi` does not work so I think it is needed.

ahoneybun:
Alright @Mic92 I was incorrect and with the current changes `nvidia-smi` shows that the NVIDIA driver is loaded though `system76-power graphics` now reports this:

[aaronh@garrus:~]$ system76-power graphics
compute

so I think it is needed for NVIDIA mode at least but with Compute it should be similar to Hybrid.

ahoneybun:
Alright so I think compute is because I'm in the 6.6 kernel now and before I was in the 6.10 kernel series. I can't build 6.11 with the system76-module which is why I'm back in 6.6.

ahoneybun:
Alright with the current state of this PR puts the system in [PRIME Offload Mode](https://nixos.wiki/wiki/Nvidia#Optimus_PRIME_Option_A:_Offload_Mode) which I think is a good middle ground. You can use the NVIDIA GPU when needed and is pretty close to system76-power's Hybrid mode.

Mic92:
@mergify queue

PR #1128, opened 2024-09-15
Adding the Legion Pro 7 16IRX9H model

felipelalli:
Initially, I simply copied it from the 16IRX8H, but I am still having issues with WiFi and sound (to be resolved).

Ref.: https://psref.lenovo.com/Product/Legion/Legion_Pro_7_16IRX9H

###### Description of changes
Adding the Legion Pro 7 16IRX9H model, copy/paste from 16IRX8H.

###### Things done

### DIFF
--- a/lenovo/legion/16irx9h/default.nix
+++ b/lenovo/legion/16irx9h/default.nix
@@ -1,0 +1,35 @@
+{
+  lib,
+  config,
+  ...
+}: {
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/gpu/nvidia/prime.nix
+    ../../../common/gpu/nvidia/ada-lovelace
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../../common/hidpi.nix
+  ];
+
+  boot.initrd.kernelModules = ["nvidia"];
+  boot.extraModulePackages = [config.boot.kernelPackages.lenovo-legion-module config.boot.kernelPackages.nvidia_x11];
+
+  hardware = {
+    nvidia = {
+      modesetting.enable = lib.mkDefault true;
+      powerManagement.enable = lib.mkDefault true;
+      #
+      prime = {
+        intelBusId = "PCI:00:02:0";
+        nvidiaBusId = "PCI:01:00:0";
+      };
+    };
+  };
+
+  # Cooling management
+  services.thermald.enable = lib.mkDefault true;
+
+  # √(2560² + 1600²) px / 16 in ≃ 189 dpi
+  services.xserver.dpi = 189;
+}
--- a/lenovo/legion/16irx9h/default.nix
+++ b/lenovo/legion/16irx9h/default.nix
@@ -12,12 +12,10 @@
     ../../../common/hidpi.nix
   ];
 
-  boot.initrd.kernelModules = ["nvidia"];
-  boot.extraModulePackages = [config.boot.kernelPackages.lenovo-legion-module config.boot.kernelPackages.nvidia_x11];
+  boot.extraModulePackages = [config.boot.kernelPackages.lenovo-legion-module];
 
   hardware = {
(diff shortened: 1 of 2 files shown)


lenovo/legion/16irx9h/default.nix
```diff
@@ -0,0 +1,35 @@
...
+    ../../../common/cpu/intel
+    ../../../common/gpu/nvidia/prime.nix
+    ../../../common/gpu/nvidia/ada-lovelace
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../../common/hidpi.nix
+  ];
+
+  boot.initrd.kernelModules = ["nvidia"];
+  boot.extraModulePackages = [config.boot.kernelPackages.lenovo-legion-module config.boot.kernelPackages.nvidia_x11];
```
Mic92, line 16: the legion module is needed, but the rest should be already covered by the nvidia module.

lenovo/legion/16irx9h/default.nix
```diff
@@ -0,0 +1,35 @@
...
+
+  hardware = {
+    nvidia = {
+      modesetting.enable = lib.mkDefault true;
```
Mic92, line 20: Already enabled.

```suggestion
```

lenovo/legion/16irx9h/default.nix
Mic92:
```suggestion
  boot.extraModulePackages = [config.boot.kernelPackages.lenovo-legion-module];
```

felipelalli:
@Mic92 I just copied it from https://github.com/NixOS/nixos-hardware/blob/master/lenovo/legion/16irx8h/default.nix is that wrong too? Should I fix it too in this same PR?

Mic92:
@felipelalli I noticed that we actually don't add nvidiea to `boot.initrd.kernelModule`, which makes me wonder do you see any difference in having it in initrd? I.e. refreshing or changing resolution during boot?

felipelalli:
> @felipelalli I noticed that we actually don't add nvidea to `boot.initrd.kernelModule`, which makes me wonder do you see any difference in having it in initrd? I.e. refreshing or changing resolution during boot?

Sorry, but I couldn't understand your question. Apologies, I am a newbie. I just copied the setup of a laptop similar to mine.

Mic92:
@felipelalli I am talking about this line:

```
  boot.initrd.kernelModules = ["nvidia"];
```

When it's added does it flicker less during boot up. This line will put your nvidia driver into initrd and hence will initialize the graphic card earlier during boot.
I believe not having it would make your graphic driver to be somewhat re-initialized once filesystem are mounted, resulting in a refreshing the screen, potentially changing the resolution

lenovo/legion/16irx9h/default.nix
```diff
@@ -0,0 +1,35 @@
...
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/gpu/nvidia/prime.nix
+    ../../../common/gpu/nvidia/ada-lovelace
+    ../../../common/pc/laptop
+    ../../../common/pc/laptop/ssd
+    ../../../common/hidpi.nix
+  ];
+
+  boot.initrd.kernelModules = ["nvidia"];
```
Mic92, line 15: I don't think we should this option unless, we know why we are doing this:


```suggestion
```

https://github.com/NixOS/nixos-hardware/pull/1144

felipelalli: OK. But it need to be fixed as well on [16irx8h](https://github.com/NixOS/nixos-hardware/blob/master/lenovo/legion/16irx8h/default.nix), am I right?

felipelalli:
@Mic92 I have applied your suggestions, but pay attention that the same need to be done in 16irx8h.

felipelalli:
Replaced by #1227

PR #1138, opened 2024-09-24
macbook air 7

theotheroracle:
###### Things done
added macbook air 7 and added to flake.nix and readme, also added readme in folder describing issues with wifi driver

- [ x ] Tested the changes in your own NixOS Configuration
- [ x ] Tested the changes end-to-end by using your fork of `nixos-hardware` and

### DIFF
--- a/apple/macbook-air/7/default.nix
+++ b/apple/macbook-air/7/default.nix
@@ -1,0 +1,9 @@
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
+    ../.
+  ];
+
+  boot.extraModulePackages = lib.mkDefault [ config.boot.kernelPackages.broadcom_sta ];
+}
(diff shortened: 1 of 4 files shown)


apple/macbook-air/7/default.nix
```diff
@@ -0,0 +1,9 @@
+{ config, lib, pkgs, ... }:
+
+{
+  imports = [
+    ../.
+  ];
+
+  boot.extraModulePackages = lib.mkDefault [ config.boot.kernelPackages.broadcom_sta ];
```
Mic92, line 8: Does nixos-generate-config not also include this already?

theotheroracle: i'm not sure, the wifi didn't work initially, but i was able to get it working later, the macbook is currenly not booting, but if i get it fixed i'll let you know

Mic92: Afaik it should be in your hardware-configuration.nix. At least this is what I remember reading the perl code.

theotheroracle: ok, i removed it

Mic92:
@mergify queue

PR #1144, opened 2024-09-27
nvidia: add module to initrd

Mic92:
So it seems sometimes needed and sometimes not.
So my question is, do we always want nvidia in initrd? At least for intel, this is the way to go.

###### Things done

### DIFF
--- a/common/gpu/nvidia/default.nix
+++ b/common/gpu/nvidia/default.nix
@@ -2,6 +2,7 @@
 
 {
   imports = [ ../24.05-compat.nix ];
+  boot.initrd.kernelModules = [ "nvidia" ];
   services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
   # TODO: this will be a default after https://github.com/NixOS/nixpkgs/pull/326369
   hardware.nvidia.modesetting.enable = lib.mkDefault true;


Mic92:
cc @ahoneybun

Mic92:
cc @matdibu

Mic92:
cc @jopejoe1

Mic92:
cc @felipelalli

matdibu:
what's the size difference with this included?

Mic92:
> what's the size difference with this included?

~/git/nixpkgs main
% du -sh /nix/store/xsdqp0p1maw6pwj32zmdp4qanbxk5lqg-nvidia-open-6.6.52-560.35.03/lib/modules/6.6.52/kernel/drivers/video/nvidia.ko.xz
6.4M

Mic92:
It's noticeable.

Mic92:
Ok. Here some explanation from a nvidia forum: https://forums.developer.nvidia.com/t/when-it-makes-sense-to-load-nvidia-modules-into-initramfs-drm-kernel-mode-setting/198266

> Nowadays, the driver should be added to the initrd for a simple reason: timing. Modern pcie ssd systems boot so fast that without loading it from initrd chances are high that the driver isn’t initialized when the Xserver/Wayland compositor has already started so you end up with a broken graphics system.
> Sidenotes: things like ‘early kms’ and stuff don’t apply to the nvidia driver since it provides neither fbdev nor kms console. The console will always be running on vga fb or efifb depending on boot method (bios/efi).
nvidia-drm.modeset=1 thus doesn’t have any influence on the console but this is needed for wayland and enabling prime sync on hybrid graphics systems.

Mic92:
^ However I think we could also solve this more elegant with the right systemd service instead.

matdibu:
> the nvidia driver ... provides neither fbdev nor kms console
but
> Additionally, with the driver version 545 and above, you can also set the experimental nvidia_drm.fbdev=1 parameter, which is required to tell the NVIDIA driver to provide its own framebuffer device instead of relying on efifb or vesafb, which do not work under simpledrm
https://wiki.archlinux.org/title/NVIDIA#DRM_kernel_mode_setting

ahoneybun:
@matdibu do we want testing to confirm which works better? I can provide that if you do not have a NVIDIA laptop, just let me know what you want me to test.

matdibu:
I have a [tigerlake+turing laptop](https://github.com/NixOS/nixos-hardware/tree/master/dell/precision/5560), but I don't really know what the pitfalls of this are.

The way I see it, this is an increase in the initrd size (bad) that helps with nvidia compatibility (good).

Seeing as an "ESP is full" problem is much simpler to deal with than "nvidia gpu broke again", I'd say this is a good tradeoff.

As a sidenote, it seems my laptop's `initrd`'s are about 34M, with `[ "xe" "i915" "nvme" "zfs" "dm_mod" ]` in `config.boot.initrd.kernelModules`, so a 6.4M module is about +18% increase in size.

matdibu (approved):

matdibu:
on a second thought, it seems like the difference is not just the 6.4MB of `nvidia.ko.xz`
with `config.boot.initrd.kernelModules = [ "xe" "nvidia" "nvme" "zfs" "dm_mod" ]`, it jumped up to 82MB:

```
[root@precision5560:/boot/EFI/nixos]# ls -lh initrd-*
-rw-r----- 1 root root 34M Sep 29 15:12 initrd-6.10.11-23imcahzihf5glgzjwhxl3irffyj7qzzfvurexste5gqys6msviq.efi
-rw-r----- 1 root root 34M Sep 29 00:04 initrd-6.10.11-r5fu2inke2v5nd2ztb5uhrxefhefkzkq6aidaukct624ksjc53kq.efi
-rw-r----- 1 root root 34M Sep 30 15:20 initrd-6.10.11-umlundootqugfa66smudamim25zszme7fh5z6jld3r4ifmzmprwq.efi
-rw-r----- 1 root root 34M Oct  1 11:32 initrd-6.10.12-ibtipyr6t2xgch4f6a2ru6sn5wbsffyrdx4yk6e5mfuhfp7k2qsa.efi
-rw-r----- 1 root root 86M Oct  1 16:09 initrd-6.10.12-tlc4aog6x4kzdpeezd6jlfpoofbbnjtch6bipfnixrkpnlvnovhq.efi
-rw-r----- 1 root root 82M Oct  1 16:15 initrd-6.10.12-ufetcjq3xnljio6iaondc4ct63flrxjlgx2glczgetz6ttlydata.efi
```

matdibu:
we might need the help of someone who understands the current situation with nvidia, their fbdev, and all other related components.

there's a lot of old info on the web, and a lot of hearsay

maybe in the end, including nvidia in initrd doesn't bring any compatibility benefits

matdibu (changes requested):
need more info on whether this is needed on modern setups, and what setups is this explicitly needed for

Mic92:
> we might need the help of someone who understands the current situation with nvidia, their fbdev, and all other related components.
> 
> there's a lot of old info on the web, and a lot of hearsay
> 
> maybe in the end, including nvidia in initrd doesn't bring any compatibility benefits in modern setups (with recent hardware, drivers and kernel)

I also haven't had a confirmed case yet, where this was actually a problem.

PR #1149, opened 2024-09-28
intel gpu: only enable xe driver starting with lunar lake

Mic92:
###### Things done

### DIFF
--- a/asus/zenbook/ux371/default.nix
+++ b/asus/zenbook/ux371/default.nix
@@ -1,8 +1,6 @@
 {
   config,
   lib,
-  pkgs,
-  inputs,
   ...
 }:
 {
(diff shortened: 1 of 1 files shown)


Mic92:
cc @fidgetingbits

fidgetingbits:
Changes look okay to me, although not tested it as on mobile atm.

My tiger lake asus laptop is using xe without explicitly using force probe, so not sure how that works if other models are requiring force probe.. maybe the whitelist mentioned in the other issue?

I don't plan to go back to i915 on my tiger lake laptop as I haven't had any issues with xe. Seems like it would be useful to have an option to still enable it on tiger lake.

Mic92:
Your chipset is probably in this whitelist: https://github.com/torvalds/linux/blob/ad46e8f95e931e113cb98253daf6d443ac244cde/drivers/gpu/drm/xe/xe_reg_whitelist.c#L26C37-L26C55
We can also override the driver again for your laptop model only as it uses the driver without any force option.

Mic92:
I added it back for zenbook.

Mic92:
@mergify queue

PR #1151, opened 2024-09-28
Framework 16: use upstream libinput keyboard detection

me-and:
###### Description of changes

Use the upstream libinput keyboard detection quirks configuration to ensure Framework 16 laptop keyboards are detected and treated as internal keyboards. Notably, this means that non-ANSI Framework keyboards are correctly detected and treated as being internal keyboards rather than external ones.

Also add a comment noting that this configuration is part of more recent upstream libinput releases, and that it can be removed once we no longer need to support older versions of libinput. That probably means once NixOS 24.05 is out of support.

###### Things done


Mic92:
@mergify queue

PR #1160, opened 2024-09-29
add dell xps 9315

Laurent2916:
###### Description of changes

Added Dell XPS 9315

###### Things done

I've been using NixOS on this laptop for a while now, so here a small (non opinionated) config. Beside the camera and microphone (ipu6 crappy drivers), everything should work well, I'll contribute my findings/fixes in the future. 

- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via `<nixos-hardware>` or Flake input

### DIFF
--- a/dell/xps/13-9315/default.nix
+++ b/dell/xps/13-9315/default.nix
@@ -1,0 +1,30 @@
+{
+  lib,
+  pkgs,
+  ...
+}: {
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+  ];
+
+  # enable firmwares with a license allowing redistribution
+  # this includes the Wi-Fi and Bluetooth firmwares
+  hardware.enableRedistributableFirmware = true;
+
+  # touchpad uses I²C, so PS/2 is unnecessary
+  boot.blacklistedKernelModules = ["psmouse"];
+
+  # enable finger print sensor
+  # configure with `sudo fprintd-enroll <username>`
+  services.fprintd.enable = true;
+  services.fprintd.tod.enable = true;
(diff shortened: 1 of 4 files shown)


dell/xps/13-9315/default.nix
```diff
@@ -0,0 +1,30 @@
...
+  hardware.enableRedistributableFirmware = true;
+
+  # touchpad uses I²C, so PS/2 is unnecessary
+  boot.blacklistedKernelModules = ["psmouse"];
+
+  # enable finger print sensor
+  # configure with `sudo fprintd-enroll <username>`
+  services.fprintd.enable = true;
+  services.fprintd.tod.enable = true;
+  services.fprintd.tod.driver = pkgs.libfprint-2-tod1-goodix;
```
Mic92, line 23: I believe this is the default:

```suggestion
```

dell/xps/13-9315/default.nix
```diff
@@ -0,0 +1,25 @@
...
+  # this includes the Wi-Fi and Bluetooth firmwares
+  hardware.enableRedistributableFirmware = true;
+
+  # touchpad uses I²C, so PS/2 is unnecessary
+  boot.blacklistedKernelModules = ["psmouse"];
+
+  # enable finger print sensor
+  # configure with `sudo fprintd-enroll <username>`
+  services.fprintd.enable = true;
+  services.fprintd.tod.enable = true;
```
Mic92, line 18: I actually was wrong about the default of services.fprintd.tod.driver, it's only an example value.

Laurent2916: ah! that explains why I couldn't see your comment from the email I received.
Let me rollback

PR #1173, opened 2024-10-10
Add dell precision 5490

teekuningas:
###### Description of changes

Add minimal configuration for dell precision 5490.

###### Things done

### DIFF
--- a/dell/precision/5490/default.nix
+++ b/dell/precision/5490/default.nix
@@ -7,6 +7,7 @@
   # or even better: boot.kernelParams = pkgs.linuxPackages_latest;
   boot.kernelParams = [ "i915.force_probe=7d55" ];
 
+  hardware.nvidia.open = true;
   hardware.nvidia.prime = {
     intelBusId = "PCI:0:2:0";
     nvidiaBusId = "PCI:1:0:0";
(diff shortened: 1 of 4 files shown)


dell/precision/5490/default.nix
```diff
@@ -0,0 +1,15 @@
+{ lib, ... }:
+{
+  imports = [
+    ../../../common/gpu/nvidia
+  ];
+
+  # or even better: boot.kernelParams = pkgs.linuxPackages_latest;
+  boot.kernelParams = [ "i915.force_probe=7d55" ];
```
Mic92, line 8: Reading the text above, sounds like we want more something like:

```suggestion
  boot.kernelParams = lib.mkIf (lib.versionOlder pkgs.linux.version "6.6")  [ "i915.force_probe=7d55" ];
  boot.kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "6.6") (lib.mkDefault pkgs.linuxPackages_latest);
```

Please replace 6.6 with the kernel version you think is the newest working one.

teekuningas:
How about now? Updated to import ada-lovelace config allowing removal of "hardware.nvidia.open"-line. Not sure if we want to force latest kernel but at least with force_probe=7d55 the graphical sessions boot even with an older kernel, and now its set only when necessary.

Mic92:
@mergify queue

PR #1175, opened 2024-10-11
apple/t2: add tiny-dfr option

reidlabwastaken:
###### Description of changes

uses `mkMerge` to separate configuration options in the `apple/t2` module, and adds an option to disable `tiny-dfr`
`tiny-dfr` is still enabled by default to not cause issues with existing setups that rely on it

###### Things done

### DIFF
--- a/apple/t2/default.nix
+++ b/apple/t2/default.nix
@@ -49,71 +49,82 @@ let
 
 in
 {
-  options = {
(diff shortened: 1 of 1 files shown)


reidlabwastaken:
if it would make more sense, i could make this another profile instead

Mic92:
@mergify queue

PR #1182, opened 2024-10-16
fix vulkan on Gaze18

matteobongio:
###### Description of changes
Finally managed to get vulkan to work on the Gaze18

###### Things done
changed the boot.initrd.kernelModules

- [ x] Tested the changes in your own NixOS Configuration

### DIFF
--- a/system76/gaze18/default.nix
+++ b/system76/gaze18/default.nix
@@ -6,7 +6,7 @@
     ../../common/gpu/nvidia/ampere
   ];
 
-  boot.initrd.kernelModules = [ "nvidia" ];
+  boot.initrd.kernelModules = [ "nvidia" "i915" "nvidia_modeset" "nvidia_drm" ];
 
   hardware.graphics = {
     enable = lib.mkDefault true;


Mic92:
@mergify queue

Mic92:
@mergify dequeue

system76/gaze18/default.nix
```diff
@@ -6,7 +6,7 @@
     ../../common/gpu/nvidia/ampere
   ];
 
-  boot.initrd.kernelModules = [ "nvidia" ];
+  boot.initrd.kernelModules = [ "nvidia" "i915" "nvidia_modeset" "nvidia_drm" ];
```
Mic92, line 9: Could you add a comment here, how this fixes vulkan support. We had recently a discussion that about whether we should have nvidia in the initrd because it adds to the initrd size. So maybe it would be better if it would be made explicit here, that for this profile it's actually needed.

Mic92:
Please undraft when you think this is ready.

Mic92:
Closing because lack of of feedback. Please re-open if you continue to work on this.

PR #1198, opened 2024-10-21
common-cpu-intel-alder-lake: init

andresilva:
###### Description of changes

Creates a new config for intel alder lake (using the existing gpu module). Also updates the thinkpad x1 6th gen config to import the cpu architecture-specific module.

###### Things done

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -301,6 +301,7 @@
         common-cpu-amd-zenpower = import ./common/cpu/amd/zenpower.nix;
         common-cpu-amd-raphael-igpu = import ./common/cpu/amd/raphael/igpu.nix;
         common-cpu-intel = import ./common/cpu/intel;
+        common-cpu-intel-alder-lake = import ./common/cpu/intel/alder-lake;
         common-gpu-intel-comet-lake =
           deprecated "992" "common-gpu-intel-comet-lake"
             (import ./common/gpu/intel/comet-lake);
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -301,6 +301,7 @@
         common-cpu-amd-zenpower = import ./common/cpu/amd/zenpower.nix;
         common-cpu-amd-raphael-igpu = import ./common/cpu/amd/raphael/igpu.nix;
         common-cpu-intel = import ./common/cpu/intel;
+        common-cpu-intel-alder-lake = import ./common/cpu/intel/alder-lake;
```
Mic92, line 304: We are actually deprecating those: https://github.com/NixOS/nixos-hardware/issues/992
You can still use them if you want but we might remove them again.

Mic92:
```suggestion
```

andresilva: I'm using nixos-hardware as a flake, how can I use these if they are not exported?

Mic92:
```
{
  imports = [  "${nixos-hardware}/common/cpu/intel/alder-lake" ];
}
```

Mic92: Just be aware we will remove this after it's in nixpkgs.

andresilva: Thanks! 👍

Mic92:
@mergify queue

PR #1221, opened 2024-11-01
common-gpu-nvidia: expose architecture modules in `nixosModules`

sebaszv:
###### Description of changes

The architecture-specific modules for `common-gpu-nvidia` are used in the
various hardware modules for specific hardware depending on what is needed.

This works well, but it would be nice to have these exposed through the flake
`nixosModules` so that users using "generic" systems can still access these
modules in the same manner as all the others. Such as in my case, I have a
desktop system running an Nvidia Ampere GPU, but have to access the
`common/gpu/nvidia/ampere` module through its path since it is not exposed in
the flake itself.

I did not add these to the `README` since none of the other `common-gpu-nvidia`
modules are presented there. I opted to follow the trend. I can add them all if
suggested.

###### Things done

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -321,6 +321,12 @@
         common-gpu-nvidia-sync = import ./common/gpu/nvidia/prime-sync.nix;
         common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
         common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
+        common-gpu-nvidia-ada-lovelace = import ./common/gpu/nvidia/ada-lovelace;
+        common-gpu-nvidia-ampere = import ./common/gpu/nvidia/ampere;
+        common-gpu-nvidia-kepler = import ./common/gpu/nvidia/kepler;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -321,6 +321,12 @@
         common-gpu-nvidia-sync = import ./common/gpu/nvidia/prime-sync.nix;
         common-gpu-nvidia-nonprime = import ./common/gpu/nvidia;
         common-gpu-nvidia-disable = import ./common/gpu/nvidia/disable.nix;
+        common-gpu-nvidia-ada-lovelace = import ./common/gpu/nvidia/ada-lovelace;
```
Mic92, line 324: Same as in https://github.com/NixOS/nixos-hardware/pull/1198#discussion_r1808526854
We don't want to expose more common/ modules because of https://github.com/NixOS/nixos-hardware/issues/992
The linked PR also contains a work around on how to use them anyway, if you must.

Mic92: Alternative: Upstream them to nixpkgs and use them from there.

PR #1225, opened 2024-11-02
apple/t2: deprecate enableTinyDfr option

soopyc:
###### Description of changes
* Adds a deprecation warning to hardware.apple-t2.enableTinyDfr as it has been upstreamed to nixpkgs already. The package and option can be removed at the next release cycle to hopefully not break people's setup.

* Adds an assertion to conflict hardware.apple-t2.enableTinyDfr with hardware.apple.touchBar. Having both enabled causes both tiny-dfr to fight for the same display device.

###### Things done

### DIFF
--- a/apple/t2/default.nix
+++ b/apple/t2/default.nix
@@ -109,6 +109,16 @@ in
       '';
     })
     (lib.mkIf t2Cfg.enableTinyDfr {
+      warnings = [
+        ''
+          hardware.apple-t2.enableTinyDfr is deprecated since the module has been upstreamed as hardware.apple.touchBar. 
+          This option will be removed from the apple/t2 profile when NixOS 24.11 is released.
+        ''
+      ];
+      assertions = lib.optionals (lib.versionAtLeast (lib.versions.majorMinor lib.version) "24.11") [{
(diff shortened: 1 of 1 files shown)


soopyc:
hmm, stable tests shouldn't break... i suspect this is due to a bug in the test flake: https://github.com/NixOS/nixos-hardware/blob/master/tests/flake.nix#L65

Mic92:
@mergify rebase

soopyc:
should the test flake lock also be updated here? tests with the updated lockfile passes except one: 
```
  error: A definition for option `hardware.tuxedo-drivers.enable' is not of type `boolean'. Definition values:
  - In `/nix/store/57c0fgv5xzbn9ck203ivg5bw9fifhg2w-source/nixos/modules/hardware/tuxedo-drivers.nix':
      {
        _type = "override";
        content = true;
        priority = 1000;
      }
...
The following 1 test(s) failed:
 'x86_64-linux.nixos-unstable-tuxedo-aura-15-gen1'
```

Mic92:
@soopyc yes. We probably should automate that.

soopyc:
resolved in #1231

PR #1236, opened 2024-11-10
framework: workaround display issues on AMD GPUs

blitz:
###### Description of changes

Framework users (including me) have issues with PSR. We can trade some battery life for stability here. So far there is no workaround except disabling PSR.

See:
- https://community.frame.work/t/fedora-kde-becomes-suddenly-slow/58459
- https://gitlab.freedesktop.org/drm/amd/-/issues/3647

###### Things done

### DIFF
--- a/framework/13-inch/common/amd.nix
+++ b/framework/13-inch/common/amd.nix
@@ -5,8 +5,17 @@
     ../../../common/gpu/amd
   ];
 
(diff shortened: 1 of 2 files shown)


miketth:
Great to see this added here. I had this option in my config for ages and I was never sure if this was the one that fixed my GPU hangs and why it did.

blitz:
There is some hope that this might be fixed in the next linux-firmware version: https://gitlab.freedesktop.org/drm/amd/-/issues/3647#note_2652400

If this doesn't pan out, I'd vote for just disabling PSR until AMD gets their act together.

PR #1247, opened 2024-11-14
Add information on Gigabyte B650M Aorus Elite AX

PanCakeConnaisseur:
There is no fix AFAIK, but maybe this information is useful nonetheless.


Mic92:
@mergify queue

PR #1253, opened 2024-11-16
framework: Add `hardware.framework.laptop13.audioEnhancement` option

miketth:
###### Description of changes

This PR adds the `hardware.framework.laptop13.audioEnhancement` option, which adds filters that enhance the audio coming from the laptop speakers. This config yields vastly enhanced audio, including much better bass.

These techniques are heavily inspired by the Asahi project's [asahi-audio](https://github.com/AsahiLinux/asahi-audio) repo and contains an equalizer transformed from the "Lappy McTopface" preset from [ee-framework-presets](https://github.com/ceiphr/ee-framework-presets) (both MIT-licensed).

I am unsure about putting the option under `hardware.framework.laptop13`, but this applies to Framework Laptop 13, but not the 16 and there wasn't an existing option for this subset.

I had this in my personal config for quite a bit of time, but decided to upstream it, in case it helps others. It isn't perfect by all means, but it makes the speakers sound much better than their defaults.

###### Things done

### DIFF
--- a/framework/13-inch/common/audio.nix
+++ b/framework/13-inch/common/audio.nix
@@ -1,0 +1,382 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.hardware.framework.laptop13.audioEnhancement;
+in
+{
+  options = {
+    hardware.framework.laptop13.audioEnhancement = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Create a new audio device called "Framework Speakers",
+          which applies sound tuning before sending the audio out to the speakers.
+          This option requires PipeWire and WirePlumber.
+
+          The filter chain includes the following:
+            - Pyschoacoustic bass enhancement
+            - Loudness compensation
+            - Equalizer
+            - Slight compression
+
+          This option has been optimised for the Framework Laptop 13 AMD 7040 series, but should work on all models.
+
+          Before applying, ensure the speakers are set to 100%,
+          because the volumes compound and the raw speaker device will be hidden by default.
+
+          You might also need to re-select the default output device.
+
+          In some cases, the added bass will vibrate the keyboard cable leading to a rattling sound,
+          a piece of foam can be used to mitigate this.
+        '';
+      };
+
+      hideRawDevice = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = ''
+          Hide the raw speaker device.
+          This option is enabled by default, because keeping the raw speaker device can lead to volume conflicts.
+        '';
+      };
+
+      rawDeviceName = lib.mkOption {
+        type = lib.types.str;
+        example = "alsa_output.pci-0000_c1_00.6.analog-stereo";
+        description = ''
+          The name of the raw speaker device. This will vary by device.
+          You can get this by running `pw-dump | grep -C 20 pci-0000`.
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable (let
+    outputName = cfg.rawDeviceName;
+    prettyName = "Framework Speakers";
+
+    # These are pre-made decibel to linear value conversions, since Nix doesn't have pow().
+    # Use the formula `10 ** (db / 20)` to calculate.
+    db = {
+      "-18.1" = 0.1244514611771385;
+      "-5.48" = 0.5321082592667942;
+      "-4.76" = 0.5780960474057181;
+      "8.1" = 2.5409727055493048;
+      "-36" = 1.5848931924611134e-2;
+    };
+
+    json = pkgs.formats.json { };
+
+    # The filter chain, heavily inspired by the asahi-audio project: https://github.com/AsahiLinux/asahi-audio
+    filter-chain = json.generate "filter-chain.json" {
+      "node.description" = prettyName;
+      "media.name" = prettyName;
+      "filter.graph" = {
+        nodes = [
+          # Psychoacoustic bass extension,
+          # it creates harmonics of the missing bass to fool our ears into hearing it.
+          {
+            type = "lv2";
+            plugin = "https://chadmed.au/bankstown";
+            name = "bassex";
+            control = {
+              bypass = 0;
+              amt = 1.2;
+              sat_second = 1.3;
+              sat_third = 2.5;
+              blend = 1.0;
+              ceil = 200.0;
+              floor = 20.0;
+            };
+          }
+          # Loudness compensation,
+          # it ensures that the sound profile stays consistent across different volumes.
+          {
+            type = "lv2";
+            plugin = "http://lsp-plug.in/plugins/lv2/loud_comp_stereo";
+            name = "el";
+            control = {
+              enabled = 1;
+              input = 1.0;
+              fft = 4;
+            };
+          }
+          # 8-band equalizer,
+          # it tries to lessen frequencies where the laptop might resonate,
+          # and tries to make the frequency curve more pleasing;
+          # this is the "Lappy McTopface" profile (https://github.com/ceiphr/ee-framework-presets)
+          # further tuned for the Framework Laptop 13 AMD 7040 series
+          # and might need some tuning on other models.
+          {
+            type = "lv2";
+            plugin = "http://lsp-plug.in/plugins/lv2/para_equalizer_x8_lr";
+            name = "fw13eq";
+            control = {
+              mode = 0;
+              react = 0.2;
+              zoom = db."-36";
+
+              fl_0 = 101.0;
+              fml_0 = 0;
+              ftl_0 = 5;
+              gl_0 = db."-18.1";
+              huel_0 = 0.0;
+              ql_0 = 4.36;
+              sl_0 = 0;
+              wl_0 = 4.0;
+
+              fl_1 = 451.0;
+              fml_1 = 0;
+              ftl_1 = 1;
+              gl_1 = db."-5.48";
+              huel_1 = 3.125e-2;
+              ql_1 = 2.46;
+              sl_1 = 0;
+              wl_1 = 4.0;
+
+              fl_2 = 918.0;
+              fml_2 = 0;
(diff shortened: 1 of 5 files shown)


framework/13-inch/common/audio.nix
```diff
@@ -0,0 +1,368 @@
...
+              input = 1.0;
+              fft = 4;
+            };
+          }
+          # 8-band equalizer,
+          # it tries to lessen frequencies where the laptop might resonate,
+          # and tries to make the frequency curve more pleasing;
+          # this is the "Lappy McTopface" profile (https://github.com/ceiphr/ee-framework-presets)
+          # further tuned for the Framework Laptop 13 AMD 7040 series
+          # and might need some tuning on other models.
```
Mic92, line 98: How would one tune this?

framework/13-inch/common/audio.nix
```diff
@@ -0,0 +1,368 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.hardware.framework.laptop13.audioEnhancement;
+in
+{
+  options = {
+    hardware.framework.laptop13.audioEnhancement = {
+      enabled = lib.mkOption {
```
Mic92, line 8:
```suggestion
      enable = lib.mkOption {
```

framework/13-inch/common/audio.nix
```diff
@@ -0,0 +1,368 @@
...
+        default = true;
+        description = ''
+          Hide the raw speaker device.
+          This option is enabled by default, because keeping the raw speaker device can lead to volume conflicts.
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enabled (let
```
Mic92, line 43:
```suggestion
  config = lib.mkIf cfg.enable (let
```

framework/13-inch/common/audio.nix
```diff
@@ -0,0 +1,368 @@
...
+    outputName = "alsa_output.pci-0000_c1_00.6.analog-stereo";
```
Mic92, line 44: I think this part needs to be configureable.
On framework 13 intel 13 Gen it's called:

```
 node.name = "alsa_output.pci-0000_00_1f.3.analog-stereo"
 node.nick = "92HD95 Analog"
 media.class = "Audio/Sink"
```

Mic92: I changed the configuration, but the result was no I now have no audio: https://github.com/Mic92/nixos-hardware/commit/23ef53a7c46b4f3b155ce74e1e6bd7886054f431#diff-2dbd4c2fc043321e7d7fd56ee926f400d2e48223add74cfd1aa2a867bc499e74R174
Toggling the option and restarting the pipewire systemd service brings audio back.
I did have my audio on 100% before activating.

framework/13-inch/common/audio.nix
miketth: The way I tuned it is that I used EasyEffects with plugin GUIs enabled and added the equalizer there. I tweaked the knobs until it sounded good and exported the options to a txt and used those numbers here.

framework/13-inch/common/audio.nix
miketth: You might have to re-select the default output device the first time you use this. I'll add that to the description (and the configurable device name) soon

framework/13-inch/common/audio.nix
Mic92: Do you know how can get the underlying filter names in easy effects that you are using here?

framework/13-inch/common/audio.nix
Mic92: Ah. Yes. Now I got it working!

framework/13-inch/common/audio.nix
Mic92: Good enough.

framework/13-inch/common/audio.nix
```diff
@@ -0,0 +1,382 @@
...
+        default = true;
+        description = ''
+          Hide the raw speaker device.
+          This option is enabled by default, because keeping the raw speaker device can lead to volume conflicts.
+        '';
+      };
+
+      rawDeviceName = lib.mkOption {
+        type = lib.types.string;
+        default = "alsa_output.pci-0000_c1_00.6.analog-stereo";
```
Mic92, line 45: We should not set any default here and instead set the defaults in the concrete model profiles. i.e. your AMD model and the intel 13th gen one.
Having no default forces users to set this option as without the option it won't likely work on their device.

```suggestion
        example = "alsa_output.pci-0000_c1_00.6.analog-stereo";
```

framework/13-inch/common/audio.nix
miketth: This is a bit complicated, but here's how I did it:
1. Open EasyEffect and set up the equalizer how you like it
2. Click "Show native window", this will open the LV2 plugin window
3. On the native window, there's a "Save" button which lets you save the equalizer state
4. I then trimmed the unused bands (EasyEffects uses a 32-band LR equalizer)


EasyEffect
Native plugin window


To get the plugin names, you can do this:
1. Run `systemctl --user cat wireplumber` and take note of the `LV2_PATH` env var
2. In the folder that `LV2_PATH` is set to, you'll find all the plugin packages specified in `passthru.requiredLv2Packages`.
3. Change into the package's folder and grep for the name.
4. Once you find the plugin's file, you can get the name from there.
For example, for me `LV2_PATH` is `/nix/store/c0pza0jwsa78h87lylr2jhi4aw3l8llm-wireplumber-lv2-plugins/lib/lv2`, I can then `cd` into `/nix/store/c0pza0jwsa78h87lylr2jhi4aw3l8llm-wireplumber-lv2-plugins/lib/lv2/lsp-plugins.lv2` and run `grep -Ri parametric .`.

framework/13-inch/common/audio.nix
miketth: Good point!

framework/13-inch/13th-gen-intel/default.nix
```diff
@@ -1,7 +1,9 @@
-{
+{ lib, ... }: {
   imports = [
     # Same config as 12th Gen. The chipsets and mainboard are similar enough
     # that no separate configuration is needed.
     ../12th-gen-intel
   ];
+
+  hardware.framework.laptop13.audioEnhancement.rawDeviceName = lib.mkDefault "alsa_output.pci-0000_00_1f.3.analog-stereo";
```
Mic92, line 8: Now we are only missing 11th gen and 12th gen.

Minionflo: I just checked 12th and 13th gen are the same

Mic92: Thanks! Now I am just waiting for 11th Gen.

miketth: Thank you for adding the defaults!

leo60228:
I tested this on my device and the filtering is also applied to headphones, which doesn't seem like the expected behavior.

Mic92:
@miketth @leo60228 any idea how to avoid this? Are you using the headphone check?

leo60228:
The instructions from https://github.com/luisbocanegra/linux-guide-split-audio-ports seem to work to split it into separate PCMs, but are quite a few steps.

leo60228:
I'm not sure this is the best solution for nixos-hardware, but I added this to my personal configuration: https://github.com/leo60228/dotfiles/commit/d8edbb9871c73be79761a3220706bb03678472f4

leo60228:
Using UCM instead of ACP seems to have the side effect of making headphones and speakers appear as separate PipeWire nodes, even though it doesn't change that they use the underlying PCM. This patch achieves this, though I'm not sure it should be upstreamed: https://github.com/leo60228/alsa-ucm-conf/commit/616f1eb856149047e642d58ea2a03bbdc92631f0

(Instead of rebuilding the world, you can override the path to the `ucm2` folder with the `ALSA_CONFIG_UCM2` environment variable. This does not appear to be mentioned in documentation.)

miketth:
Sorry, I missed that part completely as I don't use the headphone jack either... I'll look into this soon-ish, but if I don't find anything cleaner, it might be the best to go with the solution from @leo60228

Mic92:
If the audio configuration becomes more complex than this, it might be worth considering starting a new project. Audio filters and equalizer is a bit of a different skill set from what we usually have in nixos-hardware. It would be cool to host these profiles in a separate repository and maybe reference them here? Just my thoughts, nothing set in stone.

Titaniumtown:
Cool stuff! I did this myself using easyeffects: https://git.gardling.com/titaniumtown/dotfiles/src/commit/555a049637da8d679e0f5712f366e19a976ddd16/home-manager/progs/framework-13-easyeffects.nix

I'll switch over to doing it through nixos-hardware. glad to see other people found this useful.

taha-yassine:
> EDIT: This doesn't work at all and I get no audio output :( I'm on a framework 13 12th gen intel

It worked for me after a reboot.

Titaniumtown:
Rebooting also didn't fix it :(

llakala:
Can reproduce, not getting this working with 7640u. Sound *is* working on some level, as I hear the little *bling* from when I adjust the audio volume. But whenever I try to play audio within an app, it just loads infinitely and the audio never starts playing.

miketth:
If this does not work for you, you might want to try and re-select the "Framework Speakers" output even if it shows up as selected. This seems to be a visual bug with desktop environments.

As for the audio jack, unfortunately the fix for that seems really complex, and I don't use it, so I wouldn't be able to daily-drive and test that part. If someone wants to implement the headphone jack fix, I think it's a good PR candidate if it is readable. As for now, that's a caviat and can be worked around by not hiding the backing device and switching between the two manually.

nyabinary:
This causes artifacting for me when the system is under load and plays audio, for example while playing a game, or when multiple tabs are open in power saving mode.

Titaniumtown:
@chpatrick Yea, this feature is broken.

PR #1254, opened 2024-11-18
feat(Dell Precision 5530): Added Nvidia support and some other feature

KeyZox71:
###### Description of changes

Added Nvidia GPU support with prime, fwupd and thermald (also the Intel coffee lake config)

###### Things done

### DIFF
--- a/dell/precision/5530/default.nix
+++ b/dell/precision/5530/default.nix
@@ -1,22 +1,47 @@
+{ lib, config, ... }:
+
 {
   imports = [
-    ../../../common/cpu/intel
     ../../../common/pc/laptop
     ../../../common/pc/laptop/ssd
+    ../../../common/gpu/nvidia/pascal
+    ../../../common/gpu/nvidia/prime.nix
+    ../../../common/cpu/intel/coffee-lake
   ];
 
-  boot.kernelParams = [
-    # fix lspci hanging with nouveau
-    # source https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1803179/comments/149
-    "acpi_rev_override=1"
-    "acpi_osi=Linux"
-    "nouveau.modeset=0"
-    "pcie_aspm=force"
-    "drm.vblankoffdelay=1"
-    "nouveau.runpm=0"
-    "mem_sleep_default=deep"
-    # fix flicker
-    # source https://wiki.archlinux.org/index.php/Intel_graphics#Screen_flickering
-    "i915.enable_psr=0"
-  ];
+  boot = {
+    kernelParams = [
+      # fix lspci hanging with nouveau
+      # source https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1803179/comments/149
+      "acpi_rev_override=1"
+      "acpi_osi=Linux"
+      "nouveau.modeset=0"
+      "pcie_aspm=force"
+      "drm.vblankoffdelay=1"
+      "nouveau.runpm=0"
+      "mem_sleep_default=deep"
+      # fix flicker
+      # source https://wiki.archlinux.org/index.php/Intel_graphics#Screen_flickering
+      "i915.enable_psr=0"
+      "nvidia_drm.modeset=1"
+    ];
+  };
+
+  hardware = {
+    nvidia = {
+      open = lib.mkDefault false;
+      nvidiaSettings = lib.mkDefault true;
+      modesetting.enable = lib.mkDefault true;
+      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.stable;
+      prime = {
+        intelBusId = lib.mkDefault "PCI:0:2:0";
+        nvidiaBusId = lib.mkDefault "PCI:1:0:0";
+      };
+    };
+  };
+  # This will save you money and possibly your life!
+  services = {
+    fwupd.enable = lib.mkDefault true;
+    thermald.enable = lib.mkDefault true;
+  };
 }


dell/precision/5530/default.nix
```diff
@@ -1,22 +1,47 @@
...
-    "pcie_aspm=force"
-    "drm.vblankoffdelay=1"
-    "nouveau.runpm=0"
-    "mem_sleep_default=deep"
-    # fix flicker
-    # source https://wiki.archlinux.org/index.php/Intel_graphics#Screen_flickering
-    "i915.enable_psr=0"
-  ];
+  boot = {
+    kernelParams = lib.mkDefault [
```
Mic92, line 13: For lists we should not use lib.mkDefault because it won't merge:

```suggestion
    kernelParams = [
```

Mic92:
@mergify squash

Mic92:
@mergify queue

gamersekofy:
Just to clarify, Nvidia GPU support on this laptop isn't as "complete" as it is on Windows, right? I'm specifically talking about power management. The NixOS wiki states that only Turing and later support Nvidia's finegrained power management, which can completely power off a GPU when it's not in use. Correct me if I'm wrong, but enabling the Nvidia card on this laptop will constantly keep it on, resulting in power drain, right?

I wanna switch to NixOS permanently and I'm trying to fully understand the Nvidia situation on Linux.

dell/precision/5530/default.nix
```diff
@@ -1,22 +1,47 @@
...
+  };
+
+  hardware = {
+    nvidia = {
+      open = lib.mkDefault false;
+      nvidiaSettings = lib.mkDefault true;
+      modesetting.enable = lib.mkDefault true;
+      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.stable;
+      prime = {
+        intelBusId = lib.mkDefault "PCI:0:2:0";
```
Mic92, line 37: As far as I know this setup will not power on the Nvidia GPU all the time but uses intel unless requested otherwise.

gamersekofy: Woah, really? All this time I've been under the assumption that the Nvidia GPU would be turned on and consuming power even when not in use due to its inability to enter the RTD3 power state on Linux. My strong desire to use NixOS led me to develop two theories to cope:

1. I'm being silly and the card actually doesn't draw power unless needed, because that's stupid. 
2. Lack of RTD3 on the Pascal generation GPUs applies to both Linux and Windows, so the GPU would behave the same and consume the same power on NixOS as it does on Windows. 

My desire also made me create a specialization that enabled Nvidia drivers, and I'd reboot and enable it every time I wanted to play a game, then reboot again and disable afterwards for battery savings. 

How has your experience with battery life been with this laptop (on NixOS), if you use it regularly?

KeyZox71:
> Just to clarify, Nvidia GPU support on this laptop isn't as "complete" as it is on Windows, right? I'm specifically talking about power management. The NixOS wiki states that only Turing and later support Nvidia's finegrained power management, which can completely power off a GPU when it's not in use. Correct me if I'm wrong, but enabling the Nvidia card on this laptop will constantly keep it on, resulting in power drain, right?
> 
> I wanna switch to NixOS permanently and I'm trying to fully understand the Nvidia situation on Linux.

For my part, I didn't see any difference in terms of battery after enabling the Nvidia GPU on my laptop. So 2 possibility :
- The GPU was always powered on and draw has much power when properly enabled or disabled
- Or when I don't use it, it powers off (And i barely use it) so I never noticed the difference

_to note that I can only measure the power limit automatically set by the laptop and not the power drawn by the GPU (I did not go really far there is probably a way to measure it but didn't try much)_

But what I'm sure is that the difference isn't noticeable, so not really a problem (I have a pretty big battery(97Wh), so it could also be due to that)

dell/precision/5530/default.nix
KeyZox71: > 2\. Lack of RTD3 on the Pascal generation GPUs applies to both Linux and Windows, so the GPU would behave the same and consume the same power on NixOS as it does on Windows.

That is probably true, as it is a technology added only to newer GPU and if this generation does not have it all OSes are affected (Windows probably more affected by this because it tends to work poorly with a hybrid laptop that does not have the ability to power off completely)

gamersekofy: Wonderful! Gonna reinstall NixOS today (for the fifth time) and hopefully stay for good

gamersekofy:
That's good to hear. The main thing that fueled my assumption was tests I did using `powertop`. I found that having the Nvidia drivers enabled added ~4W to the power consumption and I attributed this to my (possibly flawed) understanding of how the driver was working. 

Another thing which didn't help was my initial experience with NixOS, which was a few years ago and lasted a little over a month. This was during the time of Plasma 5/X11 defaults, and I found that having the Nvidia drivers enabled made my battery drain much faster and my laptop would get hot even if I wasn't doing anything. Now that I think about it, back then, running `nvidia-smi` would always show xorg as a running process that was using the GPU; afaik this wasn't the case with Wayland when I tested again.

KeyZox71:
> Another thing which didn't help was my initial experience with NixOS, which was a few years ago and lasted a little over a month.

Yeah NixOS (and Linux) has improved quite a lot in the past years, due to the gaming on Linux becoming a thing, so give it a try it could be a good moment 👍🏻 

> This was during the time of Plasma 5/X11 defaults, and I found that having the Nvidia drivers enabled made my battery drain much faster and my laptop would get hot even if I wasn't doing anything. Now that I think about it, back then, running nvidia-smi would always show xorg as a running process that was using the GPU; afaik this wasn't the case with Wayland when I tested again.

For that I could advise going to a tiling windows manager (like Hyprland) so you could, one, lower your power consumption and two understand more deeply your system and control more of what it does. Compared to a desktop environment (like KDE or GNOME) that manage a lot more things and some automatically which can lead to problem like that

Hoping your NixOS journey will be fun 😄

gamersekofy:
That's awesome! It feels good to leave Windows for good lol. 

Regarding Hyprland, I gave it a try the last time I had NixOS installed and it was very cool. I never finished configuring it 100% though; my waybar layout was a little messed up and network-manager would always ask for my WiFi password after a reboot. I can't wait to go back and finish my customizations :)

PR #1258, opened 2024-11-19
lenovo/thinkpad/t480s: Automatically downgrade kernel to 6.10

Pandapip1:
###### Description of changes

See https://bugzilla.kernel.org/show_bug.cgi?id=219352

###### Things done

### DIFF
--- a/lenovo/thinkpad/t480s/default.nix
+++ b/lenovo/thinkpad/t480s/default.nix
@@ -1,4 +1,9 @@
-{ lib, ... }:
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
 
 {
   imports = [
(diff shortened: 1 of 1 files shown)


lenovo/thinkpad/t480s/default.nix
```diff
@@ -9,4 +14,10 @@
   ];
 
   services.throttled.enable = lib.mkDefault true;
+
+  # There is currently a kernel regression with the psmouse driver. As such, the kernel version is held back to 6.10.
+  # See https://bugzilla.kernel.org/show_bug.cgi?id=219352
+  boot.kernelPackages = lib.mkIf (lib.versionAtLeast config.boot.kernelPackages "6.11") (
+    lib.warn "Linux kernel held back to 6.10 due to regression: https://bugzilla.kernel.org/show_bug.cgi?id=219352" pkgs.linuxPackages_6_10
```
Mic92, line 21: Should we not just warn in this case? The 6.10 kernel might get removed at this point.

Pandapip1: The default kernel right now is 6.6 iirc, so I doubt that's gonna happen.

Mic92: The kernel package referenced here is 6.10. It's not an lts kernel. But the actual issue of this code is that it's an infinite recursion. You are checking `config.boot.kernelPackage` while writing to the same option.

Pandapip1: Yea, this hasn't been tested. TBH I completely forgot about this PR; I somehow forgot to put it on my tracking board after opening it.

Pandapip1:
Recent kernel versions seem to have fixed this. Closing.

PR #1267, opened 2024-12-02
enable acpi.ec_no_wakeup in Thinkpad T14 AMD Gen 5

kowale:
###### Description of changes

This PR should remove constant battery drain specific to this device (and others, but I only have this one to test) as suggested in e.g.

- https://bbs.archlinux.org/viewtopic.php?id=298895
- https://bbs.archlinux.org/viewtopic.php?id=234913

###### Things done

- [x] Tested the changes in your own NixOS Configuration (<https://github.com/kowale/homelab/commit/a46e5dea72e70b5d65e9956993947a5c624040c0#diff-37d03cb9538bb32044b321a8a2cad7aafe9ec114d6b9f83e1d40e917edcc14f4R46> is working on my machine. Before this commit, I had to keep my laptop charging at all times, as it would drain in a few hours, even on suspend, unless fully shutdown. After this commit, I can leave my laptop with its lid closed for 10 hours at the expense of about 4% of battery!)
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` (I was already using this nixos-hardware input in my config, so counting this as done, as it should be identical, right?)


kowale:
I see the tests are failing for

- `x86_64-linux.nixos-stable-focus-m2-gen1`
- `x86_64-linux.nixos-unstable-focus-m2-gen1`

but these don't seem related. Flaky tests?

PR #1268, opened 2024-12-03
E14-intel: import intel comet lake

SuperSandro2000:
The E14 has a i5-10210U which is from comet lake.

see https://www.intel.com/content/www/us/en/products/sku/195436/intel-core-i510210u-processor-6m-cache-up-to-4-20-ghz/specifications.html

###### Things done

### DIFF
--- a/lenovo/thinkpad/e14/intel/default.nix
+++ b/lenovo/thinkpad/e14/intel/default.nix
@@ -3,7 +3,7 @@
 {
   imports = [
     ../.
-    ../../../../common/cpu/intel
+    ../../../../common/cpu/intel/comet-lake
   ];
 
   services.throttled.enable = lib.mkDefault true;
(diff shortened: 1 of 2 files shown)


lenovo/thinkpad/e14/intel/default.nix
```diff
@@ -3,8 +3,10 @@
 {
   imports = [
     ../.
-    ../../../../common/cpu/intel
+    ../../../../common/cpu/intel/comet-lake
   ];
 
+  hardware.intelgpu.vaapiDriver = "intel-media-driver";
```
Mic92, line 9: Do we not set this in the intel module? Maybe we should.

SuperSandro2000: As far as I understood it, it by default pulls in intel-media-driver and intel-vaapi-driver but I don't think we need the later.

Mic92: We have several intel generation profiles configuring `intel-media-driver`. If this is the right one for intel comet lake, we should do the same.

SuperSandro2000: Pushed that

PR #1279, opened 2024-12-18
dell-precision-5560: remove redundant config and enable fwupd

matdibu:
###### Description of changes

- removed redundant `i915.modeset=1`
- enabled `services.fwupd`

###### Things done


Mic92:
@Lassulus I think you also disabled mergify in other repositories. Can you only exclude nixpkgs instead?

Lassulus:
> @Lassulus I think you also disabled mergify in other repositories. Can you only exclude nixpkgs instead?

done, unsuspended mergify again and removed access from nixpgs

PR #1280, opened 2024-12-18
pine64-rock64: init

matdibu:
###### Description of changes

Added `pine64-rock64`

###### Things done

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -271,6 +271,7 @@
         olimex-teres_i = import ./olimex/teres_i;
         pcengines-apu = import ./pcengines/apu;
         pine64-pinebook-pro = import ./pine64/pinebook-pro;
+        pine64-rock64 = import ./pine64/rock64;
         pine64-rockpro64 = import ./pine64/rockpro64;
         pine64-star64 = import ./pine64/star64;
         protectli-vp4670 = import ./protectli/vp4670;
--- a/pine64/rock64/default.nix
+++ b/pine64/rock64/default.nix
@@ -1,0 +1,26 @@
+{
+  boot = {
+    initrd.kernelModules = [
+      "pcie_rockchip_host"
+      "phy_rockchip_pcie"
+      "sdhci_pci"
+    ];
+    kernelModules = [
+      "panfrost"
+      "rockchipdrm"
+      "rockchip_dfi"
+      "rockchip_rga"
+      "rockchip_isp1"
+      "rockchip_saradc"
+      "rockchip_thermal"
+      "rockchip_vdec"
+      "snd_soc_rockchip_i2s"
+      "rk_crypto"
+      "dwmac_rk"
+      "rk3399_dmc"
+      "v4l2_h264"
+      "v4l2_mem2mem"
+      "v4l2_vp9"
+    ];
+  };
+}
--- a/README.md
+++ b/README.md
@@ -321,6 +321,7 @@ See code for all available configurations.
 | [Panasonic Let's Note CF-LX4](panasonic/letsnote/cf-lx4)               | `<nixos-hardware/panasonic/letsnote/cf-lx4>`            |
 | [PC Engines APU](pcengines/apu)                                        | `<nixos-hardware/pcengines/apu>`                        |
 | [PINE64 Pinebook Pro](pine64/pinebook-pro/)                            | `<nixos-hardware/pine64/pinebook-pro>`                  |
+| [PINE64 Rock64](pine64/rock64/)                                        | `<nixos-hardware/pine64/rock64>`                        |
 | [PINE64 RockPro64](pine64/rockpro64/)                                  | `<nixos-hardware/pine64/rockpro64>`                     |
 | [PINE64 STAR64](pine64/star64/)                                        | `<nixos-hardware/pine64/star64>`                        |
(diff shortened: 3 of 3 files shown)


pine64/rock64/default.nix
```diff
@@ -0,0 +1,26 @@
+{
+  boot = {
+    initrd.kernelModules = [
+      "pcie_rockchip_host"
+      "phy_rockchip_pcie"
+      "sdhci_pci"
```
Mic92, line 6: Is this not detected by nixos-generate-config?

Mic92: Mhm. but I guess our installer wouldn't boot, so the normal install workflow wouldn't boot.

pine64/rock64/default.nix
```diff
@@ -0,0 +1,26 @@
...
+    ];
+    kernelModules = [
```
Mic92, line 8: Why is any of this needed? I don't remember needing to specify this.

flake.nix
```diff
@@ -271,6 +271,7 @@
         olimex-teres_i = import ./olimex/teres_i;
         pcengines-apu = import ./pcengines/apu;
         pine64-pinebook-pro = import ./pine64/pinebook-pro;
+        pine64-rock64 = import ./pine64/rock64;
```
Mic92, line 274: Please also add this to the README.

pine64/rock64/default.nix
matdibu: I remember some of them not being detected automatically, but I can remove that `boot.kernelModules` list entirely if it's out-of-scope for nixos-hardware

pine64/rock64/default.nix
matdibu: https://github.com/NixOS/nixos-hardware/blob/master/pine64/rockpro64/default.nix#L3


my local config (and, in turn, what was done in this PR) is inspired by how the RockPro64 is configured

matdibu: I'm currently using a half-functional build of u-boot/tow-boot, generated from this PR https://github.com/Tow-Boot/Tow-Boot/pull/322

flake.nix
matdibu: done

pine64/rock64/default.nix
Mic92: So it won't show any graphics if you attach a monitor, or what does this fix? Usually the kernel will match pci/usb vendor ids with drivers and load them as needed.

Mic92: I am fine with adding modules if they are actually not recognize by the hardware.

pine64/rock64/default.nix
Mic92: We can leave those modules in here if you also add a README how to bootstrap the rest of the configuration i.e. instructions to build a installer image.

pine64/rock64/default.nix
matdibu: makes sense, then I'll boot once without this list, and I'll compare which modules are loaded
might take a bit, I won't have access to the SBC for a few days

pine64/rock64/default.nix
matdibu: it's just the generic aarch64 uefi installer iso

I guess it'd be more common to go the u-boot+sd-card image, like it says on the wiki

which is unlike what I have -- tow-boot+normal EFI installer

I wouldn't really recommend people use a draft PR for their own stuff though

pine64/rock64/default.nix
matdibu: using this after booting with all modules listed
```shellsession
$ ssh rock64.lan -- lsmod | cut -d' ' -f1 | sort -u > before.txt
```
and this after removing them
```shellsession
$ ssh rock64.lan -- lsmod | cut -d' ' -f1 | sort -u > after.txt
```
I can get a diff
```shellsession
$ diff before.txt after.txt
```
that shows that these are the modules that are not normally loaded, unless added there by force
```
panfrost
pcie_rockchip_host
phy_rockchip_pcie
rk3399_dmc
rockchip_dfi
rockchip_isp1
rockchip_rga
rockchip_saradc
sdhci_pci
v4l2_async
v4l2_fwnode
videobuf2_dma_sg
videobuf2_vmalloc
```
that being said, it still boots and I still get HDMI video out

matdibu: `panfrost` definitely shouldn't be there, `lima` is the correct driver for the rock64's rk3328 Mali-450 "Utgard"
https://wiki.debian.org/PanfrostLima

matdibu: in conclusion: my machine boots without anything specified in `boot.initrd.kernelModules` and `boot.kernelModules`

matdibu:
@Mic92 since it looks like the rock64 works without any of the changes here, I'm open to simply closing this

pine64/rock64/default.nix
```diff
@@ -0,0 +1,26 @@
...
+      "rockchip_vdec"
+      "snd_soc_rockchip_i2s"
+      "rk_crypto"
+      "dwmac_rk"
+      "rk3399_dmc"
+      "v4l2_h264"
+      "v4l2_mem2mem"
+      "v4l2_vp9"
+    ];
+  };
```
Mic92, line 25: So rock64 just boots without any special configuration? We can also add an empty profile in this case.

matdibu:
this isn't really necessary

PR #1296, opened 2025-01-04
starfive visionfive2: allow uboot and opensbi source overrides

RossComputerGuy:
###### Description of changes

Adds an easy way to override the sources for uboot and OpenSBI, this makes it easier for someone to switch to a the vendor's u-boot and OpenSBI.

###### Things done

### DIFF
--- a/starfive/visionfive/v2/default.nix
+++ b/starfive/visionfive/v2/default.nix
@@ -5,6 +5,10 @@
   ...
 }:
 {
+  imports = [
+    ./firmware.nix
+  ];
+
   boot = {
     consoleLogLevel = lib.mkDefault 7;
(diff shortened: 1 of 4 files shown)


Mic92:
@mergify queue

RossSmyth:
How are you supposed to flash the device after this PR? The README and wiki both say to use 
https://github.com/NixOS/nixos-hardware/blob/3441b5242af7577230a78ffb03542add264179ab/starfive/visionfive/v2/README.md?plain=1#L71-L83

But after this PR for some reason you can only run that tool on RISC-V. Which it doesn't make sense to me to flash my RISC-V board from my RISC-V board.

PR #1308, opened 2025-01-07
Feature.asus zephyrus g16 gu605 my backlight and keyboard rules

relief-melone:
###### Things done

### DIFF
--- a/asus/zephyrus/shared/backlight.nix
+++ b/asus/zephyrus/shared/backlight.nix
@@ -1,0 +1,9 @@
+{ ... }:
+{
+  # These flags are used to enable backlight control when the dGPU is working in hybrid mode
+  boot.kernelParams = [
+    "i915.enable_dpcd_backlight=1"
+    "nvidia.NVreg_EnableBacklightHandler=0"
+    "nvidia.NVReg_RegistryDwords=EnableBrightnessControl=0"
+  ];
+}
--- a/asus/zephyrus/gu605my/default.nix
+++ b/asus/zephyrus/gu605my/default.nix
@@ -7,6 +7,7 @@
     ../../../common/gpu/nvidia/ada-lovelace
     ../../../common/pc/laptop
     ../../../common/pc/laptop/ssd
+    ../shared/backlight.nix
   ];
 
   hardware.nvidia = {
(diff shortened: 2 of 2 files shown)


relief-melone:
would also lite to test it out. would prefer to test it with an up do date nixpkgs commit though and the current one has issues with my printer as well as some nvim features.
would want to wait for the next publish here

Mic92:
@Moraxyc still on vacation?

Moraxyc:
I tried to use only dgpu without adding these kernel parameters, and the backlight control was normal. So the answer is that we should determine whether to add these kernel parameters based on whether only dgpu is working. However, I don't know how to switch to dgpu-only state in nxios configurations. I entered the bios to switch manually, and the following content seems to be no effective

```nix
    specialisation.dgpu-only.configuration = {
      disabledModules = [
        # Containing `boot.kernelParams = [];`
        ./brightness.nix
      ];
      system.nixos.tags = [ "dgpu-only" ];
      services.xserver.exportConfiguration = true;

      services.xserver = {
        videoDrivers = lib.mkForce [ "nvidia" ];
      };

      hardware = {
        nvidia = {
          powerManagement.enable = lib.mkForce false;
          modesetting.enable = true;
          prime = {
            offload.enable = lib.mkForce false;
            sync.enable = true;
          };
        };
      };
    };
```

Mic92:
@Moraxyc So should we merge this?

Moraxyc:
@relief-melone Please add a comment to indicate that these parameters are used to enable backlight control when the dGPU is working in hybrid mode.

After this, I think we can get the PR merged.

Mic92:
Please undraft when the change is done.

phanirithvij:
friendly ping @relief-melone

asus/zephyrus/shared/backlight.nix
```diff
@@ -0,0 +1,8 @@
+{ ... }:
+{
+  boot.kernelParams = [
```
Mic92, line 3:
```suggestion
  # These flags are used to enable backlight control when the dGPU is working in hybrid mode
  boot.kernelParams = [
```

Mic92:
@mergifyio queue

Mic92:
@mergifyio queue

Issue #1297, opened 2025-01-04
Lenovo IdeaCentre K330 unstable


Gabgobie:
Hi,

I am unsure if this is the correct place to raise this Issue but ever since trying to switch over my old PC to Nix, it has been freezing shortly after logging in. It has been running stable under RockyLinux 9.3.

I am assuming that this issue is caused by NixOS running Wayland by default, even though my hardware is rather old (i7 2600 + Nvidia GT 545).
During the installation process I went with the defaults except choosing my keyboard layout, timezone and the Plasma desktop.

I will look into switching to x11 for now and report back if that stabilizes the system. I don't know if it is feasible to change the default for this depending on the hardware but thought it be better to mention it rather than just fixing it for myself.

Best,
Gab

Mic92:
Please open a pull request if you have configuration that works. We do not have this hardware.

PR #1310, opened 2025-01-08
Lenovo ideacentre k330

Gabgobie:
###### Description of changes

This will add a default configuration for the Lenovo IdeaCentre K330

Since I don't really know what I am doing here, I'll need some input from more experienced Nix users. The main goal was to help out with an instability that came from wayland being a default even on old systems (for plasma desktops). #1297 

The contributing guidelines recommend to import
- CPU (Intel)
- GPU (NVIDIA)
- PC (in this case) | Laptop
- HDD | SSD

I imported CPU, GPU and PC but am unsure about the storage component since presumably most old systmes that somebody wants to revive will use a cheap SSD, including mine while the original hardware was almost certainly a HDD (in my case a Seagate Barracuda)

###### Things done

- [x] Tested the changes in your own NixOS Configuration (somewhat as in I tested if disabling wayland and running on x11 fixes the issue, which it did)

Best,
Gab

### DIFF
--- a/lenovo/ideacentre/k330/README.md
+++ b/lenovo/ideacentre/k330/README.md
@@ -1,0 +1,17 @@
+# Lenovo IdeaCentre K330
+
+The specific system I took for reference has the following hardware configuration:
+
+- Intel Core i7 2600
+- NVIDIA GeForce GT 545 [Latest supported (proprietary) driver (390.xx)](https://www.nvidia.com/en-us/drivers/details/196213/)
+- Some SSD (originally had a Seagate Barracuda hard drive)
+
+This hardware configuration was motivated by #1297
+
+I recommend enabling xserver instead of trying to use Wayland. As documented in the above linked issue, using Wayland with this rather old hardware lead to the system freezing after a short time of operation.
+
+```nix
+{
+    services.xserver.enable = true;
+}
+```
--- a/lenovo/ideacentre/k330/default.nix
+++ b/lenovo/ideacentre/k330/default.nix
@@ -1,0 +1,19 @@
+{ config, lib, ... }:
+{
+    imports = [
+        ../../../common/cpu/intel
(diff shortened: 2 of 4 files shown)


Mic92:
@mergify queue

Mic92:
@mergify dequeue

lenovo/ideacentre/k330/README.md
```diff
@@ -0,0 +1,9 @@
+# Lenovo IdeaCentre K330
```
Mic92, line 1: Please add this to flake.nix and the table in the README.

lenovo/ideacentre/k330/default.nix
```diff
@@ -0,0 +1,18 @@
+{ lib, ... }:
+{
+    imports = [
+        ../../../common/cpu/intel
+        ../../../common/gpu # Is it possible/advisable to pin this to the 390.xx driver family in case the user wants to use non-free drivers?
+        ../../../pc
+    ];
+
+    # On my machine Wayland causes the desktop to freeze after a short time of operation
+    services.xserver.enable = true;
```
Mic92, line 10: I think we should rather document this behavior i.e. a README rather than enabling xserver.

lenovo/ideacentre/k330/default.nix
```diff
@@ -0,0 +1,18 @@
...
+    services.displayManager.sddm.wayland.enable = false;
```
Mic92, line 11: That might be ok to keep.

lenovo/ideacentre/k330/default.nix
```diff
@@ -0,0 +1,18 @@
...
+
+    # Should this be a conditional default in case plasma is activated?
+    # What if somebody installs both plasma AND another DE?
+    # The goal is to prefer x11 over wayland due to compatibility issues with the old hardware
+
+    services.displayManager.defaultSession = lib.mkDefault "plasmax11";
```
Mic92, line 17: You can use lib.mkIf to check if kde is enabled.

Mic92:
@mergify queue

PR #1317, opened 2025-01-16
gpd/pocket-4: init

miangraham:
###### Description of changes

Here's a Pocket 4 first pass based on the Pocket 3 setup.

I'm testing this on the [Pocket 4 with AI 9 HX 370](https://gpdstore.net/gpd-mini-laptop/gpd-pocket-4/#cgkit-tab-additional_information) using NixOS 24.11 and mainline kernel 6.12. Apart from the display rotation needed in kernel params and WM settings like the Pocket 3, basic desktop stuff seems to just work.

#### Verified so far:

- [x] Audio
- [x] Bluetooth
- [x] Camera: Works. Tested in OBS
- [x] Display: Built-in screen is on and oriented properly in stage-1 before LUKS passphrase. Sway requires WM-side rotation, not sure about other GUI envs.
- [x] DisplayPort over USB-C: Works
- [x] HDMI: Works. Auto-detects on hotplug and comes back after suspend. On in time for passphrase.
- [x] Internet via RJ45
- [x] MicroSD reader module: Works.
- [x] Mouse: Built-in touchpad and click buttons work fine.
- [x] P-States enabled, EPP settings available
- [x] Suspend and resume (systemctl suspend with default BIOS)
- [x] Touchscreen: Works
- [x] USB ports: All working
- [x] Video: Hardware decoding in mpv and chrome
- [x] Wifi

#### Not working:

❌ Fingerprint Reader: [HOLTEK FocalTech Fingerprint Device](https://linux-hardware.org/?id=usb:2808-0752). No driver support yet. Places to watch: [1](https://fprint.freedesktop.org/supported-devices.html), [2](https://github.com/ftfpteams/focaltech-linux-fingerprint-driver)

#### Untested:

❓ Optional hardware modules (KVM, serial, LTE). I don't have any of these to try out. YMMV.

###### Things done

### DIFF
--- a/gpd/pocket-4/default.nix
+++ b/gpd/pocket-4/default.nix
@@ -1,0 +1,29 @@
+{ lib, pkgs, ... }:
+let inherit (lib) mkDefault;
+in
+{
+  imports = [
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+    ../../common/cpu/amd
+    ../../common/cpu/amd/pstate.nix
+    ../../common/gpu/amd
+    ../../common/hidpi.nix
+  ];
+
+  boot.kernelParams = [
+    # The GPD Pocket 4 uses a tablet LTPS display, that is mounted rotated 90° counter-clockwise
+    "fbcon=rotate:1" "video=eDP-1:panel_orientation=right_side_up"
+  ];
+
+  fonts.fontconfig = {
+    subpixel.rgba = "vbgr"; # Pixel order for rotated screen
+
+    # The display has √(2560² + 1600²) px / 8.8in ≃ 343 dpi
+    # Per the documentation, antialiasing, hinting, etc. have no visible effect at such high pixel densities anyhow.
+    hinting.enable = mkDefault false;
+  };
+
+  # More HiDPI settings
+  services.xserver.dpi = 343;
+}
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ See code for all available configurations.
 | [GPD MicroPC](gpd/micropc)                                                        | `<nixos-hardware/gpd/micropc>`                          |
 | [GPD P2 Max](gpd/p2-max)                                                          | `<nixos-hardware/gpd/p2-max>`                           |
 | [GPD Pocket 3](gpd/pocket-3)                                                      | `<nixos-hardware/gpd/pocket-3>`                         |
+| [GPD Pocket 4](gpd/pocket-4)                                                      | `<nixos-hardware/gpd/pocket-4>`                         |
 | [GPD WIN 2](gpd/win-2)                                                            | `<nixos-hardware/gpd/win-2>`                            |
 | [GPD WIN Max 2 2023](gpd/win-max-2/2023)                                          | `<nixos-hardware/gpd/win-max-2/2023>`                   |
 | [GPD WIN Mini 2024](gpd/win-mini/2024)                                            | `<nixos-hardware/gpd/win-mini/2024>`                    |
--- a/flake.nix
+++ b/flake.nix
@@ -121,6 +121,7 @@
         gpd-micropc = import ./gpd/micropc;
         gpd-p2-max = import ./gpd/p2-max;
         gpd-pocket-3 = import ./gpd/pocket-3;
+        gpd-pocket-4 = import ./gpd/pocket-4;
         gpd-win-2 = import ./gpd/win-2;
         gpd-win-max-2-2023 = import ./gpd/win-max-2/2023;
         gpd-win-mini-2024 = import ./gpd/win-mini/2024;


miangraham:
Should be good to go! System's working great.

gpd/pocket-4/default.nix
```diff
@@ -0,0 +1,33 @@
+{ lib, pkgs, ... }:
+let inherit (lib) mkDefault;
+in
+{
+  imports = [
+    ../../common/pc/laptop
+    ../../common/pc/laptop/ssd
+    ../../common/hidpi.nix
+    ../../common/gpu/24.05-compat.nix
```
LucasFA, line 9: The 24.05 compatibility module should probably not be imported, not only is `24.05` EOL already and the file soon(ish?) to be removed, but it just adds an alias for an option. If you're seeing some error or warning regarding this, it might be because you're using `hardware.opengl` instead of `hardware.graphics`, the new naming from 24.11 onwards.

Also, it can be imported automatically from `../../common/gpu/amd/default.nix`

gpd/pocket-4/default.nix
```diff
@@ -0,0 +1,33 @@
...
+  ];
+
+  # Necessary kernel modules
+  boot.initrd.availableKernelModules = [ "nvme" "xhci_pci" "usbhid" "thunderbolt" ];
```
LucasFA, line 13: Are these not picked up on installation and written to the `hardware-configuration.nix` file? You can test this in an already running system with `sudo nixos-generate-config --show-hardware-config`

As far as I have seen in other configs, these are usually not added to the profile; the user is expected to import `hardware-configuration.nix`. I delegate judgement to more experienced users.

It could be worth it to add the kernel modules regarding the hardware modules, though. Make them mostly plug and play, even if they are usually unnecessary.

gpd/pocket-4/default.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  # For stage-1 screen resolution and external display output
+  boot.initrd.kernelModules = [ "amdgpu" ];
```
LucasFA, line 16: Alternatively written, `hardware.amdgpu.initrd.enable = true`. This line already appears in `common/gpu/amd/default.nix`. If that module is, in the end, imported, this can be removed.

gpd/pocket-4/default.nix
```diff
@@ -0,0 +1,33 @@
+{ lib, pkgs, ... }:
+let inherit (lib) mkDefault;
+in
+{
+  imports = [
```
LucasFA, line 5: Could `../../common/cpu/amd` and `../../common/gpu/amd` be imported?

`cpu/amd/pstate.nix` also seems relevant, but I am unfamiliar with AMD and `pstate`

LucasFA (commented):
I don't have the hardware to test this, but seems like an interesting computer. My suggestions are based purely on the specs.

gpd/pocket-4/default.nix
miangraham: Worked from the [Pocket 3](https://github.com/NixOS/nixos-hardware/blob/master/gpd/pocket-3/default.nix#L9) conf as an example. Should this be removed there too?

gpd/pocket-4/default.nix
miangraham: Yes these appear in my hardware-configuration.nix and I was already getting them. Included them based on the [Pocket 3 conf](https://github.com/NixOS/nixos-hardware/blob/master/gpd/pocket-3/default.nix#L13) doing the same, assuming some people weren't already importing them. Should they be removed there too?

gpd/pocket-4/default.nix
miangraham: Also written to echo the same done for [Pocket 3](https://github.com/NixOS/nixos-hardware/blob/master/gpd/pocket-3/default.nix#L16). Should I open a PR to change that? I don't have hardware to test.

gpd/pocket-4/default.nix
miangraham: I haven't noticed problems indicating I need those. What should I test for?

miangraham: Checking:
common/cpu/amd is entirely included in the generated hardware-configuration.nix. Doesn't seem necessary.
common/gpu/amd looks like a nice clean way to pull in the initrd setting.

miangraham: I'll try some pstate testing and see if I can figure out what's what. Thanks for all the feedback!

gpd/pocket-4/default.nix
LucasFA: Ah, I see this was above my paygrade. It was originally added [here](https://github.com/NixOS/nixos-hardware/commit/584a5e551885382c1cf9b09a990fbcd2ccbc992f) as a minimal change to the profile to make it work under 24.05. Several other packages have the same line for the same reason.

I would argue that 24.05 support can be dropped anytime and there is no reason to start a profile supporting 24.05 now, so at least delete the line here.
Regarding Pocket 3, maybe we can let that be for now and reserve that for whenever support for 24.05 is definitely dropped. Deleting it there could break configs for still unupgraded users.

gpd/pocket-4/default.nix
LucasFA: In a similar vein to dropping support for 24.05, I don't think there is a need to init a profile with the kernel modules, but changing an already existing profile is a breaking change, and I'm not comfortable suggesting breaking changes.
(Though I am confident that NixOS users are savvy enough to deal with this, I wouldn't like to be on the receiving end of a change that prevents me rebuilding my daily driver or my server)

gpd/pocket-4/default.nix
LucasFA: >common/cpu/amd is entirely included in the generated hardware-configuration.nix. Doesn't seem necessary.

That's fair. I think these minimal imports are a taste thing more than anything. The only situation where they might make a difference is in the case that the `common/cpu/amd` profile has a change in the future.
> common/gpu/amd looks like a nice clean way to pull in the initrd setting.

Nice!
>What should I test for?

Honestly? I don't really know. For starters, check the module loads with `journalctl -b 0` and grep around for `amd-pstate`, which [seems to be the name of the module](https://www.kernel.org/doc/html/latest/admin-guide/pm/amd-pstate.html)

LucasFA: >I haven't noticed problems indicating I need those.

Ah, no, there's definitely no real 'need' in most cases, less still with AMD which doesn't have issues with Linux. A mild battery duration improvement from `pstate`, perhaps.

gpd/pocket-4/default.nix
miangraham: Sounds reasonable. Will clean up and test. Thanks

gpd/pocket-4/default.nix
LucasFA: Well, it IS an inmediately reviewable and easily merged change, so low-hanging PR ;) I don't know if after thinking a little bit through Pocket 3 there are some other changes that can be done to clean it up without breaking configs. I hope to come back to this this weekend

gpd/pocket-4/default.nix
miangraham: common/cpu/amd seems like reasonable future-proofing on second consideration. Will pull that in as well.

miangraham:
P-States flipped on and briefly tested. I don't have the cycles to do extended power investigation, but they're enabled and not hurting anything. Quick results with otherwise default settings:

```console
$ dmesg | grep "Kernel command line"
[    0.041247] Kernel command line: initrd=\EFI\nixos\g219bvi25xmzhv2qjx5wxa6n2w0342hw-initrd-linux-6.12.9-initrd.efi init=/nix/store/mhwmka1j2wrr8cwvd5nic6qma20zl20c-nixos-system-yuuri-24.11.20250115.e24b4c0/init amd_pstate=active fbcon=rotate:1 video=eDP-1:panel_orientation=right_side_up loglevel=4

$ cat /sys/firmware/acpi/platform_profile_choices
low-power balanced performance

$ cat /sys/firmware/acpi/platform_profile
balanced

$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors
performance powersave

$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
powersave

$ cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
default performance balance_performance balance_power power

$ cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
balance_power
```

Not sure why there's no schedutil governor. Kernel support maybe not there yet? `powersave` geekbenches at 4% below `performance` with 0.2W idle battery draw savings, which seems perfectly reasonable as a default so no governor changes made here.

gpd/pocket-4/default.nix
miangraham: All gone now and just leaning on common/gpu/amd!

Mic92:
@mergify queue

Mic92:
@mergify dequeue

PR #1320, opened 2025-01-17
Modify apple/t2/default.nix to use configured efiSysMountPoint

3ulalia:
###### Description of changes
The configured option to use a custom bootx64.efi has a hard-coded path, which doesn't work for users (like myself) who have mounted their ESP in a different place. This PR uses the configured `config.boot.loader.efi.efiSysMountPoint` as the root dir to look for the EFI file.

###### Things done

      importing it via `<nixos-hardware>` or Flake input <- will do once I make this pr :p

### DIFF
--- a/apple/t2/default.nix
+++ b/apple/t2/default.nix
@@ -23,6 +23,7 @@ let
     });
 
   pipewirePackage = overrideAudioFiles pkgs.pipewire "spa/plugins/";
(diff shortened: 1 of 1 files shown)


Mic92:
So does the change work now?

Mic92:
Unsure if this is needed or not. Please just re-open if you think we still need this.

PR #1322, opened 2025-01-19
xiaomi/redmibook/16-pro-2024: init

Weathercold:
###### Description of changes

~~Also exposed the boot fix as a module since other models might be affected as well~~

###### Things done

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,3 +357,4 @@ See code for all available configurations.
 | [TUXEDO InfinityBook Pro 14 - Gen7](tuxedo/infinitybook/pro14/gen7)               | `<nixos-hardware/tuxedo/infinitybook/pro14/gen7>`       |
 | [TUXEDO Pulse 14 - Gen3](tuxedo/pulse/14/gen3)                                    | `<nixos-hardware/tuxedo/pulse/14/gen3>`                 |
 | [TUXEDO Pulse 15 - Gen2](tuxedo/pulse/15/gen2)                                    | `<nixos-hardware/tuxedo/pulse/15/gen2>`                 |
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -357,3 +357,5 @@ See code for all available configurations.
 | [TUXEDO InfinityBook Pro 14 - Gen7](tuxedo/infinitybook/pro14/gen7)               | `<nixos-hardware/tuxedo/infinitybook/pro14/gen7>`       |
 | [TUXEDO Pulse 14 - Gen3](tuxedo/pulse/14/gen3)                                    | `<nixos-hardware/tuxedo/pulse/14/gen3>`                 |
 | [TUXEDO Pulse 15 - Gen2](tuxedo/pulse/15/gen2)                                    | `<nixos-hardware/tuxedo/pulse/15/gen2>`                 |
+| [Xiaomi (boot fix)](xiaomi/common/fix-boot)                                       | `<nixos-hardware/xiaomi/common/fix-boot>`               |
```
Mic92, line 360: It's a bit to vague for me what "some xiaomi" laptops mean.
I would say if we have more models, we just include the module as well.
 
```suggestion
```

flake.nix
```diff
@@ -308,6 +308,8 @@
         tuxedo-infinitybook-pro14-gen7 = import ./tuxedo/infinitybook/pro14/gen7;
         tuxedo-pulse-14-gen3 = import ./tuxedo/pulse/14/gen3;
         tuxedo-pulse-15-gen2 = import ./tuxedo/pulse/15/gen2;
+        xiaomi-fix-boot = import ./xiaomi/common/fix-boot;
```
Mic92, line 311: For I would encourage users to include their own laptop module rather than using this:

```suggestion
```

README.md
Mic92: This workaround might also not be needed in the next hardware generation.

PR #1326, opened 2025-01-20
Dell E7240 -> Dell Latitude E7240

nmouha:
###### Description of changes

The E7240 is part of the Dell Latitude series. This PR moves it to the right location and corrects the corresponding files.

###### Things done


dell/e7240/default.nix
```diff
@@ -1,6 +0,0 @@
-{
-  imports = [
```
Mic92, line 2: Can we add a deprecation warning here and import the new module here?
This is a bit nicer for existing users.

Check: `common/pc/hdd/default.nix` for an example for deprecation.

Mic92:
@mergify queue

Mic92:
@mergify queue

PR #1329, opened 2025-01-20
radxa/rock-4c-plus: init

RadxaYuntian:
###### Description of changes

* Add currently undocumented `rockchip` module, currently used internally by `radxa` module
  Open to be shared with other consumers if there is a need. Documentation can follow afterwards when we see common patterns.
* Add support for ROCK 4C+

###### Things done


radxa/rock-4c-plus/default.nix
```diff
@@ -0,0 +1,24 @@
...
+    ../.
+    ../../rockchip/rk3399
+  ];
+
+  hardware = {
+    radxa.enable = true;
+    rockchip = {
+      enable = true;
+      diskoExtraPostVM = ''
+        dd conv=notrunc,fsync if=${pkgs.ubootRock4CPlus}/idbloader.img of=$out/${cfg.diskoImageName} bs=512 seek=64
```
Mic92, line 19: Interesting approach, we can potentially have this also for other rockchip based boards.

RadxaYuntian: Yes and this is just a proof-of-concept. Once this PR is accepted I'll add additional devices from us.

RadxaYuntian:
BTW if this and EFI boot chain was accepted, then we will need to worry about updating U-Boot for like RK3588 devices where upstream is moving fast with potential DTB changes. Such updating script can also be used in `diskoExtraPostVM` to reduce duplication.

The question is where should this script go? Main `nixpkgs` (since it is quite essential for the U-Boot packages there to be useful) or here?

RadxaYuntian:
Need to rework tests now since we will have a lot more Arm devices coming: https://github.com/NixOS/nixos-hardware/blob/master/tests/flake.nix#L42-L48

I tried to run the test locally but it locked my computer up :/

RadxaYuntian:
Just added `rock-4c-plus` to `aarch64Systems` for now. We can figure out how to refactor tests in another PR.

RadxaYuntian:
Move unfree config to `disko.nix` so it is not loaded by default. This allow test to run.

RadxaYuntian:
@msgilligan you might be interested.

RadxaYuntian:
@kibiz0r reported that the original flake example does not build on system without binfmt enabled, which I assume is due to the U-Boot package (does Hydra not cache unfree package?). I also thought about what's happening next, and realized that I want an option to override the U-Boot package, so I can use packages from my nixpkgs fork while working on them.

So such option is added, and an example of overriding with cross built package is provided in the example flake. It is disabled by default since it requires knowledge of implementation detail to use, so normal users are recommended to configure `binfmt` instead.

radxa/default.nix
```diff
@@ -0,0 +1,22 @@
...
+in {
+  options.hardware.radxa = {
+    enable = lib.mkEnableOption "Radxa system support";
+  };
+
+  config = lib.mkIf cfg.enable {
+    boot = {
+      # Currently enable bcachefs automatically set
+      # kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
+      kernelPackages = lib.mkOverride 990 pkgs.linuxPackages_latest;
```
Mic92, line 17:
```suggestion
      kernelPackages = lib.mkIf (lib.versionOlder pkgs.linux.version "6.12") pkgs.linuxPackages_latest;
```

RadxaYuntian: I think `kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;` is still better. The intention is to always use the latest kernel so newer features can be available sooner (like RK3588 display in 6.13), and `bcachefs` is enabled because we are using the latest kernel.

With this on newer nixpkgs the default kernel will be 6.12 until next LTS.

RadxaYuntian:
CI failed because NixOS/nixpkgs#361461 is not merged :/

radxa/default.nix
Mic92, line 17:
```suggestion
      # TODO: Consider removing this line once, we get an LTS kernel that is newer than 6.12
      kernelPackages = lib.mkOverride 990 pkgs.linuxPackages_latest;
```

Mic92:
@mergify queue

PR #1334, opened 2025-01-22
dell/latitude/7420: init

alex-robbins:
This adds an entry for `dell-latitude-7420`. Before using this change with my 7420, VA-API wasn't working, so I had no hardware accelerated video decoding; now VA-API is working.

I'm not specially setting any kernel parameters like some other entries do, because I don't think it's necessary in this case. Our Tiger Lake module already sets `i915.enable_guc=3`, and other i915 settings can be left at the default (e.g. upstream Linux defaults `i915.enable_fbc` to `-1`, meaning that FBC should be enabled for chips for which that is supported). 

###### Things done

I did not run `nix run ./tests#run .` as requested by `CONTRIBUTING.md`, because when I tried, my load average went to 60 until my OOM killer was activated. Nonetheless, this is a pretty straightforward change, so it should be fine.


Mic92:
@mergify queue

alex-robbins:
🎉

Thanks.

PR #1394, opened 2025-02-24
Simplify implementation of B550 suspend fix

marienz:
@NireBryce (contributor of the original change): any chance you can test this on your B550, or at least compare the IDs in lspci and /proc/acpi/wakeup)? It *should* do the same thing (you should see GPP0 and GPP8 "disabled" in /proc/acpi/wakeup after rebooting with this applied, and suspend/wakeup should work) but in a way not prone to accidentally togging rather than disabling wakeups.

I've had it in my local config for a while (didn't realize nixos-hardware carried motherboard-specific fixes like this one), and I'm pretty sure I'm working around the same bug, but I can't be 100% sure the device IDs on your system match mine (would surprise me if they didn't, but given the BIOS seems to be buggy to begin with I'd prefer to verify that assumption...)

###### Description of changes

Instead of using systemd oneshot services that have to be careful to not toggle wakeups back on, use a udev rule to disable wakeups by device ID.

On a B550 Vision D Rev 1.0 with BIOS F17, these do almost the same thing:

````
$ lspci -n | grep 1022:1483
00:01.1 0604: 1022:1483
00:01.2 0604: 1022:1483
00:03.1 0604: 1022:1483

$ cat /proc/acpi/wakeup
Device  S-state   Status   Sysfs node
...
GPP0      S4    *disabled  pci:0000:00:01.1
GPP8      S4    *disabled  pci:0000:00:03.1
````

Two of the three devices with the PCI vendor/device ID specified by the
udev rule correspond to devices previously disabled via ACPI (if I
understand correctly disabling these via either /proc/acpi/wakeup or
udev device attribute has the same effect).

The third device is (like the other two) using the "pcieport" driver.
Using a device connected via that port as a wakeup device still works.

###### Things done

### DIFF
--- a/gigabyte/b550/README.md
+++ b/gigabyte/b550/README.md
@@ -9,9 +9,10 @@ Symptoms:
 - It goes into suspend, then seems to boot and hang.  Sometimes it suspends successfully, but waking it from suspend puts it in the "zombie" state.
 - By playing chicken with volatile storage and flicking the power switch on the back of power supply, you can sometimes get it to wake from suspend as the card un-powers before volatile storage does.
 
-Fix: disable GPP0 and GPP8 (And, for some cards, potentially PTXH, I can't test) in /proc/acpi/wakeup
-    - To do this permanently, a systemd service is provided
+This can be fixed by disabling GPP0 and GPP8 (And, for some cards, potentially PTXH, I can't test) in /proc/acpi/wakeup.
 
+But because /proc/acpi/wakeup only supports toggling (not disabling) wakeups, we use a udev rule to disable wakeups for
+the same device by PCI device ID instead.
 
 ## This affects at least:
 - Gigabyte b550m-d3sh (my machine)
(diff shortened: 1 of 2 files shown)


NireBryce:
oh, yes.  I haven't tested it yet but I'll try to tomorrow.

It's also not in the nixos-hardware flake, I had forgotten about this project when I had to move house suddenly and never followed up on that.

marienz:
> It's also not in the nixos-hardware flake

...oh, right, #884 didn't wire it up in flake.nix. But looks like someone else fixed that, see #978.

NireBryce:
yep, appears to work on my b550m-d3sh! It's also much cleaner looking.

The README.md mentions the systemd script if you want to change it to udev rule, otherwise I'll try to remember tomorrow.

marienz:
> The README.md mentions the systemd script

Ah thanks, I would've missed that.

> For the record:

Ok, that matches what I have: one PCIe device without corresponding /proc/acpi/wakeup entry, no harmful side-effects observed from disabling wakeups for it.

Thanks for testing! I'll finish this up (README, use my nixos-hardware fork from my own flake) and submit this for review once I have a spare moment.

NireBryce:
thanks!

PR #1398, opened 2025-02-27
Add Dell Precision 5520 (based on 5530)

cjdell:
###### Description of changes
Added the definition for Dell Precision 5520 using the already existing 5530 as a template.

Confirmed nVidia hybrid graphics are working.

###### Things done


flake.nix
```diff
@@ -81,6 +81,7 @@
         dell-poweredge-r7515 = import ./dell/poweredge/r7515;
         dell-precision-3541 = import ./dell/precision/3541;
         dell-precision-5490 = import ./dell/precision/5490;
+        dell-precision-5520 = import ./dell/precision/5520;
```
Mic92, line 84: Please also add this model to the README.md table. Otherwise our CI doesn't test this.

dell/precision/5520/default.nix
```diff
@@ -0,0 +1,47 @@
...
+      "nvidia_drm.modeset=1"
+    ];
+  };
+
+  hardware = {
+    nvidia = {
+      open = lib.mkDefault false;
+      nvidiaSettings = lib.mkDefault true;
+      modesetting.enable = lib.mkDefault true;
+      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.stable;
```
Mic92, line 35: Is this not the default?

dell/precision/5520/default.nix
```diff
@@ -0,0 +1,47 @@
...
+      # fix flicker
+      # source https://wiki.archlinux.org/index.php/Intel_graphics#Screen_flickering
+      "i915.enable_psr=0"
+      "nvidia_drm.modeset=1"
+    ];
+  };
+
+  hardware = {
+    nvidia = {
+      open = lib.mkDefault false;
```
Mic92, line 32: Afaik we do set this in nvidia generation specific profiles i.e. `../../../common/gpu/nvidia/maxwell`

dell/precision/5520/default.nix
```diff
@@ -0,0 +1,47 @@
...
+      nvidiaSettings = lib.mkDefault true;
```
Mic92, line 33: Should be already the default: https://search.nixos.org/options?channel=24.11&show=hardware.nvidia.nvidiaSettings&from=0&size=50&sort=relevance&type=packages&query=nvidiaSettings

dell/precision/5520/default.nix
```diff
@@ -0,0 +1,47 @@
...
+      modesetting.enable = lib.mkDefault true;
+      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.stable;
+      prime = {
+        intelBusId = lib.mkDefault "PCI:0:2:0";
+        nvidiaBusId = lib.mkDefault "PCI:1:0:0";
+      };
+    };
+  };
+  # This will save you money and possibly your life!
```
Mic92, line 42: Comment is in the wrong line.

cjdell:
@Mic92 Thank you for the review. I have blatantly stolen the config from the 5530 which is almost identical. In fact only lines 7 and 9 are different:

https://github.com/cjdell/nixos-hardware/blob/master/dell/precision/5530/default.nix
https://github.com/cjdell/nixos-hardware/blob/master/dell/precision/5520/default.nix

I may be out of my depth here but if you think there are improvements to be made then that would appear to be true of the 5530 model config as well which was merged 3 months ago. I deliberately made as few changes as possible.

Mic92:
I am dropping those defaults now also in the other module: https://github.com/NixOS/nixos-hardware/pull/1456

cphalpin:
@cjdell do you want to finish this or do you mind if I do?

PR #1413, opened 2025-03-13
feat: add Lenovo ThinkPad P16s Intel Gen 2

yajo:
@moduon MT-9339

###### Things done


lenovo/thinkpad/p16s/intel/default.nix
```diff
@@ -0,0 +1,8 @@
+{modulesPath, lib, ...}: {
+  imports = [
+    (modulesPath + "/installer/scan/not-detected.nix")
```
Mic92, line 3: nixos-hardware depends on nixos-generate-config. We try to not replace parts that can be auto-generated.

```suggestion
```

lenovo/thinkpad/p16s/intel/gen2/default.nix
```diff
@@ -0,0 +1,28 @@
...
+
+  boot.initrd.availableKernelModules = [
+    "nvme"
+    "sd_mod"
+    "thunderbolt"
+    "usb_storage"
+    "usbhid"
+    "xhci_pci"
+  ];
+  boot.kernelModules = ["kvm-intel"];
```
Mic92, line 18: Same as above:

```suggestion
```

yajo: However, nixos-generate-config only gets this:

```nix
  boot.initrd.availableKernelModules = [ "xhci_pci" "thunderbolt" "nvme" ];
```

Should I just remove those?

Mic92: Yes, but please provide a comment why those have been added manually.

yajo: Self-answer. TIL about `boot.initrd.includeDefaultModules`, which [adds most of those](https://github.com/NixOS/nixpkgs/blob/47ef6f56d4e473a814c546948f5c2c99f4d522e2/nixos/modules/system/boot/kernel.nix#L283-L342). It's enabled by default. Thus, none of these is actually needed indeed. Changed.

yajo:
All attended.

PR #1416, opened 2025-03-14
lenovo/thinkpad/x13-yoga: add thunderbolt support

horscchtey:
###### Description of changes

During setting up my Lenovo Thinkpad X13 Yoga with a Thinkpad Thunderbolt Dock, I discovered that my device failed to communicate with the dock, especially for the ethernet connection. By adding the ``services.hardware.bolt.enable = true`` configuration, my laptop now successfully connects to the ethernet connected to the Thunderbolt dock.

Happy to discuss this suggested change in more detail.

###### Things done

### DIFF
--- a/lenovo/thinkpad/x13/yoga/default.nix
+++ b/lenovo/thinkpad/x13/yoga/default.nix
@@ -5,4 +5,6 @@
   ];
 
   services.xserver.wacom.enable = lib.mkDefault config.services.xserver.enable;
+  
+  services.hardware.bolt.enable = lib.mkDefault true;
 }


lenovo/thinkpad/x13/yoga/default.nix
```diff
@@ -5,4 +5,6 @@
   ];
 
   services.xserver.wacom.enable = lib.mkDefault config.services.xserver.enable;
+  
+  services.hardware.bolt.enable = lib.mkDefault true;
```
Mic92, line 9: Is this useful outside of gnome?

```suggestion
  services.hardware.bolt.enable = lib.mkIf config.services.xserver.desktopManager.gnome.enable (lib.mkDefault true);
```

horscchtey: Enabling the `bolt` service brings up `boltctl`, which can be used without gnome.
Therefore, to answer your question, yes: I use it inside of `sway` without any gnome environment.

PR #1427, opened 2025-03-21
framework: fix TRRS headphones modprobe

leana8959:
###### Description of changes
It seems like the modprobe configuration is not guareded with the right condition -- matching with `pkgs.linux.version` doesn't reflect the kernel the user might have opted into.

With this change, `boot.kernelPackages = pkgs.linuxPackages_6_1;` will correctly set the modprobe configuration no matter what the default kernel nixpkgs ships.

Note: I have changed this option for both 13 and 16 inch models, however I have only tested it for the 13 inch model. 

###### Things done

### DIFF
--- a/framework/13-inch/common/default.nix
+++ b/framework/13-inch/common/default.nix
@@ -1,4 +1,5 @@
-{ lib, pkgs, ... }: {
+{ lib, config, ... }:
+{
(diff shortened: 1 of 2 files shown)


Mic92:
@mergify queue

PR #1449, opened 2025-04-19
Add yt6801 driver for TUXEDO InfinityBook Pro 14 - Gen9 - INTEL

j0hax:
###### Description of changes

This commit adds the Motorcomm YT6801 module for the InfinityBook Pro 14 Gen 9. Because both the Intel and AMD variants of the laptop have this Ethernet controller, the existing Nix module for the Intel variant was moved to its own directory.

###### Things done


flake.nix
```diff
@@ -330,7 +330,7 @@
         tuxedo-aura-15-gen1 = import ./tuxedo/aura/15/gen1;
         tuxedo-infinitybook-v4 = import ./tuxedo/infinitybook/v4;
         tuxedo-infinitybook-pro14-gen7 = import ./tuxedo/infinitybook/pro14/gen7;
-        tuxedo-infinitybook-pro14-gen9-intel = import ./tuxedo/infinitybook/pro14/gen9-intel;
```
Mic92, line 333: If we want to rename this, we need to leave a deprecation warning in the old module because non-flake users will others get an import error.

Here is an example: 

https://github.com/NixOS/nixos-hardware/blob/master/common/pc/hdd/default.nix

j0hax: Would using `lib.mkRenamedOptionModule` in the old module be a better option?

Mic92: Not sure this applies here. We want to deprecate a file rather than a NixOS option.

j0hax: Okay. I added a deprecation warning and import to the new module to ensure that nothing breaks for the time being 😄

Mic92:
@mergifyio queue

PR #1456, opened 2025-04-29
dell-precision-5530: remove unnecessary default nvidia options

Mic92:
In the nvidia module, we already disable the open source driver and we default to the stable nvidia package.

###### Things done


Mic92:
cc @symphorien

symphorien:
i'm on the 5515 which does not have a nvidia card

PR #1458, opened 2025-04-29
omen/*: check kernel version through `config` instead of `pkgs`

sebaszv:
###### Description of changes

The previous implementation was checking the kernel version through
`pkgs.linux`, which is only representative of the final system if
`boot.kernelPackages` is left as the default value of
`pkgs.linuxPackages`.

You can of course change this to other package sets, such as
`pkgs.linuxPackages_latest`. Instead, we now reference the kernel
through `config.boot.kernelPackages.kernel`.

###### Things done

- [x] Tested evaluation of modules with `nix run ./tests#run .` to succeed.

### DIFF
--- a/omen/14-fb0798ng/default.nix
+++ b/omen/14-fb0798ng/default.nix
@@ -1,4 +1,4 @@
-{ lib, pkgs, ... }:
+{ config, lib, ... }:
 
 {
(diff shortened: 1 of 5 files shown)


Mic92:
@mergify queue

PR #1462, opened 2025-04-29
gmktec/nucbox/g3-plus: init

sebaszv:
###### Description of changes

Product page:
<https://www.gmktec.com/products/nucbox-g3-plus-enhanced-performance-mini-pc-with-intel-n150-processor>

This profile just configures the Intel Twin Lake N150 CPU and integrated graphics for this mini-PC. fstrim is also enabled for the SSD. That's all this seemed to need to function properly. As is now expected from Intel NUC systems, it provides a solid "out-of-the-box" experience. No special quirks are apparent.

We import the Alder Lake modules since Twin Lake is just a refreshed version of the Alder Lake-N series. Re-using those seems to be fine for this purpose.

###### Things done

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -128,6 +128,7 @@
         focus-m2-gen1 = import ./focus/m2/gen1;
         gigabyte-b550 = import ./gigabyte/b550;
         gigabyte-b650 = import ./gigabyte/b650;
+        gmktec-nucbox-g3-plus = import ./gmktec/nucbox/g3-plus;
         google-pixelbook = import ./google/pixelbook;
         gpd-micropc = import ./gpd/micropc;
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -128,6 +128,7 @@
         focus-m2-gen1 = import ./focus/m2/gen1;
         gigabyte-b550 = import ./gigabyte/b550;
         gigabyte-b650 = import ./gigabyte/b650;
+        gmktec-nucbox-g3-plus = import ./gmktec/nucbox/g3-plus;
```
Mic92, line 131: Please also add this to the table in the README.

sebaszv: Done.

Mic92:
@mergify queue

Issue #1465, opened 2025-04-30
Consider not enabling TLP by default


jolars:
Consider dropping TLP by default for laptops. Modern laptops usually have firmware that enables power profiles, so it often no longer relevant. Other times the distro has its own way of handling power profiles, like gnome, which is already handled here:

https://github.com/NixOS/nixos-hardware/blob/72081c9fbbef63765ae82bff9727ea79cc86bd5b/common/pc/laptop/default.nix#L9-L10

I am not entirely sure what the scope of nixos-hardware is, but to me TLP seems like an opinionated choice that is not objectively good for every laptop. Searching around the issues on TLP and elsewhere, it is clear that is not always working well or provides a wanted benefit for the user. 

I think it would be a better idea to enable TLP discriminatively for select hardware where it is verified to be working well, rather than to enable it across the board.

Mic92:
At least if you use `config.services.power-profiles-daemon.enable` it's not used which is the case for gnome/kde etc.

Mic92:
But potentially we can invert the default and switch to power-profiles-daemon by default?

jolars:
On Wed Apr 30, 2025 at 9:15 AM CEST, Jörg Thalheim wrote:
> Mic92 left a comment (NixOS/nixos-hardware#1465)
>
> At least if you use `config.services.power-profiles-daemon.enable` it's not used which is the case for gnome/kde etc.

Yes, and it's also easy to disable by just using `programs.tlp.enable = false`, but I think it shouldn't be enabled by default.

TLP is pretty aggressive and I don't need it on my Thinkpad t14 gen 5. After starting to use nixos-hardware, it
took me quite a bit of time before I realized that TLP had been enabled and was causing my laptop to
slow down. I'm sure that many users want this kind of power management, but I thought that nixos-hardware
was more of a set of hardware configurations that (more or less) everyone could agree on, and I'm not
really sure TLP is that.

Mic92:
I think the choice to use tlp in nixos-hardware is also older than power-profiles-daemon. At the time this might have been the right solution for the hardware back than.

ciferkey:
To add another data point, tlp will be enabled by default on the Framework 13, and Framework [explicitly recommends](https://knowledgebase.frame.work/en_us/optimizing-ubuntu-battery-life-Sye_48Lg3) not using tlp.

PR #1474, opened 2025-05-11
laptop: replace tlp with power-profiles-daemon

Mic92:
TLP is often no longer the best choice since firmware provides power profiles instead. That's why we are removing TLP as the default option. We may want to bring back TLP to some old hardware if it makes sense i.e. certain thinkpad modules.

Fixes https://github.com/NixOS/nixos-hardware/issues/1465

###### Things done


lenovo/ideapad/15arh05/default.nix
```diff
@@ -17,9 +17,4 @@
 
   # Cooling management
   services.thermald.enable = lib.mkDefault true;
-
-  # tlp defaults to "powersave", which doesn't exist on this laptop
-  services.tlp.settings = {
```
Mic92, line 22: cc @Guekka this removes your tlp configuration. This laptop is from 2020 and hopefully also works well enough with power-profiles-daemon.

Guekka: I guess this should work. Let's go with this and we will revert if I notice an issue, I'm still using this laptop

JohnAZoidberg:
Actually, ppd also has been superceded by tuned in many distros.
See Fedora: https://fedoraproject.org/wiki/Changes/TunedAsTheDefaultPowerProfileManagementDaemon

tuned is fully compatible with ppd profiles, so it's a drop-in compatible but also a superset.

Mic92:
Happy to switch to tuned once it lands in nixpkgs.

common/pc/laptop/default.nix
```diff
@@ -1,11 +1,5 @@
...
-{
+{ lib, config, ... }: {
   imports = [ ../. ];
 
-  # Gnome 40 introduced a new way of managing power, without tlp.
-  # However, these 2 services clash when enabled simultaneously.
-  # https://github.com/NixOS/nixos-hardware/issues/260
-  services.tlp.enable = lib.mkDefault ((lib.versionOlder (lib.versions.majorMinor lib.version) "21.05")
-                                       || !config.services.power-profiles-daemon.enable);
+  config.services.power-profiles-daemon.enable = lib.mkDefault config.services.tlp.enable;
```
bubuntux: should this be enabled when config.services.tlp.enable is false? (missing `!`)
meaning we need 
`config.services.power-profiles-daemon.enable = lib.mkDefault !config.services.tlp.enable;`
instead

Mic92: Yes. Thanks.

common/pc/laptop/default.nix
```diff
@@ -1,13 +1,6 @@
...
   imports = [ ../. ];
 
-  # Gnome 40 introduced a new way of managing power, without tlp.
-  # However, these 2 services clash when enabled simultaneously.
-  # https://github.com/NixOS/nixos-hardware/issues/260
-  services.tlp.enable = lib.mkDefault (
-    (lib.versionOlder (lib.versions.majorMinor lib.version) "21.05")
-    || !config.services.power-profiles-daemon.enable
-  );
+  config.services.power-profiles-daemon.enable = lib.mkDefault (!config.services.tlp.enable);
```
trueNAHO, line 5: What about conflicts with `services.auto-cpufreq.enable`, as already mentioned in https://github.com/NixOS/nixos-hardware/pull/1630#discussion_r2444855724?

Should `services.auto-cpufreq.enable` be ignored, blindly hoping `services.power-profiles-daemon.enable` and `services.auto-cpufreq.enable` work together, or should this conflict be explicitly handled with something like:

```suggestion
  config.services.power-profiles-daemon.enable = lib.mkDefault (
    !(config.services.auto-cpufreq.enable || config.services.tlp.enable)
  );
```

PR #1482, opened 2025-05-18
Add Lenovo Ideapad 5 Pro 14IMH9 / XiaoXin Pro 14IMH9 2024

pokon548:
###### Description of changes
Added the definition for `Lenovo Ideapad 5 Pro 14IMH9 / XiaoXin Pro 14IMH9 2024`. They are basically the same product but rebranded and sold in different regions.

As there were a lot of vendor specific changes, `README.md` does not document all of them. Comments in `default.nix` should be self-explanatory, though.

###### Things done

### DIFF
--- a/lenovo/ideapad/14imh9/default.nix
+++ b/lenovo/ideapad/14imh9/default.nix
@@ -1,0 +1,110 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel/meteor-lake
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+  ];
+
+  boot = {
+    # Workaround: Out of the box, resuming from hibernation will break sounds.
+    #             See https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/issues/766
+    extraModprobeConfig = ''
+      options snd-hda-intel model=generic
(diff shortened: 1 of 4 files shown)


lenovo/ideapad/14imh9/default.nix
```diff
@@ -0,0 +1,122 @@
...
+      options snd-hda-intel snd-intel-dspcfg.dsp_driver=1
+      blacklist snd_soc_skl
+    '';
+
+    initrd.availableKernelModules = [
+      # Necessary mods
+      "xhci_pci"
+      "nvme"
+      "usbhid"
+      "usb_storage"
```
Mic92, line 24: We actually have nvme enabled by default in nixos. In general we are trying not to replicate nixos-generate-config and complement it. So unless there is some module that doesn't get detected by nixos-generate-config please remove it.

lenovo/ideapad/14imh9/default.nix
```diff
@@ -0,0 +1,122 @@
+{ lib, pkgs, ... }:
+
+{
+  imports = [
+    ../../../common/cpu/intel/meteor-lake
+    ../../../common/gpu/intel/meteor-lake
```
Mic92, line 6: The cpu model will also include the gpu one as of today.

```suggestion
```

PR #1483, opened 2025-05-18
lenovo/legion/15ich: Use Coffee Lake CPU

wrmilling:
###### Description of changes

Updated CPU import for the Lenovo Legion Y530-15ICH to use the generation specific CPU rather than the generic import. 

###### Things done

### DIFF
--- a/lenovo/legion/15ich/default.nix
+++ b/lenovo/legion/15ich/default.nix
@@ -2,7 +2,7 @@
 
 {
   imports = [
-    ../../../common/cpu/intel
+    ../../../common/cpu/intel/coffee-lake
     ../../../common/gpu/nvidia/prime.nix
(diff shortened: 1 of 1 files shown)


lenovo/legion/15ich/default.nix
```diff
@@ -2,7 +2,7 @@
 
 {
   imports = [
-    ../../../common/cpu/intel
+    ../../../common/cpu/coffee-lake
```
Mic92, line 5: error: path '/nix/store/ngfdd9j2fvz3zf5a6vc8zvcg5yxrllql-source/common/cpu/coffee-lake' does not exist

wrmilling: Huh, I have been running it forever, but it was way behind master until today. Will double check.

PR #1491, opened 2025-05-29
dell-precision-3490: add variant without nvidia GPU

s1syph0s:
###### Description of changes

This PR adds the 3490 variant without the nvidia GPU.

The `dell-precision-3490` output is now deprecated, and `*-nvidia` and `*-intel` output are created.

###### Things done

### DIFF
--- a/dell/precision/3490/default.nix
+++ b/dell/precision/3490/default.nix
@@ -1,21 +1,16 @@
 {
+  warnings = [
+    ''
+      DEPRECATED: The <nixos-hardware/dell/precision/3490> module has been deprecated.
+
+      Either use
+        <nixos-hardware/dell/precision/3490/nvidia>
+      for NVIDIA graphics or
+        <nixos-hardware/dell/precision/3490/intel>
+      for Intel graphics.
+    ''
+  ];
   imports = [
-    ../../../common/cpu/intel/meteor-lake
-    ../../../common/gpu/nvidia/ada-lovelace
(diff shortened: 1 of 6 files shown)


dell/precision/3490/intel/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  ];
+
+  boot.initrd.availableKernelModules = [
+    "nvme"
+    "sd_mod"
+    "thunderbolt"
+    "usb_storage"
+    "vmd"
+    "xhci_pci"
+  ];
```
Mic92, line 14: Please only include modules not detected by nixos-generate-config

```suggestion
```

Some of those are also included in NixOS by default (when nixos-generate-config includes the not-detected profile).

dell/precision/3490/default.nix
```diff
@@ -1,21 +1,6 @@
 {
   imports = [
```
Mic92, line 13: Noticed the original profile was still there but unused. How does this look?

s1syph0s: The original profile was meant for backwards compatibility, so that this update won't break the config with 3490 + nvidia that still points to the original profile.

Mic92: Got it. About how this variant that also deprecates the module on this level.

s1syph0s:
the new commit looks good to me

PR #1503, opened 2025-06-17
Add dell precision-5680 hardware

bubuntux:
###### Description of changes
Add dell precision-5680 hardware

###### Things done

### DIFF
--- a/dell/precision/5680/default.nix
+++ b/dell/precision/5680/default.nix
@@ -1,0 +1,66 @@
+{
+  lib,
+  ...
+}:
+{
+  imports = [
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+    ../../../common/cpu/intel
+    ../../../common/gpu/nvidia/prime.nix
+  ];
+
+  boot = {
+    kernelModules = [ "kvm-intel" ];
+    initrd.availableKernelModules = [
+      "xhci_pci"
+      "ahci"
+      "thunderbolt"
+      "nvme"
+      "usb_storage"
+      "sd_mod"
+      "rtsx_pci_sdmmc"
+    ];
+  };
+
+  hardware = {
+    # Audio
+    enableRedistributableFirmware = lib.mkDefault true;
+
+    # Webcam
+    ipu6 = {
+      enable = lib.mkDefault true;
+      platform = lib.mkDefault "ipu6ep";
+    };
+
+    bluetooth = {
+      enable = lib.mkDefault true;
+      powerOnBoot = lib.mkDefault true;
+    };
+
+    graphics.enable = lib.mkDefault true;
+    intel-gpu-tools.enable = lib.mkDefault true;
+
+    nvidia = {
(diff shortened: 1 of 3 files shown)


dell/precision/5680/default.nix
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    graphics = {
+      enable = lib.mkDefault true;
+      enable32Bit = lib.mkDefault true;
+    };
+
+    intel-gpu-tools.enable = lib.mkDefault true;
+    intelgpu = {
+      driver = lib.mkDefault "xe";
```
Mic92: we usually enable that via intel gpu profiles. There is one for each intel generation.

dell/precision/5680/default.nix
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    bluetooth = {
+      enable = lib.mkDefault true;
+      powerOnBoot = lib.mkDefault true;
+    };
+
+    graphics = {
+      enable = lib.mkDefault true;
+      enable32Bit = lib.mkDefault true;
```
Mic92: Useful for gaming but since not everyone has 32-bit games, we don't include those anymore.

```suggestion
```

dell/precision/5680/default.nix
```diff
@@ -0,0 +1,76 @@
...
+    initrd.availableKernelModules = [
+      "xhci_pci"
+      "ahci"
+      "thunderbolt"
+      "nvme"
+      "usb_storage"
+      "sd_mod"
+      "rtsx_pci_sdmmc"
+    ];
+  };
```
Mic92, line 24: We try not to duplicate nixos-generate-config. nixos-hardware is meant to be used together in conjunction with nixos-facter/nixos-generate-config.

```suggestion
```

dell/precision/5680/default.nix
```diff
@@ -0,0 +1,76 @@
...
+        intelBusId = lib.mkDefault "PCI:00:02:0";
+        nvidiaBusId = lib.mkDefault "PCI:01:00:0";
+      };
+    };
+  };
+
+  services = {
+    fwupd.enable = lib.mkDefault true; # update firmware
+    hardware.bolt.enable = lib.mkDefault true; # use thunderbolt
+    pcscd.enable = lib.mkDefault true; # card reader
```
Mic92: I know this module from yubikeys. Does this hardware has some sort of sd-card slot?

dell/precision/5680/default.nix
bubuntux: removed

dell/precision/5680/default.nix
bubuntux: removed

dell/precision/5680/default.nix
bubuntux: removed

dell/precision/5680/default.nix
bubuntux: yes, this laptop comes with a sd-card reader

PR #1509, opened 2025-06-20
Add Lenovo ThinkPad P14s Intel Gen 2

borja-rojo-ilvento:
###### Description of changes

Added a configuration for the Lenovo ThinkPad P14s Gen 2 (which is my current daily driver). Seems it just needed a PCI bus adjustment. It uses an NVidia T500 and Intel Xe graphics. It mimics the implementations for the Gen 3 and Gen 5 models otherwise.

###### Things done

- [x] Ran `nix run ./tests#run .` and passes


borja-rojo-ilvento:
Found the issue. It was an omission of the `common/gpu/intel/tiger-lake` module, which I believe will include the relevant media drivers.

borja-rojo-ilvento:
@mergify queue

lenovo/thinkpad/p14s/intel/gen2/default.nix
```diff
@@ -0,0 +1,18 @@
+{
+  imports = [
+    ../../../../../common/gpu/nvidia/prime.nix
+    ../../../../../common/gpu/nvidia/turing
+    ../../../../../common/cpu/intel/tiger-lake
```
Mic92, line 5: the cpu profile also includes gpu modules. I had it in my memories that the "modesettings" driver was included by default in xserver? That's why I dropped it for now. However please correct me on that in which case I would add this rather to the intel gpu profile than here.

lenovo/thinkpad/p14s/intel/gen2/default.nix
```diff
@@ -0,0 +1,18 @@
...
+    ../../../../../common/pc/laptop
+    ../../../../../common/pc/ssd
+    ../../../../../common/pc
+    ../.
+  ];
+
+  # For suspending to RAM to work, set Config -> Power -> Sleep State to "Linux S3" in EFI.
```
Mic92: Could be a candidate for a README... but not a mandatory change for now.

borja-rojo-ilvento: A README in the some directory as this module? I wouldn't think this would fit in the general README haha

lenovo/thinkpad/p14s/intel/gen2/default.nix
borja-rojo-ilvento: Ok, so I have done testing on this previous, and just double checked the effect of including or excluding `services.xserver.videoDrivers = [ "modesetting" ];` with your other changes. We can indeed exclude the redundant tiger-lake gpu module.

Originally, I found that excluding it caused the computer to hang when changing into suspend mode. After lots of research, I understand that this has to do with the hand-off failing on systems with a discrete GPU, such as thing one. This is also noted in the wiki, [here](https://wiki.nixos.org/wiki/NVIDIA#Offload_mode).

A quick `rg \"modesetting\'` call in the repo yeilds:

```
raspberry-pi/4/modesetting.nix
109:      "modesetting" # Prefer the modesetting driver in X11

raspberry-pi/5/default.nix
24:    Driver "modesetting"

common/gpu/amd/default.nix
6:    services.xserver.videoDrivers = lib.mkDefault [ "modesetting" ];

lenovo/thinkpad/p14s/intel/gen2/default.nix
19:  services.xserver.videoDrivers = [ "modesetting" ];
```

So, it turns out that this isn't actually set. Ostensibly, this 'should be set at needed, implicitly', but reading the wiki, it seems that the 'modesetting' functionality requires different drivers depending on what CPU you have, either AMD or Intel. This means that the `common/gpu/nvidia/prime.nix` module which defines PRIME Offload needs to detect which CPU type you have, and include the correct driver accordingly. Or, a different solution that is CPU aware which I am not well-versed enough to design and implement...

I believe, that although `services.xserver.videoDrivers` does include "modesetting", as per [this](https://search.nixos.org/options?channel=25.05&show=services.xserver.videoDrivers&from=0&size=50&sort=relevance&type=packages&query=services.xserver.videoDrivers) options search, the `common/gpu/nvidia/default.nix` module overrides this with `services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];`, so that the default now excludes "modesetting" and "fbdev".

borja-rojo-ilvento: As a follow-up, I'm going to see what happens if I change `services.xserver.videoDriver = mkDefault [ "nvidia" ];` to `services.xserver.videoDriver = [ "nvidia" ];` and see if Nix will just add the "nvidia" drivers, so I don't need to 'add back' the "modesetting" drivers.

lenovo/thinkpad/p14s/intel/gen2/default.nix
Mic92: Yes. We have per model READMEs usually containing information/quirks that cannot be automated away.

borja-rojo-ilvento: I'll make one. I think it's good form.

lenovo/thinkpad/p14s/intel/gen2/default.nix
borja-rojo-ilvento: That didn't seem to work. Suspend still breaks.

I am adding back in the `"modesetting"` driver. I'll add the link to the Wiki in the upcoming README for clarification.

Mic92: I see. Probably a lot of our hybrid configurations are wrong than. I never had a device dual intel/nvidia so far.

PR #1511, opened 2025-06-22
README: fix team url

pbsds:
I assume this is the intended link


RossComputerGuy (approved):
LGTM, if the homepage PR gets merged soon then let's update the link to that. Otherwise, we can just merge this soon.

PR #1535, opened 2025-07-11
Added Dell Precision 5570

fidgetingbits:
###### Description of changes

Add support for Dell Precision 5570

###### Things done

### DIFF
--- a/dell/precision/5570/default.nix
+++ b/dell/precision/5570/default.nix
@@ -1,0 +1,37 @@
+{
+  config,
+  lib,
+  ...
+}:
+{
+  imports = [
+    ../../../common/cpu/intel/alder-lake
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+    ../../../common/gpu/nvidia/prime-sync.nix
+  ];
+
+  hardware.intelgpu.driver = lib.mkIf (lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8") "xe";
+
+  boot.kernelParams = lib.mkIf (config.hardware.intelgpu.driver == "xe") [
+    "i915.force_probe=!9a49"
+    "xe.force_probe=9a49"
+  ];
+
+  hardware.nvidia = {
+    nvidiaSettings = lib.mkDefault true;
+    modesetting.enable = lib.mkDefault true;
+    open = lib.mkDefault false;
+    prime = {
+      intelBusId = "PCI:0:2:0";
+      nvidiaBusId = "PCI:1:0:0";
+    };
+  };
+
+  # Override the Intel gpu driver setting imported above
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkOverride 990 "nvidia");
+  };
+
+  services.thermald.enable = lib.mkDefault true;
+}
--- a/dell/precision/5570/default.nix
+++ b/dell/precision/5570/default.nix
@@ -8,14 +8,14 @@
     ../../../common/cpu/intel/alder-lake
     ../../../common/pc/laptop
     ../../../common/pc/ssd
-    ../../../common/gpu/nvidia/prime-sync.nix
+    ../../../common/gpu/nvidia/prime.nix
   ];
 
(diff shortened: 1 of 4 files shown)


dell/precision/5570/default.nix
```diff
@@ -0,0 +1,50 @@
...
+    };
+  };
+  hardware.graphics.enable = lib.mkDefault true;
+
+  # Override the Intel gpu driver setting imported above
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkOverride 990 "nvidia");
+  };
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
```
Mic92: We already do this in the undetected profile in nixos-generate-config:

```suggestion
```

dell/precision/5570/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  hardware.nvidia = {
+    nvidiaSettings = lib.mkDefault true;
+    modesetting.enable = lib.mkDefault true;
+    open = lib.mkDefault false;
+    prime = {
+      intelBusId = "PCI:0:2:0";
+      nvidiaBusId = "PCI:1:0:0";
+    };
+  };
+  hardware.graphics.enable = lib.mkDefault true;
```
Mic92: might be worth considering enabling this on all laptops, but than we should do this consistently and not in some modules only:

```suggestion
```

dell/precision/5570/default.nix
```diff
@@ -0,0 +1,50 @@
...
+
+  # Override the Intel gpu driver setting imported above
+  environment.variables = {
+    VDPAU_DRIVER = lib.mkIf config.hardware.opengl.enable (lib.mkOverride 990 "nvidia");
+  };
+
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+  services.thermald.enable = lib.mkDefault true;
+  powerManagement.cpuFreqGovernor = lib.mkDefault "powersave";
```
Mic92: We no longer do set those in nixpkgs, so maybe we shouldn't set it in nixos-hardware either: https://github.com/NixOS/nixpkgs/commit/fb602e5608735f7cb427226f0d2ea539b5129cc2

```suggestion
```

dell/precision/5570/default.nix
```diff
@@ -0,0 +1,50 @@
...
+
+  boot.initrd.availableKernelModules = [
+    "xhci_pci"
+    "thunderbolt"
+    "vmd"
+    "nvme"
+    "usb_storage"
+    "rtsx_pci_sdmmc"
+  ];
+  boot.kernelModules = [ "kvm-intel" ];
```
Mic92: nixos-generate-config already generates those:

```suggestion
```

fidgetingbits:
Thanks for the review. Should be fixed now.

dell/precision/5570/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  config,
+  lib,
+  ...
+}:
+{
+  imports = [
+    ../../../common/cpu/intel/alder-lake
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+    ../../../common/gpu/nvidia/prime-sync.nix
```
KarthikNayak: Why use sync instead of offload?

dell/precision/5570/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  ];
+
+  hardware.intelgpu.driver = lib.mkIf (lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.8") "xe";
+
+  boot.kernelParams = lib.mkIf (config.hardware.intelgpu.driver == "xe") [
+    "i915.force_probe=!9a49"
+    "xe.force_probe=9a49"
```
KarthikNayak: Shouldn't this be `46a6`?

dell/precision/5570/default.nix
fidgetingbits: I don't remember. I've switched it to offload since it seems what most systems use atm, with the caveat that I haven't tested it as I have limited access to that laptop atm.

KarthikNayak:
Thanks, looks good to me now!

KarthikNayak:
> Thanks, looks good to me now!

@Mic92 could you help merge this?

PR #1551, opened 2025-07-24
framework/12-inch: Fix tabletmode 

JohnAZoidberg:
###### Description of changes

NixOS kernels don't have pinctrl_tigerlake built-in, we need to make sure it loads before soc_button_array. Adding it to the initrd ensures that.

###### Things done

### DIFF
--- a/framework/12-inch/13th-gen-intel/default.nix
+++ b/framework/12-inch/13th-gen-intel/default.nix
@@ -4,4 +4,10 @@
     ../common
     ../../../common/cpu/intel
   ];
+
+  # If this module isn't built into the kernel, we need to make sure it loads
+  # before soc_button_array. Otherwise the tablet mode gpio doesn't work.
+  # If correctly loaded, dmesg should show
+  # input: gpio-keys as /devices/platform/INT33D3:00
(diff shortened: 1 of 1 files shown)


framework/12-inch/common/default.nix
```diff
@@ -8,6 +8,12 @@
     ../../framework-tool.nix
   ];
 
+  # If this module isn't built into the kernel, we need to make sure it loads
+  # before soc_button_array. Otherwise the tablet mode gpio doesn't work.
+  # If correctly loaded, dmesg should show
+  # input: gpio-keys as /devices/platform/INT33D3:00
+  boot.initrd.kernelModules = [ "pinctrl_tigerlake" ];
+
```
0x4A6F: This is specific to Intel platform?
So this should be a separate include from `framework/12-inch/common/intel.nix` for all (future) models using this.
If this is probably not used in future models, then specifying this directly in `framework/12-inch/13th-gen-intel/default.nix` makes more sense.

JohnAZoidberg: Second one is true, thanks. I updated it.

Issue #858, opened 2024-02-05
Raspberry Pi 4b: FATAL: Module tpm-crb not found in directory /nix/store


JeanMertz:
I'm trying to use the Raspberry Pi 4 module for my rpi setup, but I can't seem to make it work:

```
error: builder for '/nix/store/s0gf53d463gjqj674b1cbl6zd9z6p8hk-linux-6.1.63-stable_20231123-modules-shrunk.drv' failed with exit code 1;
       last 10 log lines:
       > root module: uas
       >   builtin dependency: uas
       > root module: autofs
       >   builtin dependency: autofs4
       > root module: tpm-tis
       >   copying dependency: /nix/store/gaxzrdl8840rsd18a7vgrq1bqxvfsjjs-linux-6.1.63-stable_20231123-modules/lib/modules/6.1.63/kernel/drivers/char/tpm/tpm.ko.xz
       >   copying dependency: /nix/store/gaxzrdl8840rsd18a7vgrq1bqxvfsjjs-linux-6.1.63-stable_20231123-modules/lib/modules/6.1.63/kernel/drivers/char/tpm/tpm_tis_core.ko.xz
       >   copying dependency: /nix/store/gaxzrdl8840rsd18a7vgrq1bqxvfsjjs-linux-6.1.63-stable_20231123-modules/lib/modules/6.1.63/kernel/drivers/char/tpm/tpm_tis.ko.xz
       > root module: tpm-crb
       > modprobe: FATAL: Module tpm-crb not found in directory /nix/store/gaxzrdl8840rsd18a7vgrq1bqxvfsjjs-linux-6.1.63-stable_20231123-modules/lib/modules/6.1.63
```

This is using the Flakes setup.

This seems to be related to `boot.kernelPackages`.

levigross:
Set the following setting `boot.initrd.systemd.enableTpm2 = false;` and you will be able to build.

PR #1559, opened 2025-08-02
fix(raspberry-pi-4): enable building kernel

yajo:
Fixes https://github.com/NixOS/nixos-hardware/issues/858

### DIFF
--- a/raspberry-pi/4/default.nix
+++ b/raspberry-pi/4/default.nix
@@ -37,6 +37,9 @@
       "reset-raspberrypi" # required for vl805 firmware to load
     ];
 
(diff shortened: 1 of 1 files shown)


flake.lock
panchoh, line 1: Hi!

Note that there is an issue with `flake.lock` not being synced with the actual `flake.nix` that we have, which I’m trying to address on #1570.

At any rate, I’d recommend removing `flake.lock` from the scope of this PR (to keep things orthogonal, that is).

Thanks!

yajo: Thanks! I didn't notice. Solved.

PR #1358, opened 2025-02-03
framework/13-inch/intel-core-ultra-series1: fix intel_vpu firmware error -2

nmouha:
###### Description of changes

Boot error message before patch:

```
$ journalctl | grep vpu
Feb 03 03:10:51 nixos kernel: intel_vpu 0000:00:0b.0: [drm] *ERROR* ivpu_fw_request(): Failed to request firmware: -2
Feb 03 03:10:51 nixos kernel: intel_vpu 0000:00:0b.0: [drm] ivpu_hw_power_down(): NPU not idle during power down
Feb 03 03:10:51 nixos kernel: intel_vpu 0000:00:0b.0: probe with driver intel_vpu failed with error -2
```

Fixed after patch:

```
$ journalctl | grep vpu
Feb 03 03:13:30 nixos kernel: intel_vpu 0000:00:0b.0: [drm] Firmware: intel/vpu/vpu_37xx_v0.0.bin, version: 20240221*MTL_CLIENT_SILICON-release*2101*ci_tag_ud202408_vpu_rc_20240221_2101*845c105994a
Feb 03 03:13:30 nixos kernel: [drm] Initialized intel_vpu 1.0.0 for 0000:00:0b.0 on minor 0
```

Source: https://discourse.nixos.org/t/new-installation-on-asus-zenbook-ux5406-intel-vpu-firmware-error-2/58732/2

###### Things done

- [X] Tested the changes in your own NixOS Configuration
- [X] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

PR #1568, opened 2025-08-06
framework/13-inch/intel-core-ultra-series1: revert #1358 (fix intel_vpu firmware error -2)

nmouha:
###### Description of changes

This reverts #1358 (framework/13-inch/intel-core-ultra-series1: fix intel_vpu firmware error -2).

The firmware is now upstreamed to linux-firmware (https://web.git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/?id=9b870dde196d030a6c8872e8965dd28a0146c990), so #1358 is no longer needed.

Right now, there is no impact to my proposed change (all this does is avoid downloading a file twice). However, it is better to use the upstreamed version in case the firmware gets updated in the future.

###### Things done

### DIFF
--- a/framework/13-inch/intel-core-ultra-series1/default.nix
+++ b/framework/13-inch/intel-core-ultra-series1/default.nix
@@ -17,30 +17,6 @@
     lib.mkDefault pkgs.linuxPackages_latest
   );
 
-  # Intel NPU Driver
-  # https://discourse.nixos.org/t/new-installation-on-asus-zenbook-ux5406-intel-vpu-firmware-error-2/58732/2
-  hardware.firmware = lib.optionals (config.hardware.enableRedistributableFirmware) [
-    (
-      let
-        model = "37xx";
-        version = "0.0";
-
-        firmware = pkgs.fetchurl {
-          url = "https://github.com/intel/linux-npu-driver/raw/v1.13.0/firmware/bin/vpu_${model}_v${version}.bin";
-          hash = "sha256-Mpoeq8HrwChjtHALsss/7QsFtDYAoFNsnhllU0xp3os=";
-        };
-      in
-      pkgs.runCommand "intel-vpu-firmware-${model}-${version}" { } ''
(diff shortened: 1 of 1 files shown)


framework/13-inch/intel-core-ultra-series1/default.nix
```diff
@@ -17,30 +17,6 @@
...
-      pkgs.runCommand "intel-vpu-firmware-${model}-${version}" { } ''
-        mkdir -p "$out/lib/firmware/intel/vpu"
-        cp '${firmware}' "$out/lib/firmware/intel/vpu/vpu_${model}_v${version}.bin"
-      ''
-    )
-  ];
-
-  warnings = lib.mkIf (!config.hardware.enableRedistributableFirmware) [
-    ''For Intel NPU support, set the option: hardware.enableRedistributableFirmware = true;''
-  ];
```
JohnAZoidberg: This is still needed I think

nmouha: I don't think the warning is needed. I am suggesting to remove the warning for two reasons:
- This cleanly reverts #1568, without leaving any lines behind.
- A warning specifically about "NPU support" doesn't make sense anymore, as all the lines related to the Intel NPU Driver will be removed.

Note that enableRedistributableFirmware will still work as expected: if set to `false`, nixpkgs will not include linux-firmware (nor the NPU driver binaries that it now contains).

That said, it's just a warning. If there is a preference to keep all warnings that were previously present, then we can keep this one.

@JohnAZoidberg: Let me know how you'd like this to be resolved!

Mic92: @JohnAZoidberg ping.

nmouha: @Mic92: Could you please approve this PR? I'm just proposing to cleanly revert an earlier contribution of mine (because it is no longer necessary). Being out-of-sync with upstream has some obvious security implications...

PR #1572, opened 2025-08-08
Add ThinkPad T14 Intel Gen 6 support

bashfulrobot:
## Hardware Profile

Adds support for Lenovo ThinkPad T14 Intel Gen 6 with Intel Core Ultra processors (Lunar Lake architecture).

## Changes

- **CPU/GPU Support**: Uses  module for Intel Core Ultra 5/7 processors with Intel Arc integrated graphics
- **Kernel Requirements**: Ensures kernel 6.8+ for full Lunar Lake support  
- **Hardware Optimizations**: 
  -  for ThinkPad brightness control
  -  for touchpad click registration
  - Disables  service for modern Intel power management
- **Module Structure**: Added both base Intel T14 profile and Gen 6 specific profile
- **Documentation**: Updated README.md table and flake.nix entries

## Hardware Tested

- Intel Core Ultra 7 268V processor
- Intel Arc Graphics 130V/140V 
- All hardware functionality verified working
- Kernel 6.16.0 with xe driver loaded successfully

## Feedback Requested

The kernel parameters and throttled service settings are marked with TODO comments as they're based on common ThinkPad patterns. Feedback from other T14 Gen 6 users would help validate these optimizations.

## Test Plan

- [x] Graphics acceleration working (Mesa Intel Graphics LNL)  
- [x] xe driver loading correctly
- [x] Kernel 6.8+ requirement met
- [x] Hardware detection functional
- [ ] Community testing of brightness keys
- [ ] Community testing of touchpad behavior
- [ ] Community validation of power management settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,7 @@ See code for all available configurations.
 | [Lenovo ThinkPad T14 AMD Gen 4](lenovo/thinkpad/t14/amd/gen4)                     | `<nixos-hardware/lenovo/thinkpad/t14/amd/gen4>`         | `lenovo-thinkpad-t14-amd-gen4`    |
 | [Lenovo ThinkPad T14 AMD Gen 5](lenovo/thinkpad/t14/amd/gen5)                     | `<nixos-hardware/lenovo/thinkpad/t14/amd/gen5>`         | `lenovo-thinkpad-t14-amd-gen5`    |
 | [Lenovo ThinkPad T14](lenovo/thinkpad/t14)                                        | `<nixos-hardware/lenovo/thinkpad/t14>`                  | `lenovo-thinkpad-t14`              |
+| [Lenovo ThinkPad T14 Intel Gen 6](lenovo/thinkpad/t14/intel/gen6)                | `<nixos-hardware/lenovo/thinkpad/t14/intel/gen6>`       | `lenovo-thinkpad-t14-intel-gen6`   |
 | [Lenovo ThinkPad T14s AMD Gen 1](lenovo/thinkpad/t14s/amd/gen1)                   | `<nixos-hardware/lenovo/thinkpad/t14s/amd/gen1>`        | `lenovo-thinkpad-t14s-amd-gen1`    |
 | [Lenovo ThinkPad T14s AMD Gen 4](lenovo/thinkpad/t14s/amd/gen4)                   | `<nixos-hardware/lenovo/thinkpad/t14s/amd/gen4>`        | `lenovo-thinkpad-t14s-amd-gen4`    |
 | [Lenovo ThinkPad T14s](lenovo/thinkpad/t14s)                                      | `<nixos-hardware/lenovo/thinkpad/t14s>`                 | `lenovo-thinkpad-t14s`             |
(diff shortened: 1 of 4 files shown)


flake.lock
panchoh, line 1: Hi!

Note that there is an issue with `flake.lock` not being synced with the actual `flake.nix` that we have, which I’m trying to address on #1570.

At any rate, I’d recommend removing `flake.lock` from the scope of this PR (to keep things orthogonal, that is).

Thanks!

panchoh:
> I reviewed the error, and the component that the `nix fmt` is complaining about also does not exist in your current flake. Can I ignore that? Or if there is something I need to do. Please let me know. I am currently running this code from my repo on my laptop. Feel free to let me know if there is anything else you need. Appreciate your time.
> 
> ```
>  nix fmt ..                nixos-hardware on  master took 6m5s error: flake 'git+file:///home/dustin/dev/nix/nixos-hardware' does not provide attribute 'formatter.x86_64-linux'
> ```
> 
> Cheers.

(Sorry, I missed this comment before.)

The formatter has been recently moved (a7600cef40e350c6f3fcb6eefb6f0d6b09c7d05e) to the `tests` flake, so the GitHub Actions workflow enters the `tests` directory before running `nix fmt`:

https://github.com/NixOS/nixos-hardware/blob/26ed7a0d4b8741fe1ef1ee6fa64453ca056ce113/.github/workflows/test.yml#L14

So, just run `cd tests`, and then  run `nix fmt .. -- --fail-on-change` a couple of times until no errors are detected, and you are good to go. (Tecnically, the first pass will fix all the errors, but with the second pass, you get feedback that it won’t fail when run by the workflow.)

Hope that it helps!

Happy hacking,

flake.lock
bashfulrobot: I did end up removing the lock file previously. So that is taken care of. :+1:

panchoh: Um, that doesn't fly, I'm afraid. If this PR gets accepted, the project **as a whole** will end up **without** the `flake.lock` file. We don't want that. What I propose is that you drop the edit  to `flake.lock` altogether. The resulting PR should then contain no mention of `flake.lock`. And, when the previous fix is accepted by the project, this PR will have no conflicts as a bonus.

My suggestion would be that you rewrite the history of your PR, and when the changes to `flake.lock` are gone, then force-push to your branch. Let me know if you need and/or want assistance with this.

Hint: `git rebase --interactive` is your friend.

Thanks!

bashfulrobot: My bad! I thought that lock came in on my local test builds. :fearful: 

I have restored it from the main repo. And it no longer shows in the affected files. Should be good there! 

Cheers.

panchoh:
> Ok, the checks all passed! That's something. 🤣 is there anything else I need to do to get this PR in shape to get approved?
> 
> Cheers.

**Disclaimer**: I’m just a user here.

That being said, I would suggest that you rewrite the history of this PR, leaving only the relevant parts. Ideally, a single commit. This makes the review process easier for the maintainers.

For instance, the `flake.lock` coming and going is just noise here. Or having a commit that adds a chunk only to be reverted in the next one (`security.pam.services`, I’m looking at you :wink:), is of little use, just adds load to the reviewer.

Hope that it helps and, at any rate, thanks for your contribution!

PR #1590, opened 2025-08-26
framework 16: Remove headset quirk

JohnAZoidberg:
###### Description of changes
Framework 16 does not have a built-in headphone jack, this is not needed.
The kernel has also reverted the quick, so matching on a kernel version does nothing useful.

###### Things done

### DIFF
--- a/framework/16-inch/common/default.nix
+++ b/framework/16-inch/common/default.nix
@@ -8,12 +8,6 @@
     ../../framework-tool.nix
   ];
 
-  # Fix TRRS headphones missing a mic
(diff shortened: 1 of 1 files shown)


JohnAZoidberg:
fyi @leana8959 @jkoking, you edited this previously

JohnAZoidberg:
> The kernel has also reverted the quick, so matching on a kernel version does nothing useful.

For reference: https://github.com/torvalds/linux/commit/e799bdf51d54bebaf939fdb655aad424e624c1b1

PR #1597, opened 2025-09-01
panasonic: add Let's Note CF-LX3 configuration

yvan-sraka:

### DIFF
--- a/panasonic/letsnote/cf-lx3/default.nix
+++ b/panasonic/letsnote/cf-lx3/default.nix
@@ -1,0 +1,11 @@
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc/laptop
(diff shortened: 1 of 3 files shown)


panasonic/letsnote/cf-lx3/default.nix
```diff
@@ -0,0 +1,11 @@
+{
```
Mic92, line 1: error: reading from file '/nix/store/l7wh0xx17rl6pl792ya5x1bw4hzdycbc-source/panasonic/letsnote/cf-lx3/default.nix': Is a directory

Remove one level of default.nix to fix the eval error.

Mic92: You basically created a directory called `default.nix`.

PR #1616, opened 2025-09-18
Remove deprecations until there's a migration path

cphalpin:
###### Description of changes

Removes profile deprecations from flake. As discussed in #992 they shouldn't be deprecated until there's a suitable alternative.


cphalpin:
@Lyndeno are you happy for these deprecations to be removed until there's an alternative available?

cphalpin:
@Mic92 Any possibility of getting this merged?

PR #1673, opened 2025-11-13
broadcom-wl: enable WiFi/Bluetooth driver as discussed in PR #1580

masrlinu:
###### Description of changes
Re-enable the Broadcom `wl` (broadcom-sta) WiFi/Bluetooth driver that was removed in PR #1580, and add an option to disable it for users who prefer not to use the vulnerable driver (for example because they replaced it with a USB Wi-Fi adapter due to the security issue).
By default, the Broadcom driver is enabled, so that affected devices regain working Wi-Fi.
The driver is included via an overridden `broadcom_sta` package with cleared `meta.knownVulnerabilities`, which avoids evaluation failures while keeping the module functional.
Tested on a MacBook Air 6,2.

###### Things done

- [X] Tested the changes in my own NixOS Configuration on a MacBook Air 6.2
- [x] Tested the changes end-to-end on a MacBook Air 6.2 by using my fork of `nixos-hardware` and

### DIFF
--- a/apple/imac/14-2/README.md
+++ b/apple/imac/14-2/README.md
@@ -20,6 +20,12 @@
 
 ## Wifi
 - [x] ok
+> **Note:** Enabling WiFi and Bluetooth functionality on this hardware requires the proprietary Broadcom driver. Due to outstanding security issues, you need to explicitly opt-in by setting:
+>
+> ```nix
+> hardware.broadcom.wifi.enableLegacyDriverWithKnownVulnerabilities = true;
+> ```
+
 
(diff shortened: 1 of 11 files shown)


kmk142789 (approved):

masrlinu:
@Mic92, just a quick check: Is this ready to be merged, or do you have any further feedback? 😊

kingscott:
It looks like this is still blocking older Apple hardware from being usable with version 25.11? If so, is there a timeline wherein we can expect a merge? @Mic92

PR #1684, opened 2025-11-25
framework-16-amd-ai-300-series: add nvidia submodule

Mic92:
Add a dedicated submodule for Framework Laptop 16 AMD AI 300 Series with NVIDIA dGPU (RTX 5070). This provides hybrid graphics configuration with PRIME offload, allowing the AMD iGPU to run by default for better battery life while making the NVIDIA dGPU available on demand via nvidia-offload.

Exposed as nixosModules.framework-16-amd-ai-300-series-nvidia

###### Things done

### DIFF
--- a/framework/16-inch/amd-ai-300-series/nvidia/default.nix
+++ b/framework/16-inch/amd-ai-300-series/nvidia/default.nix
@@ -1,0 +1,28 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../.
+    ../../../../common/gpu/nvidia/blackwell
+    ../../../../common/gpu/nvidia/prime.nix
+  ];
+
+  # Explicitly set nvidia as video driver to override modesetting from AMD module
+  services.xserver.videoDrivers = [ "nvidia" ];
+
+  hardware.nvidia = {
+    # Hybrid graphics with PRIME offload for better battery life
+    # AMD iGPU by default, NVIDIA dGPU on demand via nvidia-offload command
+    prime = {
+      # WARNING: These defaults may not match your system!
+      # Bus IDs vary depending on installed expansion cards and NVMe drives.
+      # You MUST override these values - see README.md for instructions.
+      # Bus IDs can be found with `lspci | grep -E "VGA|3D|Display"`
+      amdgpuBusId = lib.mkDefault "PCI:195:0:0";
+      nvidiaBusId = lib.mkDefault "PCI:194:0:0";
+    };
+
+    # Power management for hybrid graphics
+    powerManagement.enable = lib.mkDefault true;
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -173,6 +173,7 @@
           framework-amd-ai-300-series = import ./framework/13-inch/amd-ai-300-series;
           framework-16-7040-amd = import ./framework/16-inch/7040-amd;
           framework-16-amd-ai-300-series = import ./framework/16-inch/amd-ai-300-series;
+          framework-16-amd-ai-300-series-nvidia = import ./framework/16-inch/amd-ai-300-series/nvidia;
           framework-desktop-amd-ai-max-300-series = import ./framework/desktop/amd-ai-max-300-series;
           friendlyarm-nanopc-t4 = import ./friendlyarm/nanopc-t4;
           friendlyarm-nanopi-r5s = import ./friendlyarm/nanopi-r5s;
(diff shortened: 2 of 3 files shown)


framework/16-inch/amd-ai-300-series/nvidia/default.nix
```diff
@@ -0,0 +1,25 @@
...
+  ];
+
+  # Explicitly set nvidia as video driver to override modesetting from AMD module
+  services.xserver.videoDrivers = [ "nvidia" ];
+
+  hardware.nvidia = {
+    # Hybrid graphics with PRIME offload for better battery life
+    # AMD iGPU by default, NVIDIA dGPU on demand via nvidia-offload command
+    prime = {
+      # Bus IDs can be found with `lspci | grep -E "VGA|3D"`
```
JohnAZoidberg:
```suggestion
      # Bus IDs can be found with `lspci | grep -E "VGA|3D|Display"`
```

I need to grep for Display to show the internal GPU

```
> lspci | grep -E "VGA|3D|Display"
c2:00.0 VGA compatible controller: NVIDIA Corporation GB206M [GeForce RTX 5070 Max-Q / Mobile] (rev a1)
c3:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Strix [Radeon 880M / 890M] (rev c1)
```

framework/16-inch/amd-ai-300-series/nvidia/default.nix
```diff
@@ -0,0 +1,25 @@
...
+      amdgpuBusId = lib.mkDefault "PCI:194:0:0"; # c2:00.0 in lspci (AMD iGPU)
+      nvidiaBusId = lib.mkDefault "PCI:193:0:0"; # c1:00.0 in lspci (NVIDIA dGPU)
```
JohnAZoidberg: These are all my non-CPU devices:

```
bf:00.0 Non-Volatile memory controller: Micron Technology Inc 2400 NVMe SSD (DRAM-less) (rev 03)
c0:00.0 Network controller: MEDIATEK Corp. MT7925 (RZ717) Wi-Fi 7 160MHz
c1:00.0 Non-Volatile memory controller: Sandisk Corp SN8000S NVMe SSD (rev 01)
c2:00.0 VGA compatible controller: NVIDIA Corporation GB206M [GeForce RTX 5070 Max-Q / Mobile] (rev a1)
c3:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Strix [Radeon 880M / 890M] (rev c1)
c4:00.0 Non-Essential Instrumentation [1300]: Advanced Micro Devices, Inc. [AMD] Strix/Strix Halo PCIe Dummy Function
c5:00.0 USB controller: Advanced Micro Devices, Inc. [AMD] Device 151f
```

I think because I have two NVMe drives, the graphics cards get pushed back by one.

I regularly dump my system config when I update the BIOS or make changes. I have seen NVIDIA show up at 193 and 194. And AMD at 194, 195, 196

framework/16-inch/amd-ai-300-series/nvidia/default.nix
```diff
@@ -0,0 +1,25 @@
...
+    };
+
+    # Power management for hybrid graphics
+    powerManagement.enable = lib.mkDefault true;
```
JohnAZoidberg, line 26: I think this is fine.
But for reference, nvidia-powerd does nothing because we found it couldn't support some configurations we needed, so we hardcoded the power value that powerd reads from the VBIOS and we adjust it from the EC firmware instead.

This also enables nvidia-persistenced: https://docs.nvidia.com/deploy/driver-persistence/index.html
I wasn't familiar with that yet, but that also seems fine to enable.

framework/16-inch/amd-ai-300-series/nvidia/default.nix
Mic92: Should we rather not set this or as a default and ask user to set their own value?

framework/16-inch/amd-ai-300-series/nvidia/default.nix
Mic92: Actually this enables a script writes to /proc/driver/nvidia/suspend to tell the NVIDIA driver to save/restore its state:

On suspend/hibernate: Switches to VT 63, writes "suspend"/"hibernate" to /proc/driver/nvidia/suspend
On resume: Writes "resume" to /proc/driver/nvidia/suspend, restores the original VT

Mic92: This is not related `nvidia-powerd` afaik.

PR #1711, opened 2025-12-29
Add module to patch ath12k firmware

TwoUnderscorez:
###### Description of changes

Added a module that I wrote for myself upon receiving my Qualcomm QCNCM865 WiFi card and discovering that it didn't work out of the box.

For more details please read the [readme](https://github.com/TwoUnderscorez/nixos-hardware/blob/patch-ath12k-fw/linux-firmware/ath12k/README.md) and [this](https://bugzilla.kernel.org/show_bug.cgi?id=220897) but report on the Kernel's Bugzilla.

###### Things done

### DIFF
--- a/linux-firmware/ath12k/default.nix
+++ b/linux-firmware/ath12k/default.nix
@@ -3,7 +3,8 @@
   lib,
   config,
   ...
-}: {
+}:
+{
   options = {
     linux-firmware-ath12k-patched = {
       enable = lib.mkEnableOption "enables linux-firmware-ath12k-patched";
@@ -38,10 +39,11 @@
   };
 
   config = lib.mkIf config.linux-firmware-ath12k-patched.enable {
-    hardware.firmware = with pkgs;
-    with config.linux-firmware-ath12k-patched; [
-      (stdenvNoCC.mkDerivation
-        rec {
+    hardware.firmware =
+      with pkgs;
+      with config.linux-firmware-ath12k-patched;
+      [
+        (stdenvNoCC.mkDerivation rec {
           pname = "linux-firmware-ath12k-patched";
           version = "0-unstable-2025-12-05";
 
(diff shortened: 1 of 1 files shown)


linux-firmware/ath12k/README.md
```diff
@@ -0,0 +1,81 @@
+# linux-firmware/ath12k
```
Mic92, line 1: Do you you know any laptop module where this apply? I mainly want to get some basic evaluation coverage for this code.

TwoUnderscorez: Is [this](https://community.frame.work/t/guide-successful-wi-fi-7-802-11be-on-framework-13-amd-with-qualcomm-qcncm865-and-arch-linux/44723) a good example?
I don't know of any laptop that comes with a wifi chip with unsupported device ids as I have a desktop. I did try to plug the module into a few laptops of mine in case it was my desktop motherboard's fault, but all my computers ended up not recognizing this chip exactly the same.
Because the code merely adds the device id you define to the json and doesn't override anything, you could test it on a computer with Qualcomm wifi chip that already works and make sure that it still works when using this module. 
Sorry, this is indeed a very quirky module for very quirky hardware, let me know if there is anything else I can do.

Mic92: Maybe it's time to find an alternative to parsing our README.md table to figure out what needs to be evaluated and instead use flake outputs to import each module into its own nixos configuration. Than your module would be tested automatically.

TwoUnderscorez: Oh I didn't know that the readme was being parsed automatically.
Is there anything you want me to do?

TwoUnderscorez: @Mic92 Do you think this can be merged or is there anything you'd like me to change?

PR #1713, opened 2025-12-30
add razer-blade-14-RZ09-0530

Mic92:
###### Description of changes

fixes https://github.com/NixOS/nixos-hardware/pull/1706

###### Things done


common/wifi/mediatek/mt7925/default.nix
```diff
@@ -0,0 +1,80 @@
+# MediaTek MT7925 WiFi 7 / Bluetooth support
```
Mic92, line 1: I think framework has the same. Have to check in the new year.

AdrielVelazquez: The framework I have has a MT7925e. I can also test out these kernel changes.

Issue #1603, opened 2025-09-04
Framework Laptop 13 microphone stopped working


ncfavier:
After updating to nixpkgs dfb2f12e899db4876308eba6d93455ab7da304cd (Linux 6.16.3, PipeWire 1.4.7) on my Framework Laptop 13 (AMD Ryzen AI 300), the built-in microphone stopped working. Following the advice of the [Arch Linux wiki](https://wiki.archlinux.org/title/PipeWire#Microphone_is_not_detected_by_PipeWire) I was able to fix this with the following configuration:

```nix
{
  services.pipewire.wireplumber.extraConfig.no-ucm = {
    "monitor.alsa.properties" = {
      "alsa.use-ucm" = false;
    };
  };
}
```

With UCM (Use Case Manager) profiles enabled, the default selected profile under the Configuration tab of pavucontrol under "Family 17h/19h/1ah HD Audio Controller" is "Play HiFi quality Music (Headphones, Headset, Mic1)", which, aside from being an absolute trainwreck of a sentence, renders the built-in microphone inoperational. Switching to a different profile (with different items inside the brackets) made it work again, but more things were broken.

With UCM disabled, the default profile is "Analog Stereo Duplex" as before, which works just fine.

I don't know what this is or where it came from but it clearly is not ready yet, so I suggest we add the above configuration for Framework 13 laptops (which ones?).

Iceburgino:
It's not only frameworks this this should be investigated for. I have this issue (and the fix is working for me) on Legion 5 with Ryzen AI 7 350.

bLaCkwEw:
Similar (same?) issue here.
I'm on a desktop with an AMD Ryzen 5 5600G CPU.
The microphone appears as an input device, it can be selected, it just doesn't pick up sound anymore.

I tried your fix, but it unfortunately doesn't work for me.

I have no idea what triggered the issue. Yesterday the microphone worked fine, today it just doesn't anymore. I didn't do any kind of updates in this period.

I tried booting into previous versions of NixOS with no luck.

bLaCkwEw:
> I didn't do any kind of updates in this period.

I did [add lazygit to my system](https://github.com/bLaCkwEw/nix-config/commit/5782404e067ada76f83de3e717e1ef7464e0572f) though home-manager. Though, I'm unsure how this could have triggered the issue.

d-goldin:
[...]

PR #1717, opened 2025-12-30
Disable UCM globally on FW16

dniku:
For Framework Laptop 16 *only*, this works around https://github.com/NixOS/nixos-hardware/issues/1603. The extra snippet is taken from the the first comment in the issue (with the `10-` prefix added to the rule name for ordering stability).

For FW13, https://github.com/NixOS/nixos-hardware/pull/1709 was merged a few days ago. I'm not sure if it resolves the same issue, and I would appreciate a comment from @alirezamirsepassi on whether disabling UCM is the correct solution for FW16, since its `default.nix` looks a bit different compared to FW13.

###### Things done

### DIFF
--- a/framework/16-inch/amd-ai-300-series/default.nix
+++ b/framework/16-inch/amd-ai-300-series/default.nix
@@ -26,4 +26,15 @@
     "amdgpu.sg_display=0"
     "amdgpu.abmlevel=0"
   ];
+
+  # Work around broken ALSA UCM profiles. Symptoms include incorrect default routing
+  # (e.g. “Headphones” with nothing plugged in) and broken internal mic/speaker routes.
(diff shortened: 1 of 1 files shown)


dniku:
@alirezamirsepassi see https://github.com/NixOS/nixos-hardware/issues/1603.

tl;dr: I put together a [fix for FW16](https://github.com/NixOS/nixos-hardware/pull/1718) that a maintainer of of `alsa-ucm-conf` [cleaned up and merged into upstream](https://github.com/alsa-project/alsa-ucm-conf/issues/673); unfortunately, it seems to have [broken FW13](https://github.com/NixOS/nixos-hardware/issues/1603#issuecomment-3762819356).

I suggest we consolidate the discussion in https://github.com/NixOS/nixos-hardware/issues/1603.

PR #1728, opened 2026-01-14
gpd-pocket4: fix suspend and device type

MarcelCoding:
###### Description of changes

Suspend got blocked by the fingerprint reader, (like 9 out of 10 times) This is probably caused, because there is not linux driver for the device,

I am unsure whether this is suited for nixos-hardware?
```nix
systemd.tpm2.enable = false;
boot.initrd.systemd.tpm2.enable = false;
````
If I don't disable the tpm it delays the boot for like 1.5min until the systemd timeout is reached. I don't know if this happens because I did not configure it properly (I actually did not configure anything tpm related) or because it's not supported?

Maybe some other GPD Pocket 4 users could verify the changes?

###### Things done

### DIFF
--- a/gpd/pocket-4/default.nix
+++ b/gpd/pocket-4/default.nix
@@ -33,4 +33,17 @@ in
 
   # More HiDPI settings
   services.xserver.dpi = 343;
+
+  # disable fingerprint reader, it currently has no linux drivers anyways and also blocks suspend (at least on my system)
+  services.udev.extraRules = /* udev */ ''
+    ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="2808", ATTR{idProduct}=="0752", ATTR{authorized}="0"
+  '';
+
+  # classify device as convertible (see hostnamectl)
+  environment.etc.machine-info = {
+    text = ''
+      CHASSIS=convertible
+    '';
+    mode = "0440";
+  };
 }

### CODE REVIEW

**Comment by flokli** [comment]
So I do see a delay due to the TPM device, but it's 15s or so, not 1.5mins (and I don't know how much of that time is actually waiting for me to finish typing in my passphrase).

I don't see how the fingerprint reader is related to the TPM. Can you elaborate?

File: gpd/pocket-4/default.nix [hunk 1]
```diff
@@ -33,4 +33,17 @@ in
 
   # More HiDPI settings
   services.xserver.dpi = 343;
+
+  # disable fingerprint reader, it currently has no linux drivers anyways and also blocks suspend (at least on my system)
+  services.udev.extraRules = /* udev */ ''
+    ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="2808", ATTR{idProduct}=="0752", ATTR{authorized}="0"
+  '';
+
+  # classify device as convertible (see hostnamectl)
```
flokli, line 42: This should be fixed upstream, via hwdb. I just sent https://github.com/systemd/systemd/pull/40354.

In the meantime, you can add this to your config:

```nix
  services.udev.extraHwdb = ''
    # GPD Pocket 4 chassis type
    dmi:bvnAmericanMegatrendsInternational*:rvnGPD:rnG1628-04*
        ID_CHASSIS=convertible
  '';
```

It needed a reboot, but with this, chassis also shows as "convertible"

MarcelCoding:
Hmm, my friend has the same device and also don't has these issues. He is using hyprland and I am using gnome. I thought that some gnome stuff is trying to using the fingerprint reader and then it blocks suspend because of a strange state.

PR #1749, opened 2026-01-28
framework: Add JohnAZoidberg as codeowner

JohnAZoidberg:
I'd like to review new changes - I am firmware/software engineer at Framework.


CODEOWNERS
```diff
@@ -9,6 +9,6 @@ pine64/rockpro64 @tomfitzhenry
 pine64/star64 @fgaz
 purism/librem/13v3 @yegortimoshenko
 system76/darp6 @khumba
-framework @emiller88
+framework @emiller88 @johnazoidberg
```
Mic92, line 12: @JohnAZoidberg I think this needs write access to the nixos-hardware repo. I can grant you that access.

PR #1754, opened 2026-02-04
Add Minisforum UM series mini PC configurations

johbo:
###### Description of changes

Add hardware support for:
- UM690 (AMD Ryzen 9 6900HX, Radeon 680M)
- UM690S (slim variant)
- UM790 Pro (AMD Ryzen 9 7940HS, Radeon 780M)

These share a common um-series.nix module for consistent configuration across the UM series AMD-based mini PCs.

###### Things done

### DIFF
--- a/minisforum/um-series.nix
+++ b/minisforum/um-series.nix
@@ -1,6 +1,6 @@
 # Common configuration for Minisforum UM series mini PCs
 # (UM690, UM690S, UM790 Pro)
-{ lib, ... }:
+{ ... }:
 
 {
   imports = [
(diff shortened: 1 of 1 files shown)


johbo:
Thanks for taking a look and merging @Mic92 👍 

Have one question which I could not quite resolve based on the readme: I wonder if the removal of the aspects which `nixos-generate-config` creates doesn't make it more difficult to use this repository. My assumption was that I would include a module and typically be all set, with the removal it seems the idea is that I still should have an additional module where I maintain the aspects which would be detected by `nixos-generate-config`. Is this the indented approach for `nixos-hardware`?

PR #1773, opened 2026-02-21
raspberry-pi/5: set generic-extlinux-compatible by default

fkokosinski:
###### Description of changes

This PR provides an overlay for Raspberry Pi 5 U-Boot packages. Two U-Boot packages are provided: rev 1.0 and rev 1.1 compatible ones. Additionally, extlinux is made default in the Raspberry Pi 5 profile.

~This PR provides an installer as well for the board. The `options.hardware.rpiRevision` option is provided to choose between Raspberry Pi 5 board revisions. This in turn chooses the correct U-Boot package from the overlay.~

###### Things done

### DIFF
--- a/raspberry-pi/5/default.nix
+++ b/raspberry-pi/5/default.nix
@@ -12,6 +12,8 @@ in
   imports = [ ../common/default.nix ];
 
   boot = {
+    loader.grub.enable = lib.mkDefault false;
+    loader.generic-extlinux-compatible.enable = lib.mkDefault true;
     kernelPackages = lib.mkDefault (
       pkgs.linuxPackagesFor (pkgs.callPackage ../common/kernel.nix { rpiVersion = 5; })
     );


fkokosinski:
@Cryolitia @Mic92 @concatime - pinging you since you seem to be active w/r to RPi support. I'd appreciate if you could take a look at this PR  and review the changes. Thanks!

raspberry-pi/5/overlay.nix
```diff
@@ -0,0 +1,16 @@
...
+    defconfig = "rpi_arm64_defconfig";
+    extraMeta.platforms = [ "aarch64-linux" ];
+    filesToInstall = [ "u-boot.bin" ];
+  };
+
+  ubootRaspberryPi5_rev1_1 = final.buildUBoot {
+    defconfig = "rpi_arm64_defconfig";
+    # Mote: this patch can be removed once U-Boot starts setting
+    # `fdtfile=bcm2712d0-rpi-5.dtb` automatically for rev1.1 boards
+    patches = [ ./uboot-rpi5-bcm2712d0.patch ];
```
fkokosinski: It's something that I'm actually trying to fix with this series: https://lore.kernel.org/u-boot/20260216184452.18732-1-filip.kokosinski@gmail.com/.

But they are still under review & testing, so I didn't want to straight up include them here.

raspberry-pi/5/overlay.nix
```diff
@@ -0,0 +1,16 @@
+final: _prev: {
+  ubootRaspberryPi5_rev1_0 = final.buildUBoot {
+    defconfig = "rpi_arm64_defconfig";
+    extraMeta.platforms = [ "aarch64-linux" ];
+    filesToInstall = [ "u-boot.bin" ];
+  };
```
concatime: This should probably be upstreamed to nixpkgs after [this block](https://github.com/NixOS/nixpkgs/blob/82871139974077ba7ccfbf00803be636ed71be59/pkgs/misc/uboot/default.nix#L749-L753). It should probably be renamed to `ubootRaspberryPiAarch64` like `ubootQemuAarch64`. WDYT?

concatime (commented):
It's been quite some time since I used the sd-image installer, but can't we instead upstream the changes to nixpkgs's `nixos/modules/installer/sd-card/sd-image-aarch64.nix`?

raspberry-pi/5/overlay.nix
fkokosinski: Good idea - it looks like doing that in `nixpkgs` is fine as long as we use upstream sources, which we do.

Upstreaming this in https://github.com/NixOS/nixpkgs/pull/493816.

fkokosinski:
> My Raspberry Pi 5 is destroyed and fixing, I can test it later when I get it.

That would be much appreciated. Thanks!

raspberry-pi/5/overlay.nix
JamieMagee, line 1: I think you can drop this now https://github.com/NixOS/nixpkgs/pull/493816 has merged?

raspberry-pi/5/overlay.nix
JamieMagee: Looks like this got merged
- https://github.com/u-boot/u-boot/commit/c15a791972548150d8c0616a53e70dada1d358ca
- https://github.com/u-boot/u-boot/commit/4e64ea88585d9650af867718e1c629a1e57c0157

No longer necessary once 2026.4 is cut?

raspberry-pi/5/default.nix
```diff
@@ -5,6 +5,13 @@
   ...
 }:
 {
+  nixpkgs.overlays = [
+    (import ./overlay.nix)
+  ];
+
+  boot.loader.grub.enable = lib.mkDefault false;
+  boot.loader.generic-extlinux-compatible.enable = lib.mkDefault true;
```
JamieMagee: nit: move these inside the existing `boot` block

raspberry-pi/5/overlay.nix
concatime: Congrats @fkokosinski on getting it merged!

raspberry-pi/5/overlay.nix
JamieMagee: I opened a PR to upgrade to u-boot 2026.4 https://github.com/NixOS/nixpkgs/pull/509062

fkokosinski: Yup, it is no longer necessary - I've dropped it from this PR.

raspberry-pi/5/overlay.nix
fkokosinski: Yup, done.

raspberry-pi/5/default.nix
fkokosinski: Done

fkokosinski:
@JamieMagee @concatime sorry guys for the delay here. I've applied your suggestions and updated the commit msg accordingly. Would appreciate if you could take another look at this PR. Thanks!

JamieMagee (approved):

PR #1788, opened 2026-03-09
raspberry-pi: add config.txt generation module

JamieMagee:
Two new files in `raspberry-pi/common/` for generating Raspberry Pi config.txt from Nix.
   
`config-txt.nix` adds `hardware.raspberry-pi.configtxt.settings`, which covers options, dtparam, dtoverlay, and conditional filters (`[pi5]`, `[cm4]`, etc.).
   
`config-txt-defaults.nix` ships the standard Raspberry Pi OS defaults[^1], all `mkDefault`.
   
Not exposed in flake.nix. These are internal, meant to be imported by board profiles once the RPi 5 sub-module work follows. Doesn't write to `/boot/firmware/` yet, that needs the bootloader module.
   
Context: porting config.txt infrastructure from nvmd/nixos-raspberrypi (see nvmd/nixos-raspberrypi#78).

[^1]: https://github.com/RPi-Distro/pi-gen/blob/master/stage1/00-boot-files/files/config.txt

### DIFF
--- a/raspberry-pi/common/config-txt-defaults.nix
+++ b/raspberry-pi/common/config-txt-defaults.nix
@@ -1,0 +1,29 @@
+# Default config.txt settings matching the official Raspberry Pi OS defaults.
+#
+# Source: https://github.com/RPi-Distro/pi-gen/blob/master/stage1/00-boot-files/files/config.txt
+# All values use mkDefault so they can be easily overridden.
+#
+# Reference: https://www.raspberrypi.com/documentation/computers/config_txt.html
+
+{ lib, ... }:
+
+{
+  hardware.raspberry-pi.configtxt.settings = {
+    all = {
+      camera_auto_detect = lib.mkDefault true;
+      display_auto_detect = lib.mkDefault true;
+      max_framebuffers = lib.mkDefault 2;
+      disable_fw_kms_setup = lib.mkDefault true;
+      disable_overscan = lib.mkDefault true;
+      arm_boost = lib.mkDefault true;
+      dtparam = lib.mkDefault [ "audio=on" ];
+      dtoverlay = lib.mkDefault [ "vc4-kms-v3d" ];
+    };
+    cm4 = {
+      otg_mode = lib.mkDefault true;
+    };
+    cm5 = {
+      dtoverlay = lib.mkDefault [ "dwc2,dr_mode=host" ];
+    };
+  };
+}
--- a/raspberry-pi/common/config-txt.nix
+++ b/raspberry-pi/common/config-txt.nix
@@ -1,0 +1,130 @@
+# config.txt generation module for Raspberry Pi
+#
+# Generates config.txt from structured Nix options. The type system models
+# config.txt as nested attrs: each nesting level adds a conditional filter,
+# and leaves are config values. Repeated keys (dtparam, dtoverlay, gpio)
+# are supported via lists.
+#
+# Based on work from nvmd/nixos-raspberrypi (MIT License) and rendering
+# approach suggested by @quentinmit. Follows RFC 42.
+#
+# Reference: https://www.raspberrypi.com/documentation/computers/config_txt.html
+
+{ lib, config, ... }:
+
+let
+  cfg = config.hardware.raspberry-pi.configtxt;
+
+  mkValueString =
+    v:
+    if builtins.isInt v then
+      toString v
+    else if builtins.isString v then
+      v
+    else if true == v then
+      "1"
+    else if false == v then
+      "0"
+    else
+      builtins.abort "config.txt: unsupported value type: ${builtins.typeOf v}";
+
+  mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
+
+  # Recursively flatten nested attrs into a list of { conditionals, name, value } records.
+  # Each nesting level (except the leaf) is a conditional filter.
+  # "all" filters are omitted from the conditionals list since they reset state.
+  recurse =
+    path: value:
+    if builtins.isAttrs value && !(builtins.isList value) && !(value ? _type) then
+      lib.flatten (lib.mapAttrsToList (name: recurse ([ name ] ++ path)) value)
+    else
+      {
+        conditionals = lib.sort builtins.lessThan (lib.filter (k: k != "all") (lib.tail path));
+        name = lib.head path;
+        inherit value;
+      };
+
+  # Group flattened items by their conditional filter set, then render.
+  groupItems =
+    items:
+    lib.mapAttrsToList (groupJSON: groupItems: {
+      conditionals = builtins.fromJSON groupJSON;
+      items = builtins.listToAttrs groupItems;
+    }) (builtins.groupBy (x: builtins.toJSON x.conditionals) items);
+
+  mkGroup =
+    group:
+    lib.concatMapStrings (k: "[${k}]\n") group.conditionals
+    + lib.generators.toKeyValue {
+      inherit mkKeyValue;
+      listsAsDuplicateKeys = true;
+    } group.items;
+
+  toConfigTxt =
+    attrs:
+    let
+      groups = lib.sort (a: b: a.conditionals < b.conditionals) (
+        groupItems (lib.flatten (recurse [ ] attrs))
+      );
+    in
+    lib.concatMapStringsSep "\n[all]\n" mkGroup groups;
+
+in
+{
+  options.hardware.raspberry-pi.configtxt = {
+    settings = lib.mkOption {
+      type =
+        with lib.types;
+        let
+          atom = oneOf [
+            str
+            int
+            bool
+          ];
+          molecule = oneOf [
+            atom
+            (listOf atom)
+            (attrsOf molecule)
+          ];
+        in
+        attrsOf molecule;
+      default = { };
+      description = ''
+        Structured configuration for the Raspberry Pi `config.txt` file.
+
+        Top-level keys are conditional filter sections (`all`, `pi4`, `pi5`,
+        `cm4`, etc.). Nesting adds stacked conditional filters. Leaves are
+        config values rendered as `key=value`. Lists produce repeated keys.
+
+        Booleans render as `0`/`1`. Use `null` with `mkForce` to remove a
+        default.
+
+        See <https://www.raspberrypi.com/documentation/computers/config_txt.html>
+      '';
+      example = lib.literalExpression ''
+        {
+          all = {
+            arm_boost = true;
+            disable_overscan = true;
+            dtparam = [ "audio=on" ];
+            dtoverlay = [ "vc4-kms-v3d" ];
+          };
+          pi5.arm_freq = 2400;
+          cm4.otg_mode = true;
+        }
+      '';
+    };
+
+    file = lib.mkOption {
+      type = lib.types.str;
+      description = ''
+        The generated config.txt content. Defaults to the rendered output of
+        `settings`, but can be overridden to supply a custom config.txt.
+      '';
+    };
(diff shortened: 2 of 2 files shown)


raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+      options,
+      dt-params,
+      dt-overlays,
+    }:
+    let
+      allConfig = lib.concatStringsSep "\n" (
+        lib.filter (x: x != "") [
+          (renderOptions (filterSettings options))
+          (renderDtParams (filterSettings dt-params))
+          (renderDtOverlays (filterSettings dt-overlays))
```
quentinmit: Why are these rendered separately? Can't the submodule set `options.dtparam` and `options.dtoverlay` and then let `renderOptions` handle them?

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+      rendered = lib.filter (x: x != "") (map (name: renderConfigSection name conf.${name}) orderedNames);
+    in
+    lib.concatStringsSep "\n" rendered;
+
+  # Overlay params type: attribute set of nullable values
+  dtOverlayParamsType = with lib.types; attrsOf (nullOr valueType);
+
+  configSectionType = lib.types.submodule {
+    options = {
+      options = lib.mkOption {
```
quentinmit: Having these three keys nested under a conditional (`configtxt.settings.all.options`, `configtxt.settings.all.dt-params`, `configtxt.settings.all.dt-overlays`) feels very cumbersome to me.  I think it would be more in keeping with RFC 42 to keep `settings` a direct map to the `config.txt` file, and put `dt-params` one level up. For example, `configtxt.settings.all.arm_boost`, `configtxt.dt-params.all.audio`, `configtxt.dt-overlays.all.vc4-kms-v3d`.

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+        type = with lib.types; attrsOf (nullOr valueType);
```
quentinmit: Some parameters (like `dtparam`, `dtoverlay`, and `gpio`) can be specified multiple times. I suggest representing that here as `nullOr (oneOf valueType (listOf valueType))`.

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+      paramStr = lib.concatStringsSep "," paramList;
+    in
+    if paramStr == "" then "dtoverlay=${name}" else "dtoverlay=${name},${paramStr}";
+
+  renderDtOverlays =
+    overlays:
+    let
+      filtered = filterSettings overlays;
+    in
+    lib.concatMapStringsSep "\n" (name: renderDtOverlay name filtered.${name}) (lib.attrNames filtered);
```
quentinmit: Instead of manually rendering all of these values, I suggest using the existing machinery in `lib.generators`. Here's the logic I use to turn attrs into a `config.txt` (while also supporting multiple conditions and multiple values for the same key):

```nix
let
  toConfigTxt = with builtins; let
    recurse = path: value:
      if isAttrs value then
        lib.mapAttrsToList (name: recurse ([ name ] ++ path)) value
      else
        {
          conditionals = lib.lists.sort builtins.lessThan (filter (k: k != "all") (tail path));
          name = head path;
          inherit value;
        };
    groupItems = items:
      (lib.attrsets.mapAttrsToList
        (groupJSON: items:
          {
            conditionals = fromJSON groupJSON;
            inherit items;
          })
        (lib.attrsets.mapAttrs
          (k: builtins.listToAttrs)
          (builtins.groupBy
            (x: toJSON x.conditionals)
            items
          )
        )
      );
    mkValueString = v:
      if isInt v then toString v
      else if isString v then v
      else if true == v then "1"
      else if false == v then "0"
      else abort "the value is not supported: ${toPretty {} (toString v)}";
    mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
    mkGroup = group:
      lib.strings.concatMapStrings (k: "[${k}]\n") group.conditionals
        + lib.generators.toKeyValue { inherit mkKeyValue; listsAsDuplicateKeys = true; } group.items
    ;
    in
      attrs:
      let
        groups = lib.lists.sort
          (a: b: a.conditionals < b.conditionals)
          (groupItems (lib.flatten (recurse [] attrs)));
      in
        lib.strings.concatMapStringsSep "\n[all]\n" mkGroup groups;
```

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+        `hardware.raspberry-pi.configtxt.generated`.
+
+        Note: This module only generates the config.txt content. Writing it to
+        the firmware partition (`/boot/firmware/config.txt`) must be handled
+        separately, either manually or by a bootloader module.
+      '';
+    };
+
+    settings = lib.mkOption {
+      type = with lib.types; attrsOf configSectionType;
```
quentinmit: `config.txt` actually supports [multiple conditional filters](https://www.raspberrypi.com/documentation/computers/config_txt.html#combine-conditional-filters) for each config block, so a single `attrsOf` level doesn't allow the full set of `config.txt` configuration. I suggest representing this as arbitrary nested `attrsOf`, so e.g. `configtxt.pi4."gpio4=1".arm_boost = true` means "Turn on `arm_boost` on a Pi 4 when GPIO 4 is true at boot.

This is the type I settled on to fully represent `config.txt`:

```nix
        type = with types; let
          atom = oneOf [str int bool];
          molecule = oneOf [atom (listOf atom) (attrsOf molecule)];
        in attrsOf molecule;
```

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+          (renderDtOverlays (filterSettings dt-overlays))
+        ]
+      );
+    in
+    # Skip empty sections entirely
+    if allConfig == "" then
+      ""
+    else
+      ''
+        [${conditionalFilter}]
```
quentinmit: This actually doesn't properly scope the block. While `config.txt` looks vaguely like an INI file, conditional filters are actually additive - so

```
[gpio2=0]
arm_boost=true
[pi4]
max_framebuffers=2
```
will only apply `max_framebuffers=2` on a Pi 4 *with gpio2 true`.

According to https://www.raspberrypi.com/documentation/computers/config_txt.html#combine-conditional-filters, you need to first insert an `[all]` to reset the current conditionals, before applying (a) new one(s)

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+            dt-overlays.vc4-kms-v3d = { };
+          };
+          pi5.options.arm_freq = 2400;
+        }
+      '';
+    };
+
+    generated = lib.mkOption {
+      type = lib.types.str;
+      readOnly = true;
```
quentinmit: Does this need to be read-only? I feel like I've seen a bunch of other NixOS modules that have a writable `file` attribute that defaults to a generated file, so users can also opt out of this config system and supply their own.

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,223 @@
...
+      description = ''
+        The generated config.txt content. Read-only.
+        This can be used by bootloader modules or scripts to write the file
+        to the firmware partition.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.configtxt.enable {
```
quentinmit: Why is `generated` gated on `configtxt.enable`? Why not unconditionally set it? It's not as if this module is currently installing a config.txt file.

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,130 @@
...
+        attrsOf molecule;
+      default = { };
+      description = ''
+        Structured configuration for the Raspberry Pi `config.txt` file.
+
+        Top-level keys are conditional filter sections (`all`, `pi4`, `pi5`,
+        `cm4`, etc.). Nesting adds stacked conditional filters. Leaves are
+        config values rendered as `key=value`. Lists produce repeated keys.
+
+        Booleans render as `0`/`1`. Use `null` with `mkForce` to remove a
```
quentinmit, line 107: Just to check, does `null` still work? I think right now you don't allow `null` as a value type (nor do I see anything that would filter it out when generating the `config.txt`).

raspberry-pi/common/config-txt-defaults.nix
```diff
@@ -0,0 +1,29 @@
...
+      disable_overscan = lib.mkDefault true;
+      arm_boost = lib.mkDefault true;
+      dtparam = lib.mkDefault [ "audio=on" ];
+      dtoverlay = lib.mkDefault [ "vc4-kms-v3d" ];
+    };
+    cm4 = {
+      otg_mode = lib.mkDefault true;
+    };
+    cm5 = {
+      dtoverlay = lib.mkDefault [ "dwc2,dr_mode=host" ];
```
quentinmit, line 26: Hm, so you opted to remove the special handling of `dtparam` and `dtoverlay` entirely. I think they probably do want to keep a different higher-level config surface, because right now it's not possible to override a single param or overlay param without setting the entire list. (e.g. setting `dtoverlay = [ "spi" ];` in a config would disable the `dwc2` overlay as a side effect.)

But I also think that could be added later without breaking any existing configs, so maybe it's better to start small. *shrug*

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,130 @@
...
+            dtoverlay = [ "vc4-kms-v3d" ];
+          };
+          pi5.arm_freq = 2400;
+          cm4.otg_mode = true;
+        }
+      '';
+    };
+
+    file = lib.mkOption {
+      type = lib.types.str;
```
quentinmit: A `file` option should be of type `path`. See e.g. `services.mysql.configFile`, `services.logrotate.configFile`, `services.httpd.configFile`, etc.

raspberry-pi/common/config-txt.nix
```diff
@@ -0,0 +1,130 @@
...
+      description = ''
+        The generated config.txt content. Defaults to the rendered output of
+        `settings`, but can be overridden to supply a custom config.txt.
+      '';
+    };
+  };
+
+  config = {
+    hardware.raspberry-pi.configtxt.file = lib.mkDefault (toConfigTxt cfg.settings);
```
quentinmit: Why not set this as the `default` in the option? (The pattern appears to be to do so alongside a `defaultText` saying it's a generated file - see the options I referenced in my other comment.)

raspberry-pi/common/config-txt.nix
JamieMagee: Yeah, it was broken. Fixed so they drop out during flattening.

raspberry-pi/common/config-txt-defaults.nix
JamieMagee: Agreed, starting small. The list form works for now, and a structured overlay surface can go on top later without breaking anything.

raspberry-pi/common/config-txt.nix
JamieMagee: Fixed in 8fbb4ed. `types.path` with `pkgs.writeText` now, following the `configFile` pattern.

raspberry-pi/common/config-txt.nix
JamieMagee: Yep, moved it to the option default with `defaultText` in the same commit.

quentinmit (approved):
Looks great to me!

PR #1821, opened 2026-04-03
dell/precision/5820: init

Stunkymonkey:
###### Description of changes
add Dell Precision 5820 Tower

###### Things done

### DIFF
--- a/dell/precision/5820/default.nix
+++ b/dell/precision/5820/default.nix
@@ -1,0 +1,31 @@
+{
+  config,
+  lib,
+  ...
+}:
+{
+  imports = [
+    ../../../common/cpu/intel
+    ../../../common/pc
+    ../../../common/pc/ssd
+    ../../../common/gpu/nvidia
+  ];
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+  boot.initrd.availableKernelModules = [
+    "xhci_pci"
+    "ahci"
+    "vmd"
+    "nvme"
+  ];
+
+  hardware = {
+    nvidia = {
+      nvidiaSettings = lib.mkDefault true;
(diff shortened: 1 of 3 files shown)


dell/precision/5820/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  boot.initrd.availableKernelModules = [
+    "xhci_pci"
+    "ahci"
+    "vmd"
+    "nvme"
+    "usb_storage"
+    "usbhid"
+    "sd_mod"
+  ];
```
doronbehar, line 20: Do you  know to explain why is each of the following modules is needed?

- `usbhid`
- `usb_storage`
- `sd_mod`

dell/precision/5820/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  hardware = {
+    nvidia = {
+      nvidiaSettings = lib.mkDefault true;
+      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.legacy_580;
+      open = lib.mkDefault false;
+    };
+  };
+
+  services.thermald.enable = lib.mkDefault true;
```
doronbehar: Why enable this by default?

dell/precision/5820/default.nix
Stunkymonkey: No, I don't have a specific reason for these - they were carried over from the generated hardware-configuration.nix but aren't actually needed for this hardware. Removed all three in the latest commit, keeping only `xhci_pci`, `ahci`, `vmd`, and `nvme`.

dell/precision/5820/default.nix
Stunkymonkey: Fair point - thermald is mainly useful for laptops managing thermal throttling, and this is a Tower workstation with normal desktop cooling. Removed it in the latest commit.

PR #1825, opened 2026-04-10
Add Dependabot for nix flake updates

JamieMagee:
See https://github.blog/changelog/2026-04-07-dependabot-version-updates-now-support-the-nix-ecosystem/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


Mic92:
https://github.com/NixOS/nixos-hardware/blob/master/flake.nix we don't actually have flake inputs in the main flake. Will see if this works for sub flakes.

JamieMagee:
@Mic92 It won't work for sub flakes by default, but let me send out a follow-up PR to support that.

PR #1826, opened 2026-04-11
raspberry-pi-4: add `gpio-fan` overlay

matiasyocca:
###### Description of changes

Adds support for some PWM GPIO fans in a similar way to the official `gpio-fan` overlay. Tested with Argon Mini Fan.

Here's the instruction manual for the fan for reference: https://malnapc.hu/custom/malnapc/image/data/docs/RS/A700000007741984.pdf

###### Things done


doronbehar:
@JamieMagee do you think you could test this PR?

JamieMagee:
> @JamieMagee do you think you could test this PR?

I don't own this exact fan, but comparing with the upstream [`gpio-fan-overlay.dts`][1] the only changes are:

1. Compatibility changed from `"brcm,bcm2835"` (All Raspberry Pis) to `"brcm,bcm2711"` (Raspberry Pi 4+)
2. GPIO pin changed from 12 to 18

I assume both of these are due to the fixed pins on this fan only working with the exact layout of the Raspberry Pi 4, and forcing pin 18. In that regard, this looks correct.

I think I'd like to try and land a generic `hardware.raspberry-pi.overlays` mechanism that consumes the upstream `.dtbo`s with their  `__overrides__` parameters. With that, this overlay would already be supported by `gpio-fan.dtbo` and we'd avoid the duplication.

But I don't know how long that would take, so I'm not opposed to this PR. I'd maybe just make it clearer that the defaults are for the Argon case fan, and not something like [the official Raspberry Pi 4 fan][2]

[1]: https://github.com/raspberrypi/linux/blob/rpi-6.18.y/arch/arm/boot/dts/overlays/gpio-fan-overlay.dts
[2]: https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/

raspberry-pi/4/gpio-fan.nix
```diff
@@ -0,0 +1,91 @@
...
+  cfg = config.hardware.raspberry-pi."4".gpio-fan;
+in
+{
+  options.hardware = {
+    raspberry-pi."4".gpio-fan = {
+      enable = lib.mkEnableOption "support for Raspberry Pi style gpio-fan control";
+
+      pin = lib.mkOption {
+        type = lib.types.int;
+        default = 18;
```
doronbehar, line 13: @matiasyocca Maybe it'd be more correct to specify `12`? Or 14 maybe? Here's what I saw [here](https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/):

Image

raspberry-pi/4/gpio-fan.nix
```diff
@@ -0,0 +1,91 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.hardware.raspberry-pi."4".gpio-fan;
+in
+{
+  options.hardware = {
+    raspberry-pi."4".gpio-fan = {
+      enable = lib.mkEnableOption "support for Raspberry Pi style gpio-fan control";
```
doronbehar, line 9: Maybe:

```suggestion
      enable = lib.mkEnableOption "support for [Raspberry Pi gpio-fan](https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/) control";
```

JamieMagee:
FYI @matiasyocca I have started work on the generic `hardware.raspberry-pi.overlays` mechanism I mentioned above. I outlined it in #1946 and started the implementation in #1947. This is still fine IMO, and I can migrate it later.

matiasyocca:
> FYI @matiasyocca I have started work on the generic `hardware.raspberry-pi.overlays` mechanism I mentioned above. I outlined it in #1946 and started the implementation in #1947. This is still fine IMO, and I can migrate it later.

Oh, thanks for the swift response! I was just getting this up-to-date as it has been a while, so I'll take the chance to check your implementation as reference.

doronbehar:
@matiasyocca I don't think we want more in-tree `dtsText`s like added here. Are we sure that the more general solution available in https://github.com/NixOS/nixos-hardware/pull/1947 or even since #1788 can help you set gpio-fan overlay in a declarative manner?

doronbehar:
Closing because this should be supported with https://github.com/NixOS/nixos-hardware/pull/1947 .

PR #1748, opened 2026-01-26
Add lenovo-thinkpad-t14-amd-gen6

Seercat3160:
###### Description of changes

Add configuration for the Lenovo ThinkPad T14 AMD Gen 6, based on that of the Gen 5.

This device uses the mt7925 wireless adapter, so I've imported the relevant common module. My own experience suggests that importing the iwd module from that section makes matters worse, so I have not used that.

The `acpi.ec_no_wakeup=1` kernel param needed for gen5 no longer seems relevant as far as I can tell.

###### Things done

- [x] Tested the changes in your own NixOS Configuration
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

###### Additional notes

Closes #1742

PR #1827, opened 2026-04-12
fix: enable acpi.ec_no_wakeup in Thinkpad T14 AMD Gen 6

codingCoffee:
###### Description of changes

Add acpi.ec_no_wakeup=1 kernel param to stop the embedded controller from waking the system during s2idle. Fixes wake-on-plug for USB-C charger and peripheral hot-plug on ThinkPad T14 Gen 6 AMD.

This is the same fix required in Gen 5 of the Thinkpad T14 series.

###### Things done


codingCoffee:
@Mic92 can you check out this PR, I've tested it and it works on my Thinkpad T14 Gen 6. It's very similar to this [PR](https://github.com/NixOS/nixos-hardware/pull/1267)

m00wl:
In the table in the README, you link to the old gen5 directory (lenovo/thinkpad/t14/amd/gen5).
In addition, there is also #1748. They seem to not need `acpi.ec_no_wakeup=1` anymore.
I have access to a Gen6, can you tell me how to confirm that the parameter is needed?

codingCoffee:
I too have a Gen 6, let me check and confirm as well but in case you want to check in the meantime here are the steps
- suspend your laptop
- plug in either a charging cable in a usb-c port or plugin in a device (mobile etc) into a usb c port
- the laptop should wakeup from suspend withot the kernel param 

Maybe linux kernel 7.0.0 already solves this without the kernel param. I'll have to check

PR #1838, opened 2026-04-23
framework: Add package for official wallpapers

JohnAZoidberg:
###### Description of changes
See README

image

###### Things done

### DIFF
--- a/framework/12-inch/common/default.nix
+++ b/framework/12-inch/common/default.nix
@@ -5,6 +5,7 @@
     ../../../common/pc/ssd
     ../../kmod.nix
     ../../framework-tool.nix
+    ../../wallpapers.nix
   ];
 
   # Fix TRRS headphones missing a mic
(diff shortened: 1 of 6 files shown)


ctsdownloads:
@JohnAZoidberg Tested on a Framework 13. I tried building all five packs via hardware.framework.wallpapers.enableAll. Four build fine, but laptop16 fails with a hash mismatch:

```
specified: sha256-u+zWlIVZ/Gqk71F6r35Y1akLCwcVxcP/CzDZWcmsSU8=
actual:    sha256-3uMwoLOO+QmGvR2braFRE6cprBfidkRk8m8RN8G9//c=
```

The specified value is the hash for laptop16 in framework/wallpapers.nix; actual is what the zip currently serves. Repro:

`nix store prefetch-file --unpack https://downloads.frame.work/assets/framework-laptop16-wallpaper-pack.zip`

Looks like the asset changed since the hash was recorded, updating the laptop16 hash to that value should fix it. desktop/laptop12/laptop13/laptop13pro all match.

Two other things: should these declare a meta.license (frame.work/wallpapers states no redistribution terms)? And there's a merge conflict needing a rebase.

Unless I am missing something?

JohnAZoidberg:
you're right, will update the hash.
And check license internally

PR #1855, opened 2026-05-04
Init Raxda Rock3C

nappairam:
###### Description of changes

Add board definition for Radxa Rock3c. I did not have to do much, and just follow rock4c board convention.

Switch the default `rockchip` disko root filesystem (and `radxa.supportedFilesystems`) from `bcachefs` to `ext4`.

bcachefs was removed from mainline Linux 6.18. It now ships only as an out-of-tree module via `pkgs.bcachefs-tools.kernelModule`. With the previous default, `disko.imageBuilder` fails on 6.18+ because `kernel-modules-shrunk` cannot locate a `bcachefs` module in the kernel's module tree.

###### Things done

- [x] Built `nix build .#nixosConfigurations.<board>.config.system.build.diskoImages` on Linux 6.18 — succeeds with ext4 root.
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via flake input

Note that this PR depends on https://github.com/NixOS/nixpkgs/actions/runs/25323149751/job/74237102769?pr=516516 for rock 3c uboot image.


nappairam:
cc @RadxaYuntian @Mic92 - would appreciate review. 

@RadxaYuntian since this swaps the rockchip disko default you introduced (bcachefs -> ext4) due to 6.18 dropping bcachefs from mainline.

radxa/default.nix
```diff
@@ -14,11 +14,8 @@ in
 
   config = lib.mkIf cfg.enable {
     boot = {
-      # Currently enable bcachefs automatically set
-      # kernelPackages = lib.mkDefault pkgs.linuxPackages_latest;
       # TODO: Consider removing this line once, we get an LTS kernel that is newer than 6.12
       kernelPackages = lib.mkOverride 990 pkgs.linuxPackages_latest;
-      supportedFilesystems = [ "bcachefs" ];
```
RadxaYuntian, line 21: Will this break existing users when their image is built with the default Disko profile with `bcachefs` in use already?

radxa/README.md
```diff
@@ -22,7 +22,7 @@ definition of "support":
 ## Characteristic of the default profiles
 
 * Using `linuxPackages_latest`
-* `bcachefs` rootfs with zstd compression and password-less encryption enabled
+* `ext4` rootfs
```
RadxaYuntian: If we want to drop bcachefs I'd go with btrfs instead of ext4. Compression should help with slow IO on SBCs.

flake.nix
```diff
@@ -414,6 +414,7 @@
           raspberry-pi-3 = import ./raspberry-pi/3;
           raspberry-pi-4 = import ./raspberry-pi/4;
           raspberry-pi-5 = import ./raspberry-pi/5;
+          rock-3c = import ./radxa/rock-3c;
```
RadxaYuntian:
```suggestion
          radxa-rock-3c = import ./radxa/rock-3c;
```
We have some products not under ROCK brand, so future products should add `radxa-` prefix to sort them together.
Not sure if we can add alias for existing products.

nappairam: Sure. I will update the name

radxa/default.nix
nappairam: Yes probably. Will revert this change to support bcachefs

radxa/README.md
nappairam: Ack, I went with ext4 since it is rock solid fs and default in most distros.
But btrfs sounds good too and is maintained within kernel. Will update it.

radxa/default.nix
RadxaYuntian: I'm not sure if this change will affect Disko to match partition to boot in addition to bcachefs support itself. Need to check the code to see if they generate some kind of partition signature (including filesystem type), or just use GPT partition label to select mount point.

nappairam: Yes, disko generates fileSystems with fsType
https://github.com/nix-community/disko/blob/63b4e7e6cf75307c1d26ac3762b886b5b0247267/lib/types/filesystem.nix#L90

It will break existing sd image which try to upgrade nixos-hardware flake since filesystem mismatch.

The default bcachefs sdimage generation was failing in 6.18 kernel.
How do you like to proceed ?

1. We can move default to btrfs and document in changelog
2. Leave bcachefs as it is. it is currently breaking since it is not a regular module.

RadxaYuntian: I'll take a look at the sdimage generation issue first.

RadxaYuntian: bcachefs issue is related to nix-community/disko#1151. A workaround is to pin disko input to nix-community/disko#1159. I'd suggest solving it there.

nappairam: Okie, makes sense. I will remove this commit from my PR then.

RadxaYuntian (approved):

nappairam:
Thanks for the review @RadxaYuntian 


@Mic92 Please take a look once

nappairam:
@Mic92 Gentle reminder to review

PR to add radxa rock3c uboot package is already merged to nixpkgs nixos-unstable branch
https://github.com/NixOS/nixpkgs/pull/516516

So this PR is good to merge

PR #1858, opened 2026-05-05
framework/13-inch: fix enhanced audio device priority

Stebalien:
###### Description of changes

On framework laptops, ensure that the "enhanced audio" sink for the laptop speakers has a lower priority than bluetooth speakers (by giving it the same priority as the built-in laptop speakers, [1009](https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/7fa44ef8d0b2f603bd826d399718cef8cf6a800a/src/scripts/monitors/alsa.lua#L218)). Bluetooth speakers have a priority of [1010](https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/7fa44ef8d0b2f603bd826d399718cef8cf6a800a/src/scripts/monitors/bluez.lua#L275) by default.

NOTE: the existing comment in the code claiming that higher values mean lower priority is incorrect; lower priorities values mean lower priority.

###### Things done

After testing, my audio switches from my laptop speakers to my bluetooth headphones when I connect them.

### DIFF
--- a/framework/13-inch/common/classic-audio.nix
+++ b/framework/13-inch/common/classic-audio.nix
@@ -315,12 +315,11 @@ in
           "device.api" = "dsp";
           "node.virtual" = "false";
 
-          # Lower seems to mean "more preferred",
-          # bluetooth devices seem to be ~1000, speakers seem to be ~2000
-          # since this is between the two, bluetooth devices take over when they connect,
-          # and hand over to this instead of the speakers when they disconnect.
-          "priority.session" = 1500;
-          "priority.driver" = 1500;
+          # 1009 is the default priority of the "first" analog audio device.
+          # Higher priorities (e.g., 1010 for bluetooth and 1100 for usb audio)
+          # are preferred.
+          "priority.session" = 1009;
+          "priority.driver" = 1009;
           "state.default-volume" = 0.343;
           "device.icon-name" = "audio-card-analog-pci";
         };


JohnAZoidberg:
How did you figure out the numbers?

Stebalien:
1. `wpctl status` to get the devices.
2. `wpctl inspect DEV_NUM` to get the properties.

Then I grepped around in the wireplumber source code and found where the priorities were configured (see the links in the PR description). The bluez priority is always 1010 for sinks and 2010 for sources. The calculation for built-in audio devices is trickier, but it looks like it [should be a stable 1009](https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/7fa44ef8d0b2f603bd826d399718cef8cf6a800a/src/scripts/monitors/alsa.lua#L207-221) for Framework laptops, assuming it always shows up as the "first" device.

JohnAZoidberg:
Did you figure out where the default numbers come from? Is this actually hardware/firmware dependent?
You say the previous patch here was wrong, are the defaults also not okay?

Stebalien:
> Did you figure out where the default numbers come from? Is this actually hardware/firmware dependent?

The defaults come from the code I linked in the PR description.

- The default priority for bluetooth sinks is always 1010, and 2010 for sources.
- The default for "alsa" devices depends on the device number and device type. The formula defined [here](https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/7fa44ef8d0b2f603bd826d399718cef8cf6a800a/src/scripts/monitors/alsa.lua#L207-229), and the value for the built-in sound card should be 1009 unless other devices are present on boot. Unfortunately, I don't know how the kernel assigns audio device numbers, so I can't guarantee that the priority is stable if other audio devices are present.

> You say the previous patch here was wrong, are the defaults also not okay?

The [defaults picked by wireplumber](https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/7fa44ef8d0b2f603bd826d399718cef8cf6a800a/src/scripts/monitors/alsa.lua#L207-229) are sane. The priority (lowest to highest) is roughly:

1. Built-in audio sinks (lowest priority, ≤1009).
2. Bluetooth audio sinks (1010).
3. USB audio sinks (±1100).
4. "PRO" audio sinks (±1500).
5. Audio capture sinks (highest priority, ±2000).

The previous code was incorrect because:

1. It made the assumption "lower seems to mean 'more prefered'" (comment in the code). Higher priorities are prefered.
2. It gave the "enhanced" audio device a priority on-par with "PRO" audio sinks, much greater than bluetooth speakers/headphones.

JohnAZoidberg:
Thanks for the explanation. If it's purely software deciding what to use, then this wouldn't be a fix for a specific type of hardware, right?

Stebalien:
Ah, I see the confusion.

I'm fixing the priority of the virtual "enhanced audio device" _created_ in the modified file specifically (https://github.com/NixOS/nixos-hardware/pull/1253) for Framework 13 laptops.

The file I'm editing creates a virtual audio device to improve the sound quality on Framework 13 laptops. However, it assigns a priority higher than the default priority of bluetooth headphones, which clearly (@miketth can correct me if I'm wrong) wasn't the intended behavior given the comment in the code I replaced.

Stebalien:
Sorry to be a bother, but is there anything I can do to help move this along?

Without this change, the PipeWire fails to automatically switch from the built-in speakers to bluetooth speakers when connected (when the "enhanced audio" feature is enabled).

miketth (approved):

miketth:
Looks good from my end; I was struggling to find a good value for those, so the 1500 was mostly guesswork.

I'm not a maintainer though, so my blessing isn't too useful here.

Stebalien:
@JohnAZoidberg I'd really love to see this merged so I can get off my fork of this repo. I'm guessing there are a good number of users who are frustrated by this issue as well and likely have no idea why automatic switching from their laptop speakers to their bluetooth headphones isn't working.

rycee:
Did not test it myself but seems reasonable, if I'm overstepping feel free revert 🙂

PR #1841, opened 2026-04-24
raspberry-pi: add structuredExtraConfig for vendor kernel

JamieMagee:
The RPi vendor kernel is built from `bcm{rpi,2709,2711,2712}_defconfig`, but several values in nixpkgs `common-config.nix` override those defconfigs. Set them back via `structuredExtraConfig`.

Two commits:
1. Replace the `postConfigure` sed hack with `LOCALVERSION = freeform "";`.
2. Add the remaining vendor-defconfig overrides.

`mkForce` is used only where nixpkgs sets a conflicting value at normal priority; everything else is plain assignment.

| Config | Value | Forced? | Why |
|---|---|---|---|
| `LOCALVERSION` | `""` | no | vendor sets `-v7`/`-v8`/`-v8-16k`, breaks `modDirVersion` |
| `NR_CPUS` | `4` | yes | nixpkgs sets 384 |
| `CMA_SIZE_MBYTES` | `5` | yes | nixpkgs sets 32 |
| `PREEMPT` (Pi 3+) | yes | yes | nixpkgs sets `no`; arm64 vendor defconfigs use full preempt |
| `PREEMPT_VOLUNTARY` (Pi 3+) | no | no | nixpkgs only sets it via `mkOptionDefault` |
| `NFS_FS` | yes | yes | nixpkgs sets `module` |
| `NFS_V4`, `ROOT_NFS`, `IP_PNP_DHCP`, `IP_PNP_RARP` | yes | no | not set by nixpkgs |
| `IP_PNP` | yes | yes | nixpkgs sets `no` |
| `NET_CLS_BPF`, `NLS_CODEPAGE_437` | yes | yes | nixpkgs sets `module` |
| `FB_SIMPLE` | yes | no | only set by nixpkgs on <5.15 |

Ref: https://github.com/raspberrypi/linux/tree/rpi-6.12.y/arch/arm64/configs

###### Things done

- [ ] Tested the changes in your own NixOS Configuration
- [ ] Tested the changes end-to-end by using your fork of `nixos-hardware` and importing it via `` or Flake input

PR #1860, opened 2026-05-05
raspberry-pi: fix empty LOCALVERSION breaking modDirVersion

### LINKED ISSUE #1859: Error in modDirVersion check for Raspberry Pi kernel
My nightly NixOS upgrade on my Raspberry Pi failed tonight with:

`linux-rpi> Error: modDirVersion 6.12.75 specified in the Nix expression is wrong, it should be: 6.12.75""`

Inspecting the previously generated `'/build/source/build/.config' -> '/nix/store/0q2w22igh2jk0nssbx2s5p05an7g732q-linux-config-6.12.75-1+rpt1'`

reveals

`CONFIG_LOCALVERSION="\"\""`

It looks very likely to be caused by https://github.com/NixOS/nixos-hardware/pull/1841 , possibly in combination with the kernel update`raspberry-pi/common/kernel: 6.12.47-stable_20250916 -> 6.12.75-1+rpt1`. @JamieMagee 

nixos-hardware was updated from `2096f3f411ce46e88a79ae4eafcfc9df8ed41c61` to `df7783100babf59001340a7a874ba3824e441ecb`, and that was the only update to the system.

JamieMagee:
###### Description of changes

#1841 dropped the `postConfigure` sed in favour of `LOCALVERSION = freeform "";`, but `freeform ""` can't emit a real empty string through nixpkgs `mkValue`. It renders as the literal two-char string `""`, so `CONFIG_LOCALVERSION="\"\""`, modDirVersion becomes `6.12.75""`, and the build fails the sanity check.

This restores the original two-line sed in `postConfigure` (matching nixpkgs' own `linux-rpi.nix`) and drops the broken `LOCALVERSION` entry. All other vendor-defconfig overrides from #1841 (NR_CPUS, CMA_SIZE_MBYTES, NFS_*, PREEMPT, etc.) are kept.

Fixes #1859.

###### Things done

### DIFF
--- a/raspberry-pi/common/kernel.nix
+++ b/raspberry-pi/common/kernel.nix
@@ -54,10 +54,6 @@ in
     # Override nixpkgs common-config.nix defaults that conflict with the RPi vendor defconfigs.
     # See: https://github.com/raspberrypi/linux/tree/rpi-6.12.y/arch/arm64/configs
     structuredExtraConfig = {
-      # Vendor defconfigs set CONFIG_LOCALVERSION to -v7, -v8, -v8-16k, which
-      # breaks modDirVersion. Clear it.
-      LOCALVERSION = freeform "";
-
       # RPi has 4 cores; nixpkgs common-config sets 384
       NR_CPUS = lib.mkForce (freeform "4");
       # nixpkgs sets 32MB; RPi vendor defconfig uses 5MB
@@ -99,6 +95,13 @@ in
   // (args.argsOverride or { })
 )).overrideAttrs
   {
+    # TODO: Put CONFIG_LOCALVERSION in `structuredExtraConfig` above once this is resolved:
+    # https://github.com/NixOS/nixpkgs/issues/516936
+    postConfigure = ''
+      sed -i $buildRoot/.config -e 's/^CONFIG_LOCALVERSION=.*/CONFIG_LOCALVERSION=""/'
+      sed -i $buildRoot/include/config/auto.conf -e 's/^CONFIG_LOCALVERSION=.*/CONFIG_LOCALVERSION=""/'
+    '';
+
     # The vendor kernel uses different DTB names (bcm2708/bcm2709/bcm2710) than what
     # U-Boot expects (bcm2835/bcm2836/bcm2837). Starting with Pi 4, names match.
     # See: https://github.com/u-boot/u-boot/blob/master/board/raspberrypi/rpi/rpi.c


raspberry-pi/common/kernel.nix
```diff
@@ -53,11 +53,10 @@ in
 
     # Override nixpkgs common-config.nix defaults that conflict with the RPi vendor defconfigs.
     # See: https://github.com/raspberrypi/linux/tree/rpi-6.12.y/arch/arm64/configs
+    # Note: CONFIG_LOCALVERSION (-v7/-v8/-v8-16k in vendor defconfigs) is cleared
+    # via the postConfigure sed below; `freeform ""` cannot express an empty
+    # string through nixpkgs `mkValue`. See issue #1859.
     structuredExtraConfig = {
-      # Vendor defconfigs set CONFIG_LOCALVERSION to -v7, -v8, -v8-16k, which
-      # breaks modDirVersion. Clear it.
-      LOCALVERSION = freeform "";
```
doronbehar, line 59: What about `freeform "''"` or `freeform "\"\""`?

JamieMagee: Tried both, but neither works. `freeform "''"` produces `CONFIG_LOCALVERSION="''"` (two single-quote chars), and `freeform "\"\""` reproduces the current bug exactly: it's length 2 so it skips the `val == ""` branch, the default branch emits `LOCALVERSION ""`, and kconfig stores the two `"` chars verbatim.

The intermediate format has no quote handling, so no value placed there can become a real empty `CONFIG_*=""` in `.config`.

raspberry-pi/common/kernel.nix
```diff
@@ -53,11 +53,10 @@ in
 
     # Override nixpkgs common-config.nix defaults that conflict with the RPi vendor defconfigs.
     # See: https://github.com/raspberrypi/linux/tree/rpi-6.12.y/arch/arm64/configs
+    # Note: CONFIG_LOCALVERSION (-v7/-v8/-v8-16k in vendor defconfigs) is cleared
+    # via the postConfigure sed below; `freeform ""` cannot express an empty
+    # string through nixpkgs `mkValue`. See issue #1859.
```
doronbehar: I'd drop this comment.

raspberry-pi/common/kernel.nix
```diff
@@ -99,6 +98,15 @@ in
   // (args.argsOverride or { })
 )).overrideAttrs
   {
+    # The RPi vendor defconfigs set CONFIG_LOCALVERSION to -v7/-v8/-v8-16k,
+    # which breaks modDirVersion. Clear it here because `freeform ""` in
+    # structuredExtraConfig cannot emit a true empty string through nixpkgs'
+    # `mkValue` (it is rendered as the literal two-char string `""`).
```
doronbehar: And put here something short like:

```suggestion
    # NOTE: Put these in `structuredExtraConfig` above once this is resolved:
    # https://github.com/NixOS/nixpkgs/issues/516936
```

doronbehar (commented):
How exactly do you test this?

raspberry-pi/common/kernel.nix
doronbehar: `TODO` instead of `NOTE` could also work.

raspberry-pi/common/kernel.nix
doronbehar: > Tried both, but neither works. `freeform "''"` produces `CONFIG_LOCALVERSION="''"` (two single-quote chars), and `freeform "\"\""` reproduces the current bug exactly: it's length 2 so it skips the `val == ""` branch, the default branch emits `LOCALVERSION ""`, and kconfig stores the two `"` chars verbatim.
> 
> The intermediate format has no quote handling, so no value placed there can become a real empty `CONFIG_*=""` in `.config`.

Thanks for the explanation, and thanks for opening https://github.com/NixOS/nixpkgs/issues/516936 ! One more thing maybe worth trying is `null`.

JamieMagee:
@doronbehar I didn't test this thoroughly, as its just a revert commit.

hatch01:
I finally built it successfully after '18h52m7s'.

Issue #1770, opened 2026-02-19
Chromebook Audio fixes


boredsquirrel:
ChromeOS devices can make really nice coreboot devices. They work very well but the audio setup is strange.

There is a script that successfully fixed my Botenflex (lenovo flex 3) Chromebook, and supports a ton more.

Nixifying would be amazing.

https://github.com/WeirdTreeThing/chromebook-linux-audio

ferguscollins:
The solution to this for Intel chromebooks is to set `boot.kernelParams = [ "snd_intel_dspcfg.dsp_driver=1" ];` to use the old Intel HD Audio driver, along with using the modern pipewire stack instead of the old pulse audio.

```nix
boot.kernelParams = [
  "snd_intel_dspcfg.dsp_driver=1"
];

services.pipewire = {
  enable = true;
  alsa.enable = true;
  alsa.support32Bit = true; # optional for compatibility
  jack.enable = true; # optional for advanced audio control
  pulse.enable = true;
  wireplumber.enable = true;
};

security.rtkit.enable = true; # optional for low latency cpu access for audio, reduces audio glitches caused by late returns of requests

services.pulseaudio.enable = false; # explicitly disable the old audio stack to avoid conflict
```

That chromebook-linux-audio script essentially does the same as what I just listed above for Intel chromebooks, except mistakenly also installing the modern sof firmware (and some other firmware) for audio as well even though it is not compatible.

This is a problem that should exist on every intel chromebook on coreboot as far as I'm aware, but I am hesitant to generalise that far without some more confirmation.

If anyone else has an Intel chromebook with this issue then please leave a comment with whether or not setting `boot.kernelParams = [ "snd_intel_dspcfg.dsp_driver=1" ];` fixes it. I will make a configuration for the Asus Chromebox 4 Celeron with this line, but it can be converted to a generalised configuration for Intel chromebooks if needed.

lavglaab:
Hey! Chiming in from my Google Pixelbook Go (Atlas) that `"snd_intel_dspcfg.dsp_driver=1" ];` does not seem to have fixed my audio issues on NixOS 26.05. My GNOME desktop still shows no audio input devices, and only a Dummy Output. Here are the files I think are relevant to sound configuration on my system: https://gist.github.com/lavglaab/05fbc0ed0d661b019ae043c451c67bdc
[...]

PR #1886, opened 2026-05-22
init asus chromebox 4 duffy, celeron 5205u

ferguscollins:
###### Description of changes

Import comet lake cpu and gpu

Set sound driver to old Intel HD Audio driver. This does not enable sound, but if you enable pipewire then this is required to create sinks.

###### Things done

###### Note

This sound fix could be more generically applicable to Intel chromebooks on coreboot but I have no other machines to test. I have noted this in #1770.


ferguscollins:
Note: The fans only operate at 0 or 100% on coreboot, [there is a PR](https://github.com/MrChromebox/chrome-ec/pull/13#issuecomment-4506361900) open to fix this but in the meantime I have not yet found a good method for controlling the fans on nixos as none of the tools I have tried play nice with the Nix filesystem.

- fancontrol could work but pwmconfig cannot find any pwm sensors to generate a config
- nvramtool requires a cmos.layout that is not available for this board (board may not support CMOS tables, I have not researched this)
- [death7654/Chrultrabook-Tools](https://github.com/death7654/Chrultrabook-Tools/) uses ectool to manipulate chromebook features via DRA but I cannot get it to recognise any version of ectool that I install

Once that PR gets approved or I get a chance to test it out then I can better gauge if a custom fan profile is warranted or not.

PR #1890, opened 2026-05-25
Add: Dell Vostro 7590

tbaldwin-dev:
###### Description of changes
Add support for Dell Vostro 7590 Laptop

###### Things done

### DIFF
--- a/dell/vostro/7590/default.nix
+++ b/dell/vostro/7590/default.nix
@@ -1,0 +1,28 @@
+{ lib, ... }:
+
+{
+  imports = [
+    ../../../common/pc/laptop
+    ../../../common/pc/ssd
+    ../../../common/cpu/intel/coffee-lake
+    ../../../common/gpu/nvidia/turing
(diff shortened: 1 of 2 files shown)


Aron-dxd:
A few issues:

- There are hardcoded PCI Bus IDs. Use `lib.mkDefault` instead so that users can override it if their PCI topology is different
- I checked the spec sheet and found that the model was sold in [three GPU configurations](https://www.dell.com/support/manuals/en-in/vostro-15-7590-laptop/vos7590_setupspecs/video-card?guid=guid-98ce3663-9f41-4c8d-bd7a-c8871c45276a&lang=en-us): Intel UHD 630, GTX 1050 (Pascal), and GTX 1650 (Turing). 
Hardcoding `common/gpu/nvidia/turing` in the base file will break things for Intel-only models. The best way would be to make a `default.nix` with an Intel only base profile and then creating a sub-variant under `nvidia/turing` for all the card specific tuning. 
- Add the profile to `README.md` and also make changes to `flake.nix` to match the variant.

PR #1894, opened 2026-05-29
raspberry-pi: add firmware-partition install module

JamieMagee:
###### Description of changes

Adds `hardware.raspberry-pi.firmware` to stage the files the Pi firmware needs before Linux starts (GPU boot code, vendor DTBs and overlays, the rendered `config.txt`, optionally U-Boot) onto `/boot/firmware`.

Not a new boot method (boards keep using `generic-extlinux-compatible`) this just provides the files that path depends on. Exposes `populateCmd` for image builders and an opt-in activation script (off by default) for running systems.

DTB/overlay copy adapted from nvmd/nixos-raspberrypi.

###### Things done

### DIFF
--- a/raspberry-pi/README.md
+++ b/raspberry-pi/README.md
@@ -4,7 +4,7 @@ NixOS profiles and modules for Raspberry Pi boards.
 
 ## What's here
 
-- `common/` has the shared bits: the `linux-rpi` kernel build (vendor defconfig, matching firmware), the `config.txt` generation module, and a pinned wireless firmware.
+- `common/` has the shared bits: the `linux-rpi` kernel build (vendor defconfig, matching firmware), the `config.txt` generation module, a pinned wireless firmware, and the firmware-partition install module.
 - `2/`, `3/`, `4/`, `5/` are the board profiles. Each one picks the right kernel and kernel params. Pi 4 and 5 also set DT filters and the initrd modules they need.
 - The extra files under `4/` are opt-in toggles for Pi 4 hardware: audio, dwc2, GPIO, I2C, LEDs, the PoE HATs, touchscreens, and so on.
 
@@ -18,7 +18,28 @@ NixOS profiles and modules for Raspberry Pi boards.
 }
 ```
 
-These profiles assume the `generic-extlinux-compatible` bootloader (the NixOS module that writes an `extlinux.conf` for U-Boot to read), which is what aarch64 NixOS SD images use by default. There is no `boot.loader.raspberry-pi` module here. U-Boot itself still has to land on the boot partition somehow. Either your image builder does it, or you do. Nothing in here writes to `/boot/firmware/`. See [Current limits](#current-limits).
+These profiles assume the `generic-extlinux-compatible` bootloader (the NixOS module that writes an `extlinux.conf` for U-Boot to read), which is what aarch64 NixOS SD images use by default. There is no `boot.loader.raspberry-pi` module here. U-Boot and the GPU boot code still have to land on the firmware partition somehow: either your image builder does it, or you use the firmware install module below.
+
+## Firmware install
+
+`hardware.raspberry-pi.firmware` stages the files the Pi firmware needs before Linux starts onto the firmware partition (default `/boot/firmware`): GPU boot code (`bootcode.bin`, `start*.elf`, `fixup*.dat`), vendor device trees and overlays, the rendered `config.txt`, and optionally U-Boot. It is not a new boot method; it just supplies the files the existing `generic-extlinux-compatible` + U-Boot path needs.
+
+When you build an SD image, the module sets `sdImage.populateFirmwareCommands` itself, so the firmware partition is populated at build time with no extra configuration.
+
+For a running system, set `hardware.raspberry-pi.firmware.enable = true`. An activation script then repopulates the firmware partition on every `nixos-rebuild switch`. It is off by default.
+
+To chainload U-Boot from the firmware, enable `uboot.enable`. It copies `u-boot.bin` to the firmware partition and sets `config.txt`'s `kernel` line for you:
+
+```nix
+{
+  hardware.raspberry-pi.firmware = {
+    enable = true;
+    uboot.enable = true;
+  };
+}
+```
+
+`uboot.enable` defaults `uboot.package` to nixpkgs' `pkgs.ubootRaspberryPiAarch64`, built from the upstream `rpi_arm64_defconfig`, which covers every 64-bit board (Pi 3/4/5). For a 32-bit board, override `uboot.package` with the matching U-Boot build. On the Pi 5 this boots from SD, but U-Boot can't drive USB/PCIe/RP1 yet, so USB boot, NVMe boot, and a USB keyboard at the U-Boot prompt don't work until Linux takes over.
 
 ## `config.txt`
 
@@ -51,6 +72,6 @@ Top-level attrs are conditional sections (`all`, `pi4`, `pi5`, `cm4`, and so on)
 
 ## Current limits
 
-- No firmware install: Nothing writes `start*.elf`, `bootcode.bin`, `fixup*.dat`, vendor DTBs, overlays, or `config.txt` to `/boot/firmware/`. The rendered config.txt is exposed at `hardware.raspberry-pi.configtxt.file`, but nothing on disk reads it yet. You either rely on the SD-image populate step or stage those files yourself.
-- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over. So a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. Raspberry Pi OS has the GPU firmware load the kernel directly; we go through U-Boot so it reads `extlinux.conf`, which is what gives you the NixOS boot-generation menu and rollbacks. The firmware install module just stages the boot code and (optionally) U-Boot; it doesn't add a firmware-level direct-boot path. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over, so a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- Single pinned kernel: `common/kernel.nix` pins one `linux-rpi` version rather than matching each kernel to its firmware release.
 - No Pi 0/02/1 board profiles: `common/kernel.nix` accepts `rpiVersion = 1`, but there's no `0/`, `02/`, or `1/` directory wiring that kernel up into a profile you can import via `<nixos-hardware/raspberry-pi/...>`.
--- a/raspberry-pi/common/firmware.nix
+++ b/raspberry-pi/common/firmware.nix
@@ -1,0 +1,204 @@
+# Firmware-partition install for Raspberry Pi.
+#
+# Stages the files the Pi needs before Linux starts onto the firmware partition
+# (default /boot/firmware): GPU boot code (bootcode.bin, start*.elf, fixup*.dat),
+# vendor device trees and overlays, the rendered config.txt, and optionally
+# U-Boot. Not a new boot method: boards still use
+# boot.loader.generic-extlinux-compatible (U-Boot reads extlinux.conf); this just
+# provides the files that path needs, either at SD-image build time
+# (sdImage.populateFirmwareCommands, wired automatically) or on a running system
+# (the opt-in activation script).
+#
+# DTB/overlay copy adapted from nvmd/nixos-raspberrypi (MIT).
+
+{
+  lib,
+  config,
+  options,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.hardware.raspberry-pi.firmware;
+
+  # install-rpi-firmware <target-dir>
+  # Idempotent: copies via temp file + rename, and prunes stale DTBs/overlays.
+  installScript = pkgs.writeShellApplication {
+    name = "install-rpi-firmware";
+    runtimeInputs = [ pkgs.coreutils ];
+    text = ''
+      target="$1"
+      shopt -s nullglob
+
+      firmwareBoot=${cfg.package}/share/raspberrypi/boot
+      ${
+        if cfg.useGenerationDeviceTree then
+          ''
+            # Prefer the booted generation's device trees over the vendor ones.
+            dtbSrc=/run/current-system/dtbs
+            [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
+          ''
+        else
+          ''
+            dtbSrc=$firmwareBoot
+          ''
+      }
+
+      mkdir -p "$target/overlays"
+
+      # Copy via a temp file then rename so an interrupted run can't leave a
+      # half-written file behind. The firmware partition is FAT, so the rename
+      # isn't truly atomic, but it still beats a partial copy when the
+      # activation script rewrites the partition on a live system.
+      copyForced() {
+        cp "$1" "$2.tmp"
+        mv "$2.tmp" "$2"
+      }
+
+      # Track every file we copy this run, keyed by destination path, so the
+      # prune step below can delete stale device trees / overlays left behind
+      # by a previous generation.
+      declare -A kept
+
+      echo "rpi-firmware: copying device trees from $dtbSrc"
+      for dtb in "$dtbSrc"/*.dtb "$dtbSrc"/broadcom/*.dtb; do
+        dst="$target/$(basename "$dtb")"
+        copyForced "$dtb" "$dst"
+        kept[$dst]=1
+      done
+
+      if [ -d "$dtbSrc/overlays" ]; then
+        for ovr in "$dtbSrc"/overlays/*; do
+          dst="$target/overlays/$(basename "$ovr")"
+          copyForced "$ovr" "$dst"
+          kept[$dst]=1
+        done
+      fi
+
+      # Prune stale device trees / overlays.
+      for fn in "$target"/*.dtb "$target"/overlays/*; do
+        if [ "''${kept[$fn]:-}" != 1 ]; then
+          rm -v -- "$fn"
+        fi
+      done
+
+      echo "rpi-firmware: copying GPU boot code"
+      for src in "$firmwareBoot"/bootcode.bin "$firmwareBoot"/start*.elf "$firmwareBoot"/fixup*.dat; do
+        copyForced "$src" "$target/$(basename "$src")"
+      done
+
+      ${lib.optionalString cfg.uboot.enable ''
+        echo "rpi-firmware: copying U-Boot"
+        copyForced ${cfg.uboot.package}/u-boot.bin "$target/u-boot.bin"
+      ''}
+
+      echo "rpi-firmware: copying config.txt"
+      copyForced ${config.hardware.raspberry-pi.configtxt.file} "$target/config.txt"
+
+      echo "rpi-firmware: done ($target)"
+    '';
+  };
+in
+{
+  options.hardware.raspberry-pi.firmware = {
+    enable = lib.mkEnableOption ''
+      installation of the Raspberry Pi firmware partition on a running system.
+
+      An activation script repopulates {option}`hardware.raspberry-pi.firmware.path`
+      on every system switch
+    '';
+
+    path = lib.mkOption {
+      type = lib.types.str;
+      default = "/boot/firmware";
+      description = ''
+        Mount point of the Raspberry Pi firmware (FAT) partition.
+
+        `/boot/firmware` matches the NixOS aarch64 SD-image layout, and most
+        configurations should leave it there. The activation script writes here
+        only when it is a mounted partition (checked with `mountpoint`);
+        otherwise it logs a warning and skips.
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.raspberrypifw;
+      defaultText = lib.literalExpression "pkgs.raspberrypifw";
+      description = ''
+        Package providing the Raspberry Pi GPU boot code, vendor device trees,
+        and overlays under `''${package}/share/raspberrypi/boot`.
+      '';
+    };
+
+    uboot = {
+      enable = lib.mkEnableOption ''
+        chainloading U-Boot from the Raspberry Pi firmware.
+
+        Copies `u-boot.bin` from
+        {option}`hardware.raspberry-pi.firmware.uboot.package` to the firmware
+        partition and points `config.txt`'s `kernel` at it, so the GPU firmware
+        loads U-Boot, which then reads `extlinux.conf`
+      '';
+
+      package = lib.mkOption {
+        type = lib.types.package;
+        default = pkgs.ubootRaspberryPiAarch64;
+        defaultText = lib.literalExpression "pkgs.ubootRaspberryPiAarch64";
+        description = ''
+          U-Boot package whose `u-boot.bin` is copied to the firmware
+          partition when {option}`hardware.raspberry-pi.firmware.uboot.enable`
+          is enabled.
+
+          The default, nixpkgs' `pkgs.ubootRaspberryPiAarch64`, covers the
+          64-bit boards (Pi 3/4/5). For a 32-bit board, override this with the
+          matching U-Boot package.
+        '';
+      };
+    };
+
+    useGenerationDeviceTree = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Copy device trees from the booted NixOS generation
+        (`/run/current-system/dtbs`) instead of the vendor firmware package.
+        Irrelevant when generating an SD image.
+      '';
+    };
+  };
+
+  config = lib.mkMerge [
+    (lib.mkIf cfg.uboot.enable {
+      # Chainload U-Boot: the GPU firmware loads u-boot.bin, which then reads
+      # extlinux.conf. mkDefault so an explicit kernel setting still wins.
+      hardware.raspberry-pi.configtxt.settings.all = {
+        kernel = lib.mkDefault "u-boot.bin";
+        # Default U-Boot is 64-bit, but the firmware loads kernel= in 32-bit
+        # mode unless arm_64bit=1.
+        arm_64bit = lib.mkDefault pkgs.stdenv.hostPlatform.isAarch64;
+      };
+
+      # The GPU firmware merges config.txt dtoverlays into the DTB it hands to
+      # U-Boot. The default (true) adds an FDTDIR line to extlinux.conf, so
+      # U-Boot reloads bare dtbs and drops the overlays.
+      boot.loader.generic-extlinux-compatible.useGenerationDeviceTree = lib.mkDefault false;
+    })
+    # Stage the firmware partition at SD-image build time, only when an
+    # sd-image module is imported. mkForce so we override (not merge with)
+    # sd-image-aarch64.nix, which also sets this and would clobber config.txt.
+    (lib.optionalAttrs (options ? sdImage) {
+      sdImage.populateFirmwareCommands = lib.mkForce "${lib.getExe installScript} ./firmware\n";
+    })
+    (lib.mkIf cfg.enable {
+      system.activationScripts.raspberry-pi-firmware = lib.stringAfter [ "specialfs" ] ''
+        if mountpoint -q ${lib.escapeShellArg cfg.path}; then
+          ${lib.getExe installScript} ${lib.escapeShellArg cfg.path}
+        else
+          echo "rpi-firmware: ${cfg.path} is not a mounted partition, skipping firmware install" >&2
+        fi
+      '';
+    })
+  ];
+}
--- a/raspberry-pi/common/default.nix
+++ b/raspberry-pi/common/default.nix
@@ -2,6 +2,7 @@
   imports = [
     ./config-txt.nix
     ./config-txt-defaults.nix
+    ./firmware.nix
   ];
 
   boot.initrd.availableKernelModules = [


JamieMagee:
@doronbehar If you have some time, could you please review this change? Thank you!

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+        dtbSrc=$firmwareBoot
+      ''}
+
+      mkdir -p "$target" "$target/overlays"
+
+      copyForced() {
+        # copy src dst
+        cp "$1" "$2.tmp"
+        mv "$2.tmp" "$2"
+      }
```
doronbehar, line 57: Why is this needed?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+        [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
+      ''}
+      ${lib.optionalString (!cfg.useGenerationDeviceTree) ''
+        dtbSrc=$firmwareBoot
+      ''}
+
+      mkdir -p "$target" "$target/overlays"
+
+      copyForced() {
+        # copy src dst
```
doronbehar: Useless LLM comment.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+        cp "$1" "$2.tmp"
+        mv "$2.tmp" "$2"
+      }
+
+      declare -A kept
+
+      echo "rpi-firmware: copying device trees from $dtbSrc"
+      for dtb in "$dtbSrc"/*.dtb "$dtbSrc"/broadcom/*.dtb; do
+        [ -e "$dtb" ] || continue
```
doronbehar: Useless LLM check: If the file wouldn't have existed it wouldn't have been iterated in the above loop.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+          [ -e "$ovr" ] || continue
+          dst="$target/overlays/$(basename "$ovr")"
+          copyForced "$ovr" "$dst"
+          kept[$dst]=1
+        done
+      fi
+
+      # Prune stale device trees / overlays.
+      for fn in "$target"/*.dtb "$target"/overlays/*; do
+        [ -e "$fn" ] || continue
```
doronbehar: Same here...

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+  installScript = pkgs.writeShellApplication {
+    name = "install-rpi-firmware";
+    runtimeInputs = [ pkgs.coreutils ];
+    text = ''
+      target="''${1:-${cfg.path}}"
+
+      firmwareBoot=${cfg.package}/share/raspberrypi/boot
+      ${lib.optionalString cfg.useGenerationDeviceTree ''
+        # Prefer the booted generation's device trees over the vendor ones.
+        dtbSrc=$(readlink -f /run/current-system/dtbs 2>/dev/null || true)
```
doronbehar: Why is `readlink -f` so important? Also, I don't like these LLM typical `2>/dev/null || true` endings... Don't we want to be sure this directory exists, and if not expose an appropriate error?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+        [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
```
doronbehar: Same here - why can't we anticipate whether `$firmwareBoot` should be used or not?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+      ''}
+      ${lib.optionalString (!cfg.useGenerationDeviceTree) ''
+        dtbSrc=$firmwareBoot
+      ''}
+
+      mkdir -p "$target" "$target/overlays"
```
doronbehar: Another LLM bad code. This should be enough:

```suggestion
      mkdir -p "$target/overlays"
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+          copyForced "$ovr" "$dst"
+          kept[$dst]=1
+        done
+      fi
+
+      # Prune stale device trees / overlays.
+      for fn in "$target"/*.dtb "$target"/overlays/*; do
+        [ -e "$fn" ] || continue
+        if [ "''${kept[$fn]:-}" != 1 ]; then
+          rm -vf -- "$fn"
```
doronbehar: Why is `-f` needed here? Aren't we sure it will exist if it was `kept`?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,177 @@
...
+        fi
+      done
+
+      echo "rpi-firmware: copying GPU boot code"
+      for src in "$firmwareBoot"/bootcode.bin "$firmwareBoot"/start*.elf "$firmwareBoot"/fixup*.dat; do
+        [ -e "$src" ] || continue
```
doronbehar: Same here, we should be able to anticipate whether these files should exist or not, and not rely upon these `|| continue` suffixes...

doronbehar (changes requested):
Haven't read yet the docs, too many Bash code related comments. I did notice though, that the phrasing of the commit message and the PR description are pretty sloppy.

raspberry-pi/common/firmware.nix
JamieMagee: Done.

raspberry-pi/common/firmware.nix
JamieMagee: Removed.

raspberry-pi/common/firmware.nix
JamieMagee: That's only true if `nullglob` is set (which it wasn't). So on no match the loop ran once with the literal `*.dtb` pattern. I've set `shopt -s nullglob` at the top of the script and removed the guard here, plus the other ones lower down.

raspberry-pi/common/firmware.nix
JamieMagee: Same as above. Set `shopt -s nullglob`.

raspberry-pi/common/firmware.nix
JamieMagee: Yeah, removed.

raspberry-pi/common/firmware.nix
JamieMagee: Yeah, `nullglob` covers this too.

raspberry-pi/common/firmware.nix
JamieMagee: Atomic-ish replacement. The activation script can rewrite the firmware partition on a live system. It's a FAT partition so rename isn't an atomic, but it still beats a partial copy.

raspberry-pi/common/firmware.nix
JamieMagee: Replied above

raspberry-pi/common/firmware.nix
JamieMagee: I removed it and the `2>/dev/null || true` too.

The `[ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot` that's left isn't swallowing an error. When `useGenerationDeviceTree` is on we prefer the generation's DTBs, but that directory can legitimately not exist (a generation that builds no DTBs), and falling back to the vendor firmware DTBs keeps the board bootable.

raspberry-pi/README.md
```diff
@@ -18,7 +18,37 @@ NixOS profiles and modules for Raspberry Pi boards.
...
+These profiles assume the `generic-extlinux-compatible` bootloader (the NixOS module that writes an `extlinux.conf` for U-Boot to read), which is what aarch64 NixOS SD images use by default. There is no `boot.loader.raspberry-pi` module here. U-Boot and the GPU boot code still have to land on the firmware partition somehow: either your image builder does it, or you use the firmware install module below.
+
+## Firmware install
+
+`hardware.raspberry-pi.firmware` stages the files the Pi firmware needs before Linux starts onto the firmware partition (default `/boot/firmware`): GPU boot code (`bootcode.bin`, `start*.elf`, `fixup*.dat`), vendor device trees and overlays, the rendered `config.txt`, and optionally U-Boot. It is not a new boot method; it just supplies the files the existing `generic-extlinux-compatible` + U-Boot path needs.
+
+Two entry points:
```
doronbehar: What do you mean by entry point here? Is it something that one can `nix eval` and have a look at?

raspberry-pi/README.md
```diff
@@ -18,7 +18,37 @@ NixOS profiles and modules for Raspberry Pi boards.
...
+
+- `hardware.raspberry-pi.firmware.populateCmd`: populates a directory passed as its first argument. Use it from an image builder:
+
+  ```nix
+  {
+    sdImage.populateFirmwareCommands = "${config.hardware.raspberry-pi.firmware.populateCmd} ./firmware";
+  }
+  ```
+
+- `hardware.raspberry-pi.firmware.enable`: opt-in activation script that repopulates the firmware partition on every system switch. Off by default.
```
doronbehar: Nit:

```suggestion
- `hardware.raspberry-pi.firmware.enable`: opt-in activation script that repopulates the firmware partition on every system switch. `false` by default.
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+    };
+
+    ubootPackage = lib.mkOption {
+      type = lib.types.nullOr lib.types.package;
+      default = null;
+      example = lib.literalExpression "pkgs.ubootRaspberryPiAarch64";
+      description = ''
+        Optional U-Boot package. When set, its `u-boot.bin` is copied to the
+        firmware partition. Point `config.txt`'s `kernel` at `u-boot.bin` to
+        chainload it (firmware loads U-Boot, which then reads `extlinux.conf`).
```
doronbehar: Is there a situation where a user will want to use a u-boot, will set this option, and will not put in `config.txt` the `kernel=u-boot.bin` line? If not, maybe we should enable this behavior unconditionally, and only mention this happens here in the docs.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+
+        For 64-bit boards (Pi 3/4/5) use `pkgs.ubootRaspberryPiAarch64`.
```
doronbehar: And for 32-bit boards?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+      '';
+    };
+
+    useGenerationDeviceTree = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Copy device trees from the booted NixOS generation
+        (`/run/current-system/dtbs`) instead of the vendor firmware package.
+        Only meaningful for the runtime activation script.
```
doronbehar: I guess when you write `runtime activation script` you mean a situation where someone runs `nixos-rebuild` v.s generating an SD image? If so, maybe make that clearer with the negative:

```suggestion
        Irrelevant when generating an SD image.
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+
+    installScript = lib.mkOption {
+      type = lib.types.package;
+      readOnly = true;
+      internal = true;
+      default = installScript;
+      description = ''
+        Script that populates the firmware partition. Takes the target
+        directory as its first argument (defaults to
+        {option}`hardware.raspberry-pi.firmware.path`).
```
doronbehar: Let's make it clear that the default refers to the target directory of the script and not the script itself.

```suggestion
        Script that populates the firmware partition. Takes the target
        directory (defaults to {option}`hardware.raspberry-pi.firmware.path`) as its first argument.
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+      readOnly = true;
+      default = lib.getExe installScript;
+      defaultText = lib.literalExpression "<firmware install script>";
+      description = ''
+        Command that populates a firmware partition directory given as its
+        first argument. Intended for use in image builders, e.g.
+        `sdImage.populateFirmwareCommands = "''${config.hardware.raspberry-pi.firmware.populateCmd} ./firmware";`.
+      '';
+    };
+  };
```
doronbehar: Why do we need this switch? Don't we want to create an identity between sd images builds and regular `nixos-rebuild` configurations?

raspberry-pi/README.md
```diff
@@ -18,7 +18,37 @@ NixOS profiles and modules for Raspberry Pi boards.
...
+
+To chainload U-Boot from the firmware, set a U-Boot package and point `config.txt` at it:
+
+```nix
+{
+  hardware.raspberry-pi.firmware = {
+    enable = true;
+    ubootPackage = pkgs.ubootRaspberryPiAarch64;
```
doronbehar: Maybe we should just have a `useUboot = lib.mkEnableOption`.

raspberry-pi/README.md
```diff
@@ -18,7 +18,37 @@ NixOS profiles and modules for Raspberry Pi boards.
...
+
+`hardware.raspberry-pi.firmware` stages the files the Pi firmware needs before Linux starts onto the firmware partition (default `/boot/firmware`): GPU boot code (`bootcode.bin`, `start*.elf`, `fixup*.dat`), vendor device trees and overlays, the rendered `config.txt`, and optionally U-Boot. It is not a new boot method; it just supplies the files the existing `generic-extlinux-compatible` + U-Boot path needs.
+
+Two entry points:
+
+- `hardware.raspberry-pi.firmware.populateCmd`: populates a directory passed as its first argument. Use it from an image builder:
+
+  ```nix
+  {
+    sdImage.populateFirmwareCommands = "${config.hardware.raspberry-pi.firmware.populateCmd} ./firmware";
```
doronbehar: Maybe this should be set by default, and if someone builds an sd image it should just run automatically.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.raspberrypifw;
+      defaultText = lib.literalExpression "pkgs.raspberrypifw";
+      description = ''
+        Package providing the Raspberry Pi GPU boot code, vendor device trees,
+        and overlays under `share/raspberrypi/boot`.
```
doronbehar: To make it clear that the package's `$out/share` is meant here (I hope this renders OK):

```suggestion
        and overlays under `''${package}/share/raspberrypi/boot`.
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+      An activation script repopulates {option}`hardware.raspberry-pi.firmware.path`
+      on every system switch. Leave off if your image builder already handles it'';
+
+    path = lib.mkOption {
+      type = lib.types.str;
+      default = "/boot/firmware";
+      description = ''
+        Mount point of the Raspberry Pi firmware (FAT) partition.
+      '';
+    };
```
doronbehar, line 123: What would be a reason to override this?

raspberry-pi/common/firmware.nix
doronbehar: > Atomic-ish replacement. The activation script can rewrite the firmware partition on a live system. It's a FAT partition so rename isn't an atomic, but it still beats a partial copy.

Please write that in a comment, to spare someone trying to figure it out while it is committed.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,173 @@
...
+let
+  cfg = config.hardware.raspberry-pi.firmware;
+
+  # install-rpi-firmware <target-dir>
+  # Idempotent: copies via temp file + rename, and prunes stale DTBs/overlays.
+  installScript = pkgs.writeShellApplication {
+    name = "install-rpi-firmware";
+    runtimeInputs = [ pkgs.coreutils ];
+    text = ''
+      target="''${1:-${cfg.path}}"
```
doronbehar: This too can be simplified IMO - since we are mixing Nix & Bash, and this script is meant to be run only within a Nix build, maybe we can be sure the 1st argument is `${cfg.path}`? You mention this in the `path` option's doc too I think.

raspberry-pi/README.md
```diff
@@ -51,6 +81,7 @@ Top-level attrs are conditional sections (`all`, `pi4`, `pi5`, `cm4`, and so on)
...
-- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over. So a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. The firmware install module stages the boot code and (optionally) U-Boot, but it does not implement direct/generational kernel boot. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over, so a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- No NVMe boot on Pi 5: it needs BCM2712 PCIe support in U-Boot, which is still [in review upstream](https://lists.denx.de/pipermail/u-boot/2026-May/619653.html) and not in a tagged release. `ubootRaspberryPiAarch64` picks it up once nixpkgs ships a U-Boot that has it.
```
doronbehar: Not necessarily relevant to add to this poor README line, but just out of curiosity, how is our setup different from the what happens in the Debian system distributed by Raspberry?

doronbehar (commented):
Hey thanks for the Bash improvements! A few more comments now :).

raspberry-pi/README.md
JamieMagee: Just the two supported ways in: `populateCmd` is a read-only option holding the install script's path (for an image builder), and `enable` turns on the activation script for a running system. I can say "Two ways to use it" if that's clearer?

raspberry-pi/common/firmware.nix
JamieMagee: Yeah, the default only covers 64-bit. `useUboot` points `ubootPackage` at the aarch64 build (Pi 3/4/5). In a 32-bit board you'd override it with the matching U-Boot. I didn't try to auto-pick because the correct package depends on the board.

raspberry-pi/README.md
JamieMagee: Raspberry Pi OS lets the GPU firmware load the kernel directly. We keep `generic-extlinux-compatible` + U-Boot so generational boot still works. Is that what you mean?

raspberry-pi/common/firmware.nix
JamieMagee: It isn't always `${cfg.path}`. Image builders call `populateCmd ./firmware`, so the script needs the target dir as an argument. The `:-${cfg.path}` is just a fallback for the activation-script path. Would you prefer me to drop it and make it an explicit arg instead?

raspberry-pi/README.md
JamieMagee: I'd keep it opt-in, at least initially. On by default rewrites `/boot/firmware` on every `nixos-rebuild switch`, which is a nasty surprise when something else owns that partition. SD images still get the files anyway, since they call `populateCmd` from `sdImage.populateFirmwareCommands` at build time.

raspberry-pi/common/firmware.nix
JamieMagee: The default `/boot/firmware` matches the NixOS SD-image layout, so most people won't touch it. But if your firmware partition is mounted somewhere else, for example straight at `/boot` on a single-partition setup, or a custom mount point.

raspberry-pi/common/firmware.nix
JamieMagee: There's only one mechanism here. `populateCmd` and the `enable` activation script both call the same `installScript`, so an SD image and a `nixos-rebuild` get the same firmware partition. The two entry points only differ in where they run: an image builder wants a plain command for `sdImage.populateFirmwareCommands` (build sandbox, target dir passed as an arg), and a running system needs an activation script that writes to the mounted `path`. The output is the same.

raspberry-pi/common/firmware.nix
doronbehar: How would you compare this package to `uboot[A-Z]` packages from Nixpkgs? Are there overlays applied in `nixos-hardware` that are crucial for this to work?

raspberry-pi/README.md
doronbehar: > Raspberry Pi OS lets the GPU firmware load the kernel directly. We keep `generic-extlinux-compatible` + U-Boot so generational boot still works. Is that what you mean?

I think it does. Maybe the detail of our desire for generational boots would add a lot to this bullet item.

raspberry-pi/common/firmware.nix
doronbehar: The question is, why would we need a fallback for the activation script? Doesn't it always call this command with the `cfg.path` as argument? In that case, maybe simply `target="$1"` is enough, for both `populateCmd` and the activation script.

raspberry-pi/common/firmware.nix
doronbehar: > The default `/boot/firmware` matches the NixOS SD-image layout, so most people won't touch it. But if your firmware partition is mounted somewhere else, for example straight at `/boot` on a single-partition setup, or a custom mount point.

So this option should enable e.g single-partition setups to all of a sudden switch to the this new NixOS module and start manage their `config.txt` decoratively? If so, maybe it'd be nice to mention this in the `description`? I am also hopeful that there could be better ways to handle such existing setups: Maybe we should not make an effort to support old setups like this, or maybe we should be able to detect that during eval time? Or in the activation script?

raspberry-pi/README.md
doronbehar: > SD images still get the files anyway, since they call `populateCmd` from `sdImage.populateFirmwareCommands` at build time.

This will happen anyway if one sets this `sdImage.populateFirmwareCommands`?

> On by default rewrites `/boot/firmware` on every `nixos-rebuild switch`, which is a nasty surprise when something else owns that partition.

Why would an `sdImage` setting affect `nixos-rebuild switch` behavior? I thought activation scripts are not affected by `sdImage.` settings.

raspberry-pi/common/firmware.nix
doronbehar: > There's only one mechanism here. `populateCmd` and the `enable` activation script both call the same `installScript`, so an SD image and a `nixos-rebuild` get the same firmware partition. The two entry points only differ in where they run: an image builder wants a plain command for `sdImage.populateFirmwareCommands` (build sandbox, target dir passed as an arg), and a running system needs an activation script that writes to the mounted `path`. The output is the same.

I understand that the output is the same because the `installScript` is the same, but again I don't understand why would someone want to set `populateCmd` to something else other then this? If you ask me, I think this NixOS module should unconditionally set:

```nix
sdImage.populateFirmwareCommands = "${installScript} ./firmware";
```

And delete this `populateCmd` option. Setting any `sdImage.` option shouldn't affect a running system if I understand correctly.

raspberry-pi/README.md
doronbehar: The answer to this thread depends on your replies to my follow-up comments.

raspberry-pi/common/firmware.nix
JamieMagee: Done, dropped `populateCmd` and the module now sets `sdImage.populateFirmwareCommands` itself.

One tweak I had to make is guarding it with `lib.mkIf (options ? sdImage)`, because firmware.nix is imported by every board profile, and the `sdImage` options only exist once an SD-image module is loaded. Without the guard `nixos-rebuild` fails with "option `sdImage.populateFirmwareCommands' does not exist".

Does this match what you were after?

raspberry-pi/common/firmware.nix
JamieMagee: Yeah, fair enough. Simplified to just `target=$1`.

raspberry-pi/common/firmware.nix
JamieMagee: `path` isn't meant as a migration route for existing installs to take over `config.txt`. It's just the mount point of the firmware partition that the activation script writes to and the SD image populates. `/boot/firmware` is the NixOS default, but that mount point isn't universal. I'm not trying to support arbitrary layouts.

The activation script already guards at runtime with `mountpoint -q $path`. If it isn't a mounted partition it logs a warning and skips, so it won't write into a directory on the root filesystem.

I'd steer clear of an eval-time check though. The firmware partition can come from `fileSystems`, disko, or be set up out of band, so asserting on `config.fileSystems` would reject valid setups. I've reworded the path description to say it's the mount point, the default matches the SD-image layout, and most people shouldn't touch it. Is that okay, or do you want path constrained further?

raspberry-pi/README.md
JamieMagee: Can do.

raspberry-pi/README.md
JamieMagee: Removed this section, so it's moot now.

raspberry-pi/common/firmware.nix
JamieMagee: `ubootRaspberryPiAarch64` is u-boot's generic 64-bit rpi config. It supports Rpi 3/4/5 and auto-detects at runtime. The other packages (`ubootRaspberryPi3_64bit`, `ubootRaspberryPi4_64bit`, etc.) are all per-board images. There are no overlays in nixos-hardware, we're using the plain nixpkgs packages.

raspberry-pi/README.md
JamieMagee: Sorry, I think I got the two things muddled here. I meant that `enable` runs on every `nixos-rebuild switch`, and I was arguing for keeping it off by default.

I've also removed `populateCmd`, so this is a moot point.

raspberry-pi/common/firmware.nix
doronbehar: > Does this match what you were after?

Yes, thanks!

raspberry-pi/common/firmware.nix
doronbehar: > `ubootRaspberryPiAarch64` is u-boot's generic 64-bit rpi config. It supports Rpi 3/4/5 and auto-detects at runtime. The other packages (`ubootRaspberryPi3_64bit`, `ubootRaspberryPi4_64bit`, etc.) are all per-board images. There are no overlays in nixos-hardware, we're using the plain nixpkgs packages.

OK good. from some reason I wasn't sure when I wrote that line that these are plain Nixpkgs packages. Maybe it'd be nice if you'd write:

> Nixpkgs' `pkgs.uBoot...`

To make it clear that the nixos-hardware module is not overriding or inventing this? Not super important though.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,190 @@
...
+      # Copy via a temp file then rename so an interrupted run can't leave a
+      # half-written file behind. The firmware partition is FAT, so the rename
+      # isn't truly atomic, but it still beats a partial copy when the
+      # activation script rewrites the partition on a live system.
+      copyForced() {
+        cp "$1" "$2.tmp"
+        mv "$2.tmp" "$2"
+      }
+
+      declare -A kept
```
doronbehar, line 62: Could you maybe explain in a Bash comment what is the purpose of this array? I am still not sure I fully understood it.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,190 @@
...
+
+      firmwareBoot=${cfg.package}/share/raspberrypi/boot
+      ${lib.optionalString cfg.useGenerationDeviceTree ''
+        # Prefer the booted generation's device trees over the vendor ones.
+        dtbSrc=/run/current-system/dtbs
+        [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
+      ''}
+      ${lib.optionalString (!cfg.useGenerationDeviceTree) ''
+        dtbSrc=$firmwareBoot
+      ''}
```
doronbehar: Would be more readable IMO with an `if-else`:


```suggestion
    ''
    + if cfg.useGenerationDeviceTree then ''
      # Prefer the booted generation's device trees over the vendor ones.
      dtbSrc=/run/current-system/dtbs
      [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
    '' else ''
      dtbSrc=$firmwareBoot
    ''
```

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,190 @@
...
+    '';
+  };
+in
+{
+  options.hardware.raspberry-pi.firmware = {
+    enable = lib.mkEnableOption ''
+      installation of the Raspberry Pi firmware partition on a running system.
+
+      An activation script repopulates {option}`hardware.raspberry-pi.firmware.path`
+      on every system switch. Leave off if your image builder already handles it'';
```
doronbehar: Should have been forced by `nixfmt` I guess:

```suggestion
      on every system switch. Leave off if your image builder already handles it
    '';
```

doronbehar: Do you really think there are people who handle this by themselves in their image builders? Don't these kind of advanced users should naturally realize they should not enable it?

doronbehar: Also, I wrote about that earlier:


```suggestion
      on every system switch. Leave `false` if your image builder already handles it'';
```

raspberry-pi/common/firmware.nix
doronbehar: The documentation of this option is clear and it the existence of it is acceptable IMO now :)

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,190 @@
...
+      # Chainload U-Boot: the GPU firmware loads u-boot.bin, which then reads
+      # extlinux.conf. mkDefault so an explicit kernel setting still wins.
+      hardware.raspberry-pi.configtxt.settings.all.kernel = lib.mkDefault "u-boot.bin";
+    })
+    # Set only when an SD-image module is imported (it declares `sdImage`).
+    (lib.optionalAttrs (options ? sdImage) {
+      sdImage.populateFirmwareCommands = "${lib.getExe installScript} ./firmware";
+    })
+    (lib.mkIf cfg.enable {
+      system.activationScripts.raspberry-pi-firmware = lib.stringAfter [ "specialfs" ] ''
```
doronbehar, line 195: I'm not super familiar with the modules system, so could you explain maybe only to me the `stringAfter`? What is this `specialfs`?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,190 @@
...
+
+    useUboot = lib.mkEnableOption ''
+      chainloading U-Boot from the Raspberry Pi firmware.
+
+      Copies `u-boot.bin` from
+      {option}`hardware.raspberry-pi.firmware.ubootPackage` to the firmware
+      partition and points `config.txt`'s `kernel` at it, so the GPU firmware
+      loads U-Boot, which then reads `extlinux.conf`'';
+
+    ubootPackage = lib.mkOption {
```
doronbehar: What do you think of the idea to use instead `uboot.{enable,package}` options?

doronbehar (changes requested):
Looking much better, thanks for keeping work on this! Sorry for a slightly later reply this time. Left are only easy semantic review requests.

raspberry-pi/common/firmware.nix
JamieMagee: Done. Restructured to `uboot.{enable,package}`, and the package description now spells out that it's nixpkgs stock `pkgs.ubootRaspberryPiAarch64` with no nixos-hardware overrides.

raspberry-pi/common/firmware.nix
JamieMagee: `lib.stringAfter [ "specialfs" ] "..."` is shorthand for `{ deps = [ "specialfs" ]; text = "..."; }`. So this snippet runs after the built-in `specialfs` activation step, which mounts proc, sys, and dev. I ordered after it because `mountpoint -q` reads mountinfo, so proc needs to be up first.

raspberry-pi/common/firmware.nix
JamieMagee: Done

raspberry-pi/common/firmware.nix
JamieMagee: Done.

JamieMagee:
@doronbehar I think I've handled your last round of review. Mind taking another look when you get a chance? Thanks!

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,201 @@
...
+          ''
+            # Prefer the booted generation's device trees over the vendor ones.
+            dtbSrc=/run/current-system/dtbs
+            [ -d "$dtbSrc" ] || dtbSrc=$firmwareBoot
+          ''
+        else
+          ''
+            dtbSrc=$firmwareBoot
+          ''
+      }
```
doronbehar, line 46: You didn't like the string addition suggestion I gave earlier?

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,201 @@
...
+        Copy device trees from the booted NixOS generation
+        (`/run/current-system/dtbs`) instead of the vendor firmware package.
+        Irrelevant when generating an SD image.
+      '';
+    };
+
+    installScript = lib.mkOption {
+      type = lib.types.package;
+      readOnly = true;
+      internal = true;
```
doronbehar: I was wondering, what is the point of having an option like this? You don't even use `cfg.installScript` - now I noticed.

doronbehar (commented):
Looks pretty good, have you tested this?

raspberry-pi/common/firmware.nix
JamieMagee: The `+` version would mean breaking the `text` string in two and concatenating, since the rest of the install script follows right after this block. The `${if ... then ... else ...}` interpolation keeps it all in one string and still gets the single if/else, rather than two `optionalString`s.

raspberry-pi/common/firmware.nix
JamieMagee: I think this got lost in the refactoring? It was leftover from the `populateCmd` option. I'll drop it.

JamieMagee:
Yeah, tested, though build-time only. I built the sd image and inspected it: the DTBs, overlays, GPU boot code, and rendered `config.txt` all land, and with `uboot.enable` it stages `u-boot.bin` and sets `kernel=u-boot.bin`.

I haven't booted it on real hardware, though. My only Pi is a Pi 5 running NVME boot, and U-Boot still can't boot a Pi 5 off NVME, so I can't exercise the extlinux path end-to-end here.

Also: the `uboot.package` default `pkgs.ubootRaspberryPiAarch64` exists in current nixpkgs, just not our pinned lock, so I'd rather not bump it here. The dependabot PRs #1909 and #1892 already carry the nixpkgs bumps.

raspberry-pi/common/firmware.nix
doronbehar: I didn't mean to go back to `optionalString`s, but it is a nitpick so I won't block it.

raspberry-pi/common/firmware.nix
```diff
@@ -0,0 +1,192 @@
...
+    '';
+  };
+in
+{
+  options.hardware.raspberry-pi.firmware = {
+    enable = lib.mkEnableOption ''
+      installation of the Raspberry Pi firmware partition on a running system.
+
+      An activation script repopulates {option}`hardware.raspberry-pi.firmware.path`
+      on every system switch'';
```
doronbehar: Again the formatting... :).

doronbehar (approved):
Almost perfect! Will test it in 25 hours or more, and will merge afterwards.

raspberry-pi/common/firmware.nix
JamieMagee: Okay, marking this as resolved.

raspberry-pi/common/firmware.nix
JamieMagee: I have been running nixfmt, but it looks like it doesn't care about this style 🤷 Either way, fixed now.

doronbehar:
Sorry, but I'll be able to test this on real hardware only in 26 hours or 50 hours from now.

raspberry-pi/README.md
```diff
@@ -51,6 +72,7 @@ Top-level attrs are conditional sections (`all`, `pi4`, `pi5`, `cm4`, and so on)
...
-- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over. So a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- No bootloader module: There's no `boot.loader.raspberry-pi` here. Boards rely on `generic-extlinux-compatible` plus U-Boot. Raspberry Pi OS has the GPU firmware load the kernel directly; we go through U-Boot so it reads `extlinux.conf`, which is what gives you the NixOS boot-generation menu and rollbacks. The firmware install module just stages the boot code and (optionally) U-Boot; it doesn't add a firmware-level direct-boot path. Pi 5 boots from SD via U-Boot, but USB, PCIe, and the RP1 don't come up until Linux takes over, so a USB keyboard at the U-Boot prompt won't work on Pi 5 today.
+- No NVMe boot on Pi 5: it needs BCM2712 PCIe support in U-Boot, which is still [in review upstream](https://lists.denx.de/pipermail/u-boot/2026-June/620499.html) and not in a tagged release. `ubootRaspberryPiAarch64` picks it up once nixpkgs ships a U-Boot that has it.
```
JamieMagee: For future reference, this has landed in u-boot 2026.07-rc4. 2026.07 is due to release on 6th July. See https://github.com/NixOS/nixpkgs/pull/534348

doronbehar:
Hey sorry it takes me so long to reach my hardware... Here's my test `flake.nix`:

```nix
{
  description = "NixOS flake to test nixos-hardware PR #1894";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    nixos-hardware.url = "github:NixOS/nixos-hardware/pull/1894/head";
    nixos-hardware.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs =
    {
      self,
      nixpkgs,
      nixos-hardware,
      ...
    }:
    {
      nixosConfigurations.test-pi = nixpkgs.lib.nixosSystem {
        modules = [
          "${nixpkgs}/nixos/modules/installer/sd-card/sd-image-aarch64.nix"
          nixos-hardware.nixosModules.raspberry-pi-3
          ({ config, pkgs, ... }: {
            nixpkgs.hostPlatform = "aarch64-linux";
            services.openssh.enable = true;
            users.users.testuser = {
              isNormalUser = true;
              extraGroups = [
                # For sudo
                "wheel"
                # For testing hardware with alsa-utils
                "audio"
              ]; 
              hashedPassword = "$y$j9T$NPMgKje91gAZQ7ZOli5FU0$JC6PpwRVVMqfG3x6.QY/h8iAc1eig5oucY4BDd3/2b0";
            };
            environment.systemPackages = with pkgs; [
              # Testing utility
              alsa-utils
            ];
            hardware.raspberry-pi.configtxt.settings = {
              all = {
                dtoverlay = [
                  "allo-piano-dac-plus-pcm512x-audio"
                ];
              };
            };
            hardware.raspberry-pi.firmware = {
              enable = true;
              uboot.enable = true;
            };
          })
        ];
      };
    };
}
```

Going to test this with:

```sh
nix build -L .\#nixosConfigurations.test-pi.config.system.build.sdImage && zstd -d result/sd-image/nixos-image-rpi3-uboot.img.zst --stdout | sudo dd of=/dev/sda bs=4M status=progress
```

As soon as I get time with my hardware!

JamieMagee:
I managed to get my hands on a spare Raspberry Pi 4 and test this out. It works 🎉... with one minor modification 😅

The `kernel=` parameter in `config.txt` expects a 32-bit binary by default, but `uboot.package` is set to `pkgs.ubootRaspberryPiAarch64` by default. Setting `hardware.raspberry-pi.configtxt.settings.all.arm_64bit = lib.mkDefault pkgs.stdenv.hostPlatform.isAarch64;` works.

EDIT: updated `firmware.nix`

Here's my `flake.nix` as well:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    nixos-hardware.url = "git+file:///home/jamie/src/code/nixos-hardware";
    nixos-hardware.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs =
    {
      self,
      nixpkgs,
      nixos-hardware,
      ...
    }:
    {
      nixosConfigurations.rpi4 = nixpkgs.lib.nixosSystem {
        modules = [
          "${nixpkgs}/nixos/modules/installer/sd-card/sd-image-aarch64.nix"
          nixos-hardware.nixosModules.raspberry-pi-4
          (
            { config, lib, ... }:
            {
              nixpkgs.hostPlatform = "aarch64-linux";
              nixpkgs.buildPlatform = "x86_64-linux";

              boot.supportedFilesystems.zfs = lib.mkForce false;

              system.stateVersion = config.system.nixos.release;

              networking.hostName = "rpi4";

              hardware.raspberry-pi.firmware = {
                enable = true;
                uboot.enable = true;
              };

              services.openssh.enable = true;

              users.users.nixos = {
                isNormalUser = true;
                extraGroups = [ "wheel" ];
                initialPassword = "nixos";
              };
            }
          )
        ];
      };
    };
}
```

EDIT 2: It also boots fine with `boot.kernelPackages = pkgs.linuxPackages_latest;`

doronbehar:
I just find time with my Raspberry Pi 3 hardware, and wanted to test this finally (sorry it took me so long!), and encountered this error:

```
$  nix log /nix/store/8j1xbrcrmsfi32cfr4zpamgdd1y5kpcs-nixos-system-nixos-sd-card-26.11.20260629.b5aa0fb.drv
The bootloader cannot find the proper kernel image.
(Expecting /nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609/Image)
```

With the `flake.nix` above (that tests real hardware that needs dtb files).

Any idea why is that? It happened right before the end of the build...

JamieMagee:
@doronbehar Thanks for testing! I don't think this one's on the PR though. "The bootloader cannot find the proper kernel image" comes from `system.systemBuilderCommands` in nixpkgs' `kernel.nix`. It just checks that `${kernel}/Image` exists. This PR only adds the firmware module, it never touches `boot.kernelPackages` or the kernel image, and that check runs while building `system.build.toplevel`, before the SD-image firmware step even runs.

I reproduced your config and got the same kernel path, `l9fjh03z...-linux-rpi-6.18.34-stable_20260609`. Built it, and `$out/Image` is there, with the DTBs fixed up. So the recipe produces `Image` fine, which makes me suspect something local on your side.

Could you check a couple of things?

- `ls -la /nix/store/l9fjh03z...-linux-rpi-6.18.34-stable_20260609/` (is `Image` present? is the path actually a directory?)
- `nix log $(nix-store -q --deriver /nix/store/l9fjh03z...)`, in case the kernel build got interrupted or ran out of disk

If `Image` is missing, delete that path (`nix store delete`) and rebuild with `--rebuild`.

doronbehar:
> * `ls -la /nix/store/l9fjh03z...-linux-rpi-6.18.34-stable_20260609/`

Here's the output of `tree --filelimit=3 /nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609`:

```
/nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609
├── dtbs
│   ├── broadcom  [26 entries exceeds filelimit, not opening dir]
│   └── overlays  [370 entries exceeds filelimit, not opening dir]
├── System.map-6.18.34
└── vmlinux-6.18.34

4 directories, 2 files
```

As for:

> * `nix log $(nix-store -q --deriver /nix/store/l9fjh03z...)`

Gives [this log](https://gist.github.com/doronbehar/5f9333d436b45bce50e45350b71ec5b5).

Using:

```
nix store delete /nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609
```

Gave:

```
0 store paths deleted, 0.0 KiB freed
error: Cannot delete path '/nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609' since it is still alive. To find out why, use: nix-store --query --roots and nix-store --query --referrers
```

And:

```
nix-store --query --referrers /nix/store/l9fjh03zjm769i9ddh89mkz553pd3rwj-linux-rpi-6.18.34-stable_20260609
```

Gave:

```
/nix/store/f8inl3848bdx0hcqnx466zkaxv856ps5-linux-rpi-6.18.34-stable_20260609-dev
/nix/store/gb1mzi9jgnxb3qkyrwljlbfafsbsk073-boot.json
```

Which I don't know how to interpret...

JamieMagee:
@doronbehar I couldn't reproduce the bypass. I built your exact kernel derivation natively under qemu-aarch64, with and without modules, and instrumented `scripts/install.sh` to log what it does. Every time, `HOME` points at nixpkgs' `installkernel` shim, the shim wins, and `$out/Image` is produced. Cross does the same. So the same drv that gave you `vmlinux-6.18.34` gives me `Image`, which means the derivation itself isn't the problem.

The immediate cause still shows up in your log: during `make install`, `scripts/install.sh` skipped the `installkernel` shim and fell through to `arch/arm64/boot/install.sh`, which names a normal image `vmlinux-<ver>`, so NixOS can't find `$out/Image`. That script tries `$HOME/bin/installkernel`, then `/sbin/installkernel`, then the arch scripts, so it's sensitive to the build environment. Something on your build host is tipping it into the fallback, which is why it doesn't show up here. Same class as NixOS/nixpkgs#521113 and NixOS/nixpkgs#535850.

To narrow it down, could you share:
- `nix config show | grep -i sandbox`
- the OS of the machine that built the kernel (the Pi itself, or a bigger aarch64 box? your log showed `-j20`)
- `ls -l /sbin/installkernel /usr/sbin/installkernel` there
- whether deleting `/nix/store/l9fjh03z...` and rebuilding reproduces it

doronbehar:
> Easiest workaround: cross compile with `nixpkgs.buildPlatform = "x86_64-linux";`. I'll see if we can make the kernel derivation guarantee `$out/Image` so native builds don't need it.

I'm sorry but I don't yet have the time and CPU power to do that, at least not right now.

> @doronbehar thanks for testing again. I still don't think the issue you're seeing is because of this PR.

I understand :), but it still precludes me from fully approving this PR :/. I wasn't trying to complain that your PR is problematic, more asking for your help really :).

> Looks like a similar issue to [NixOS/nixpkgs#521113](https://github.com/NixOS/nixpkgs/issues/521113) and [NixOS/nixpkgs#535850](https://github.com/NixOS/nixpkgs/issues/535850).

I just built it again with the latest nixos-unstable from 2026-07-02, per the reports on that issue and it still won't work... I did manage to workaround the issue with:

```nix
            # Workaround for something similar to:
            # https://github.com/NixOS/nixpkgs/issues/521113
            system.boot.loader.kernelFile = "vmlinux-6.18.34";
```

Maybe we need to version bump the Linux kernel nixos-hardware uses?

Anyway, as for the test results with the above workaround, I flashed that image to an SD card and booted my hardware with it, and nothing booted. Not surprising TBH, as a `vmlinux` kernel file probably doesn't fit as an SD card.

I'm terribly sorry but I won't be able to merge this until I will verify that a simple flake.nix like compiles and boots successfully - I need to verify it with real hardware that actually needs one of these dtbs to function.

Do I understand @JamieMagee that you always cross compiles these images? I use the [Nix community aarch64-linux builder](https://nix-community.org/community-builders/) to build these images...

Eason20000:
### Missing `enable_uart=1` in `config-txt-defaults.nix` causes U-Boot hang on Pi 4

When deploying via `nixos-rebuild switch` (not SD image builder), the generated `config.txt` only contains settings from `hardware.raspberry-pi.configtxt.settings`. The `enable_uart=1` line — present in nixpkgs' `sd-image-aarch64.nix` (see [line 42](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/installer/sd-card/sd-image-aarch64.nix#L42)) but absent from [`config-txt-defaults.nix`](https://github.com/NixOS/nixos-hardware/blob/a79189f27a0ea8afe4a11f12009ce91cba18c5e4/raspberry-pi/common/config-txt-defaults.nix) — is missing.

On a Pi 4B, U-Boot hangs at early boot, and HDMI shows nothing (or rainbow screen if without `disable_fw_kms_setup`) and it stops there. Adding `hardware.raspberry-pi.configtxt.settings.all.enable_uart = true` resolves it.

**Reproduction context (nixos-anywhere + disko, not SD card):**

My setup uses [disko](https://github.com/nix-community/disko) for partitioning with a custom layout (FAT at `/boot`, LUKS+LVM root) deployed via `nixos-anywhere`, not the SD image builder.

Key config in `hardware.nix`:

```nix
{ config, lib, pkgs, inputs, ... }: {
  imports = [ inputs.nixos-hardware-pi.nixosModules.raspberry-pi-4 ];

  hardware.raspberry-pi.firmware = {
    enable = true;
    uboot.enable = true;
    path = "/boot";
  };

  # Required — without this U-Boot hangs on Pi 4
  hardware.raspberry-pi.configtxt.settings = {
    all.enable_uart = true;
  };
}
```

Kernel is the nixos-hardware Raspberry Pi kernel, cross-compiled from x86_64:

```nix
boot.kernelPackages =
  let
    crossPkgs = import pkgs.path {
      localSystem = "x86_64-linux";
      crossSystem = pkgs.stdenv.hostPlatform.system;
    };
  in
  crossPkgs.linuxPackagesFor (
    crossPkgs.callPackage (
      inputs.nixos-hardware-pi.outPath + "/raspberry-pi/common/kernel.nix"
    ) { rpiVersion = 4; }
  );
```

And the disko layout for the FAT partition (`disko.nix`):

```nix
partitions.ESN20-RPI4B_firmware = {
  size = "1024M";
  type = "EF00";
  content = {
    type = "filesystem";
    format = "vfat";
    mountpoint = "/boot";
    mountOptions = [ "umask=0077" ];
  };
};
```

I guess the SD image builder avoids this because its standalone `config.txt` explicitly includes `enable_uart=1`. But when using `hardware.raspberry-pi.configtxt.settings` (the only path available to non-SD-image deployments), it's missing.

Since `sd-image-aarch64.nix` documents this as a U-Boot requirement (see its comment referencing `arch/arm/mach-bcm283x/Kconfig`), it should be in `config-txt-defaults.nix` as `lib.mkDefault`.

Flake input pinned to this PR's head:

```nix
nixos-hardware-pi = {
  url = "github:NixOS/nixos-hardware/pull/1894/head";
  inputs.nixpkgs.follows = "nixpkgs";
};
```

`flake.lock` locked at `a79189f` (2026-06-29), nixpkgs at `567a49d` (channel `nixos-unstable`).

JamieMagee:
@Eason20000 Good catch! It looks like that comment is still there in u-boot[^1]:

```
This option assumes the VideoCore firmware is configured to use the
mini UART (rather than PL011) for the serial console. This is the
default on the RPi 4. To enable the UART console, the following non-
default option must be present in config.txt: enable_uart=1. This is
required for U-Boot to operate correctly, even if you only care
about the HDMI/usbkbd console.
```

[^1]: https://github.com/u-boot/u-boot/blob/18761722e67d6e3770ab296a04fe3d40022e76d2/arch/arm/mach-bcm283x/Kconfig#L165-L170

doronbehar:
BTW regarding the failure to produce an aarch64-linux SD image with/out this PR on a aarch64-linux: I just updated my [`nixos-raspberrpi`](https://github.com/nvmd/nixos-raspberrypi) based system with the latest Nixpkgs etc and used the SD image builder without any trouble. Therefor I conclude this is not an issue with Nixpkgs, but rather something specific to nixos-hardware.

doronbehar:
> @Eason20000 Good catch!

Maybe we should use `hardware.raspberry-pi.configtxt.settings.cm4.enable_uart = lib.mkDefault true;` in [`raspberry-pi/common/config-txt-defaults.nix`](https://github.com/NixOS/nixos-hardware/blob/raspberry-pi-firmware-install/raspberry-pi/common/config-txt-defaults.nix)?

JamieMagee:
@doronbehar I added `enable_uart` to match [`sd-image-aarch64.nix`][1]. And I'll request access to the nix community aarch64 builders to try and repro the issue you are seeing there.

@Eason20000 would you be able to test the updated branch please?

[1]: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/installer/sd-card/sd-image-aarch64.nix

raspberry-pi/common/config-txt-defaults.nix
```diff
@@ -18,6 +18,14 @@
...
       dtparam = lib.mkDefault [ "audio=on" ];
       dtoverlay = lib.mkDefault [ "vc4-kms-v3d" ];
+      # U-Boot needs the UART enabled to boot (see arch/arm/mach-bcm283x/Kconfig
+      # in the U-Boot tree). Without it U-Boot hangs on the Pi 4.
+      enable_uart = lib.mkDefault true;
+    };
+    # The Pi 5 has a dedicated debug UART. Leaving the mini UART on feeds ghost
+    # input into boot, so turn it back off.
+    pi5 = {
+      enable_uart = lib.mkDefault false;
```
doronbehar, line 28: I'd make these changes part of a separate commit or even a separate PR.

doronbehar:
> I'll request access to the nix community aarch64 builders to try and repro the issue you are seeing there.

Thanks. In general you will gain a lot of benefit from these builders, as your contributions are in general highly relevant to NixOS' aarch64-linux support.

raspberry-pi/common/config-txt-defaults.nix
JamieMagee: Done: https://github.com/NixOS/nixos-hardware/pull/1938

JamieMagee:
I got on the community builder 😄 

I built this PR's Pi kernel there natively and it produced `$out/Image` with nixpkgs' `installkernel` shim running normally. Same result cross-compiling and under qemu-native. Unfortunately, I can't reproduce your original failure in any of those environments ☹️ 

My best guess is the broken kernel path is stale in your local store and nix is reusing it rather than rebuilding. Could you try `nix build --rebuild`  on the kernel, or delete that path and rebuild, and check whether a fresh output still lacks `Image` ?

doronbehar:
> I got on the community builder 😄

Welcome aboard :).

> Unfortunately, I can't reproduce your original failure in any of those environments ☹️

I too can't reproduce the issue anymore! I Update my `flake.nix` to use latest Nixpkgs `master`, and your PR to include both this PR and https://github.com/NixOS/nixos-hardware/pull/1938, and flashed the SD image. Will test it on my hardware as soon I get the chance :).

Eason20000:
> @doronbehar I added `enable_uart` to match [`sd-image-aarch64.nix`](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/installer/sd-card/sd-image-aarch64.nix). And I'll request access to the nix community aarch64 builders to try and repro the issue you are seeing there.
> 
> @Eason20000 would you be able to test the updated branch please?

Tested. It works pretty well. Thanks!

doronbehar:
OK So as I said earlier, my approval test for this PR is this `flake.nix`:

```nix
{
  description = "NixOS flake to test nixos-hardware PR #1894";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs";
    # With https://github.com/NixOS/nixos-hardware/pull/1938 Rebased upon latest master
    nixos-hardware.url = "github:doronbehar/nixos-hardware/raspberry-pi-firmware-install";
    nixos-hardware.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs =
    {
      self,
      nixpkgs,
      nixos-hardware,
      ...
    }:
    {
      nixosConfigurations.test-pi = nixpkgs.lib.nixosSystem {
        modules = [
          "${nixpkgs}/nixos/modules/installer/sd-card/sd-image-aarch64.nix"
          nixos-hardware.nixosModules.raspberry-pi-3
          ({ config, pkgs, ... }: {
            nixpkgs.hostPlatform = "aarch64-linux";
            services.openssh.enable = true;
            users.users.testuser = {
              isNormalUser = true;
              extraGroups = [
                # For sudo
                "wheel"
                # For testing hardware with alsa-utils
                "audio"
              ]; 
              hashedPassword = "$y$j9T$NPMgKje91gAZQ7ZOli5FU0$JC6PpwRVVMqfG3x6.QY/h8iAc1eig5oucY4BDd3/2b0";
            };
            environment.systemPackages = with pkgs; [
              # Testing utility
              alsa-utils
            ];
            hardware.raspberry-pi.configtxt.settings = {
              all = {
                dtoverlay = [
                  "allo-piano-dac-pcm512x-audio"
                ];
              };
            };
            hardware.raspberry-pi.firmware = {
              enable = true;
              uboot.enable = true;
            };
          })
        ];
      };
    };
}
```

And it doesn't work for me - `alsamixer` didn't show an piano DAC sound card.

After a non-trivial Claude-Code session in which it read both Nixpkgs and `nixos-hardware` files (with this PR of course), it realized that I need to set:

```nix
boot.loader.generic-extlinux-compatible.useGenerationDeviceTree = false;
```

And it worked! Here's Claude's explanation (I sort-of/partially understand it): 

> With `uboot.enable = true`, the RPi GPU firmware (`start.elf`) reads `config.txt`, applies every `dtoverlay=` onto the base board DTB, and hands that merged DTB to `u-boot.bin` in RAM. But `boot.loader.generic-extlinux-compatible`'s default `useGenerationDeviceTree = true` makes the generated `/boot/extlinux/extlinux.conf` include an `FDTDIR ../nixos/<hash>-dtbs` line pointing at the kernel package's own *bare*, un-overlaid dtbs (with no `config.txt`). U-Boot's distro-boot script sees that line and loads a fresh DTB from there, discarding the overlay-applied one it was just handed — so every `dtoverlay` set via `hardware.raspberry-pi.configtxt.settings` is silently dropped before Linux ever boots.
>
> Since `config.txt`-driven overlays are this module's whole interface, and `uboot.enable` is the documented way to boot, we probably should set `boot.loader.generic-extlinux-compatible.useGenerationDeviceTree = lib.mkDefault false;` by the `hardware.raspberry-pi.configtxt.firmware` module — inside the existing `lib.mkIf cfg.uboot.enable { ... }` block in `raspberry-pi/common/firmware.nix`.
>
> `nixos-raspberrypi` independently reached the same conclusion: its [`boot.loader.raspberry-pi.useGenerationDeviceTree` option](https://github.com/nvmd/nixos-raspberrypi/blob/9977789f7d9b98c2825b423bd4e811078baec6a0/modules/system/boot/loader/raspberrypi/default.nix#L286-L300) defaults to `false` whenever its `bootloader` isn't `"kernel"` (i.e. whenever U-Boot is chainloaded), and that value is [forwarded straight into `boot.loader.generic-extlinux-compatible.useGenerationDeviceTree`](https://github.com/nvmd/nixos-raspberrypi/blob/9977789f7d9b98c2825b423bd4e811078baec6a0/modules/system/boot/loader/raspberrypi/default.nix#L556-L562), with the comment:
> > if false, don't add FDTDIR to the extlinux conf file to use the device tree provided by firmware.
>
> Only its (legacy) `"kernel"` bootloader mode (RPi firmware boots Linux directly, no U-Boot/extlinux involved) keeps the default `true`, which lines up with this only mattering when U-Boot actually reads `extlinux.conf`.

Lastly: I'm not sure @Eason20000 what exactly worked for you or how you tested this PR, but I'm fairly sure of my tests.

Eason20000:
> Lastly: I'm not sure @Eason20000 what exactly worked for you or how you tested this PR, but I'm fairly sure of my tests.

I'm using my Pi 4B mainly as a headless server, so after it boots (instead of hanging up in uboot before adding enable_uart=1 in config.txt), I stopped testing and assumed it worked, without testing the device tree. Sorry for the lack of specificity before. Also, I'm a bit doubtful whether the config.txt in the generated SD image is actually provided by sd-image-aarch64.nix in nixpkgs or by this PR.

JamieMagee:
@doronbehar I was able to confirm your fix. I traced it through the extlinux builder: `useGenerationDeviceTree = true` skips the `-r` flag[^1], so it writes `FDTDIR ../nixos/<hash>-dtbs`[^2] and U-Boot loads a bare DTB from there instead of the overlay-merged one from the GPU firmware. Setting it false drops that line, matching what you saw.

I added `boot.loader.generic-extlinux-compatible.useGenerationDeviceTree = lib.mkDefault false;` in the  uboot.enable  block, same as nixos-raspberrypi it turns out[^3]

Are you happy to merge?

---

@Eason20000 with this PR imported, `config.txt` comes from the PR, not `sd-image-aarch64.nix`. The firmware module overrides `sd-image-aarch64.nix`'s `populateFirmwareCommands` and writes `config.txt` from `hardware.raspberry-pi.configtxt.settings`.

Quick way to tell which one is on the firmware partition: the PR's `config.txt` has `kernel=u-boot.bin` plus any `dtoverlay` lines you set. The nixpkgs one has `kernel=u-boot-rpi4.bin` and no overlays.

The device-tree path you didn't test is the one doronbehar hit: with U-Boot, overlays make it into `config.txt` but U-Boot then loads a bare device tree and ignores them. I pushed a fix, so overlays set via `configtxt.settings` now apply.

[^1]: https://github.com/NixOS/nixpkgs/blob/8b8484d2fceb1a9acf02fbd13da9e99d15b7e410/nixos/modules/system/boot/loader/generic-extlinux-compatible/default.nix#L110
[^2]: https://github.com/NixOS/nixpkgs/blob/8b8484d2fceb1a9acf02fbd13da9e99d15b7e410/nixos/modules/system/boot/loader/generic-extlinux-compatible/extlinux-conf-builder.sh#L111-L123
[^3]: https://github.com/nvmd/nixos-raspberrypi/blob/9977789f7d9b98c2825b423bd4e811078baec6a0/modules/system/boot/loader/raspberrypi/default.nix#L557-L562

JamieMagee:
I also removed the note in the README about NVME boot on Raspberry Pi5. It's working since https://github.com/NixOS/nixpkgs/pull/534348 was merged. Thanks again!

doronbehar:
@JamieMagee I deeply thank you for working on this issue and for being so patient - both patient for me and for delving into the technical details of this. I'm finally merging :tada: .

PR #1891, opened 2026-05-27
feat: Add ThinkPad X13 Gen 7 support

takumiando:
###### Description of changes

Add support for the ThinkPad X13 Gen 7 (Intel).

Once support for Intel Core Ultra Series 3 (Panther Lake) becomes available, this configuration should import it.

###### Things done

- [x] Tested the changes in your own NixOS Configuration
- [x] Tested the changes end-to-end by using your fork of `nixos-hardware` and
      importing it via `<nixos-hardware>` or Flake input

PR #1912, opened 2026-06-13
feat: Dell XPS 14 DA14260

cooparo:
## What this is

A NixOS hardware profile for the **Dell XPS 14 DA14260**. It makes the machine's hardware work that otherwise needs manual setup.

*Depends on [#1891](https://github.com/NixOS/nixos-hardware/pull/1891) (Panther Lake CPU profile).*

## What works with it

| Hardware | Result |
|---|---|
| **Webcam** | 4K (3840×2160) at ~28.6 fps, correct colours, via the Intel IPU7 hardware ISP |
| **Speakers** | Working; vendor-style tuning available, opt-in (see below) |
| **Microphone** | Working on Linux 7.0+ |
| **Infrared camera + face unlock** | Working, opt-in (see below) |
| **Firmware updates** | `fwupd` enabled |

Speakers, mic and both cameras do **not** work on a stock install without this profile.

## Webcam

The front camera runs through Intel's hardware image processor (camera HAL + per-sensor tuning), which fixed the low resolution, wrong colours and instability of the plain libcamera path. It vendors four packages from the not-yet-merged nixpkgs [#542085](https://github.com/NixOS/nixpkgs/pull/542085); once that lands they get replaced by `hardware.ipu7`. Thanks @aoli-al for that packaging and for testing here.

The camera needs Intel's proprietary firmware, so you have to allow it:

```nix
nixpkgs.config.allowUnfreePredicate =
  pkg: builtins.elem (lib.getName pkg) [ "ipu7-camera-bins" "ivsc-firmware" ];
```

The finer points (kernel-version handling for the CVS bridge, the `v4l2-relayd` relay service, resolution benchmarks) are in the profile README.

## Infrared camera & face unlock (opt-in)

The infrared camera — the one Windows uses for Hello face unlock — works. The earlier "blocked at the firmware level" claim was wrong; it was a software bug (the sensor's clock rate was reported in the wrong unit, so the camera ran at the wrong speed and never produced a frame).

It is **off by default** because turning it on has to compile the kernel locally — the one-line patch for this sensor is only in mainline Linux after 7.2, and the profile supports 7.1 — so only people who want face unlock pay that build cost:

```nix
hardware.dell-xps-14-da14260.irCamera.enable = true;        # the camera
hardware.dell-xps-14-da14260.irCamera.howdy.enable = true;  # + face unlock (Howdy)
```

Stock Howdy can't use this sensor (its recorders can't read the raw image format and never switch the infrared light on). The second option installs Howdy with a small purpose-built reader that does, and sets face recognition as a *convenience* — a failed scan always falls back to your password, so you can't get locked out. With both on you get face unlock at the login screen, on `sudo`, and on the lock screen.

**Caveats:** kernel 7.1.x only (enforced by an assertion); verified on two machines (this DA14260 and an XPS 16 DA16260); the Howdy part only activates on a nixpkgs new enough to ship `services.howdy` — older channels are unaffected.

## Speaker tuning (opt-in)

Laptop speakers ship voiced by the vendor's Windows sound layer, which Linux doesn't get. This adds it back as a PipeWire filter — 13 EQ bands into a limiter — in front of the internal speakers:

```nix
hardware.dell-xps-14-da14260.speakerTuning.enable = true;
```

Ported from [basecamp/omarchy@aa9f0c5](https://github.com/basecamp/omarchy/commit/aa9f0c54c5c3bd8141f1d30ecf52e6b377a45fe5), suggested by @HritwikSinghal. It's just numbers in a config file — no binary blob, and nothing from the original is redistributed.

**Off by default**, since the speakers work fine untuned and this is a matter of taste.

Two things to know if you turn it on: a second output device appears ("Laptop Speakers"), and it's about **9 dB quieter** than untuned, so turn the volume up to compare fairly. The quietness is by design — it makes room for the bass boost without clipping.

**I can't tell you whether it sounds better** — the plumbing is tested, the voicing isn't. Feedback welcome.

## Things done

- [x] Tested in my own NixOS configuration
- [x] Tested end-to-end via flake input

## Related

* nixpkgs [#542085](https://github.com/NixOS/nixpkgs/pull/542085), [#479283](https://github.com/NixOS/nixpkgs/pull/479283)
* [intel/ipu7-camera-hal#59](https://github.com/intel/ipu7-camera-hal/pull/59) · [intel/ipu7-drivers#26](https://github.com/intel/ipu7-drivers/issues/26) · [intel/vision-drivers#37](https://github.com/intel/vision-drivers/issues/37)
* [HritwikSinghal/svp7500-camera-fix-pack](https://github.com/HritwikSinghal/svp7500-camera-fix-pack)


noisesfromspace:
I've been on the 14250 since last year and appalled by the Linux support. Will test run this fork because it should have similar enough quirks. 

edit: the 50 is [Arrow Lake](https://www.intel.com/content/www/us/en/ark/products/codename/225837/arrow-lake.html) though not Panther Lake. But looking at the https://github.com/takumiando/nixos-hardware/blob/3887211270897deef92d36c4660a3fe7f63ba29a/common/gpu/intel/panther-lake/default.nix it shouldn't be too much of an issue. Just FYI for someone reading this with the same laptop

aoli-al:
Tested on DA14260. Both camera and mic are working.

aoli-al (approved):

cooparo:
How is the camera for you? There are some issues: the camera sometimes is laggy, sometimes is not, and the colors are not correct. Have you experienced these?

aoli-al:
Camera was laggy while using zoom. I don't use camera much so it was okay for me. There is another patch that does not require v4l2loopback: https://github.com/jibsta210/ipu7-camera-linux. Will this provide better performance?

cooparo:
And what about colors?

aoli-al:
color was fine for me

aoli-al:
I got hardware ISP working with this PR: https://github.com/NixOS/nixpkgs/pull/542085

To use hardware ISP you may apply the following patch:

```
diff --git a/dell/xps/14-da14260/default.nix b/dell/xps/14-da14260/default.nix
index 2fdb42d7f1..76d06a9655 100644
--- a/dell/xps/14-da14260/default.nix
+++ b/dell/xps/14-da14260/default.nix
@@ -16,6 +16,13 @@
     lib.mkDefault pkgs.linuxPackages_latest
   );
 
+  hardware.ipu7 = {
+    enable = true;
+    platform = "ipu75xa";
+  };
+
+  services.v4l2-relayd.instances.ipu7.enable = false;
+
   # Intel CVS driver for Synaptics SVP7500 camera bridge (06CB:0701).
   # Without this the IPU7 camera stack does not enumerate even though the
   # kernel-side intel_ipu7 driver detects the sensor (OVTI08F4 / OV08F4).
@@ -60,8 +67,8 @@
           gst-plugins-base
           gst-plugins-good
           gst-plugins-bad
+          icamerasrc-ipu75xa
         ])
-        ++ [ pkgs.libcamera ]
       );
       v4l2loopback-ctl = "${config.boot.kernelPackages.v4l2loopback.bin}/bin/v4l2loopback-ctl";
       deviceFile = "/run/ipu7-camera-relay/device";
@@ -80,7 +87,7 @@
       # so the input pipeline fails to build and only the black splash is shown.
       # v4l2-relayd applies the caps (copied from the output appsrc below) to its
       # internal appsink instead.
-      input = "libcamerasrc ! videoconvert ! videoscale ! videobalance saturation=1.8 ! videoflip method=vertical-flip";
+      input = "icamerasrc ! videoconvert ! videoscale ! videoflip method=vertical-flip";
 
       # A leaky queue (drops old frames) + sync=false keep latency low so the
       # viewer sees the latest frame instead of a backlog; the -b 4 buffers in
@@ -90,7 +97,7 @@
         + " ! queue leaky=downstream max-size-buffers=3 ! videoconvert ! v4l2sink name=v4l2sink device=$(cat ${deviceFile}) sync=false";
     in
     {
-      description = "Intel IPU7 camera to v4l2loopback relay (libcamera bridge)";
+      description = "Intel IPU7 camera to v4l2loopback relay (hardware ISP via camera HAL)";
       after = [ "modprobe@v4l2loopback.service" ];
       wantedBy = [ "multi-user.target" ];
       environment = {
diff --git a/xnwqqxqr.patch b/xnwqqxqr.patch
new file mode 100644
index 0000000000..e69de29bb2
```

phucisstupid:
any update?

cooparo:
Update: the camera stack is rewritten and the earlier known issues (instability, weird colors, low resolution) are fixed. The branch now uses the **IPU7 hardware ISP** — @aoli-al's stack from nixpkgs [#542085](https://github.com/NixOS/nixpkgs/pull/542085), vendored into the profile until that PR merges (thanks for the packaging and the patch!). The webcam now does **3840×2160 NV12 at ~28.6 fps with correct colors** (the sensor's single native mode; the ISP scales in hardware, so 4K costs the same as 720p).

Two gotchas discovered on **kernel 7.2** are handled by the profile, for anyone testing on `linuxPackages_latest`:
1. 7.2 gained an in-tree `intel_cvs` (`drivers/media/i2c/cvs`) and `ipu-bridge` now routes the sensor graph through its V4L2 subdev — the out-of-tree `intel_cvs` from `intel/vision-drivers` shadows it and breaks sensor binding, so the profile only vendors it for kernels < 7.2.
2. The `20260629_1` camera-HAL release can't parse the CVS-routed media graph (`STREAMON` fails with EPIPE), so the HAL is built from master (intel/ipu7-camera-hal#59) plus a config patch adding the CVS links for the ipu75xa `ov08x40` sensor — that patch should go upstream to Intel.

Note for testers: the camera blobs are unfree — you need `allowUnfreePredicate` for `ipu7-camera-bins` and `ivsc-firmware` (see the updated PR description). All commits squashed; details and benchmarks are in the profile README.

HritwikSinghal:
Hi @cooparo , the IR camera is also working now on my omarchy (XPS 14 2026, 7.1 kernel) thanks to https://github.com/jibsta210/svp7500-camera-fix-pack

see https://github.com/intel/vision-drivers/issues/37#issuecomment-5079510486 , we can look into importing these fixes to nixos also.

cooparo:
@HritwikSinghal that's really nice. I'm gonna take a look at what you did. How do you test the IR camera?

HritwikSinghal:
hi @cooparo , i am using IR for face unlocks over lockscreen and for sudo using howdy.

i did not test with `howdy test` -- that refuses with "recorder which doesn't support the test command". It's a V4L2-layer test, with one trap first. Everything below leans on the fix pack, [`jibsta210/svp7500-camera-fix-pack`](https://github.com/jibsta210/svp7500-camera-fix-pack) (and my fork with some fixes https://github.com/HritwikSinghal/svp7500-camera-fix-pack/commit/56a2687b6043224a4643f4c0cce4280888bdaf4d).

## The trap

A bare `v4l2-ctl -d /dev/videoN --stream-mmap` on the IR node gives `VIDIOC_STREAMON: Broken pipe` (EPIPE) on a fresh boot. That's not the sensor: the graph comes up with the CSI2 -> ISYS Capture link **disabled** and the CSI-2 pads at their 4096x3072 default, so link validation rejects the stream. libcamera normally does this setup, but the IR path bypasses it deliberately (libcamera debayers this sensor -- tagged `SGRBG10`, physically mono -- and renegotiates away from its native 648x368). I briefly wrote this up as evidence IR was broken; it wasn't.

Configure the graph first. Entity names from `media-ctl -p`, don't hardcode -- CSI-2 port is 2 on this DA14260 and the author's XPS 16 DA16260, but 1 on a Dell Pro 14:

```
media-ctl -d /dev/media0 --set-v4l2 '"hm1092 16-0024":0 [fmt:SGRBG10_1X10/648x368]'
media-ctl -d /dev/media0 --set-v4l2 '"Intel IPU7 CSI2 2":0 [fmt:SGRBG10_1X10/648x368]'
media-ctl -d /dev/media0 --set-v4l2 '"Intel IPU7 CSI2 2":1 [fmt:SGRBG10_1X10/648x368]'
media-ctl -d /dev/media0 -l '"Intel IPU7 CSI2 2":1 -> "Intel IPU7 ISYS Capture 16":0 [1]'

v4l2-ctl -d /dev/video16 --set-fmt-video=width=648,height=368,pixelformat=BA10 \
  --stream-mmap --stream-count=5 --stream-to=/tmp/ir.raw
```

## The tests

**`sudo ./tools/verify.sh`** is the canonical one: it resolves media device / CSI-2 port / sensor / capture node from the graph, does the setup above, streams, and counts start-of-frame interrupts from dmesg -- so it needs no illuminator and can't be fooled by a dark room. Healthy output is `hm1092 loaded (dkms)`, `link_frequency 180480000 (correct)`, `IR flood LED present (brightness=0)` (0 is correct while idle), the CSI-2 port followed from the sensor's own link, then `SOF on csi2-2  9 <-- STREAMING` / `RESULT: OK`. Exit 0 = OK, 1 = BROKEN, 2 = UNDETERMINED -- usually missing root or `v4l-utils`, and **not** a pass.

If you hand-roll the SOF check instead: `sof_event` is a `pr_debug` and emits **nothing** until you enable it, so you'll see zero SOF and conclude the sensor is dead. Enable `format sof_event +p` via `/sys/kernel/debug/dynamic_debug/control` first, then grep dmesg for `sof_event::csi2-2`. (`verify.sh` toggles it on and restores it afterwards.)

**Illuminator A/B** proves sensor + illuminator in two captures. It's a LED class device because INT3472 owns the GPIO -- which is why the sensor driver reports `ir_led=none` and the consumer has to drive it:

```
LED=/sys/class/leds/HIMX1092_00::ir_flood_led/brightness
echo 0 > $LED   # capture as above -> ir_off.raw
echo 1 > $LED   # capture as above -> ir_on.raw
echo 0 > $LED   # douse it -- see the warning below
```

Measured here just now on `7.1.8-arch1-2-ptl`, 5 frames each. Decode is not debayer: 16-bit LE containers, `bytesperline` 1344 (672 px of stride for 648 active), so crop to 648 then `>> 2` for 8-bit greyscale.

| illuminator | mean luma | max | near-black px |
|---|---|---|---|
| off | 7.8 | 22 | 100.0% |
| on | 54.7 | 255 | 0.4% |

Lit frames are a clean, correctly exposed, recognisable face -- a 7x mean-luma jump from one sysfs write.

**End to end:** `python3 /usr/lib/howdy/compare.py <user>` (exit 0 = match, ~1.2s unprivileged), then the real auth surfaces -- `sudo -k && sudo true`, `pkexec true`, lock screen.

Two Howdy-side settings: `recording_plugin = ir`, and `dark_threshold = 90` -- that's **not** brightness, it's the percentage of near-black pixels **above which a frame is rejected**. IR frames here are ~70-77% black, so the stock `60` rejects everything and working hardware looks broken.

**Warning worth passing on:** kill a scan with SIGTERM and the illuminator stays latched **on** -- screen off, emitter glowing, nothing in the journal. One watchdog kill here burned 5h05m of battery before I noticed. `cat $LED` (1 = stuck), `echo 0 >` to clear, no root needed. The fix pack now traps SIGTERM/SIGINT/SIGHUP to douse it, but a hand-rolled `media-ctl` + `v4l2-ctl` session like the one above has no such protection.

## Your README's IR section is stale

Flagging it because it currently tells readers the problem is unsolvable, which is the one thing that would stop someone porting this. The "zero activity on CSI-2 port 2" / "port-2 enable likely requires CSE/CSME authorization" / "not solvable from the Linux side" bullets cite intel/vision-drivers#37 at 2026-05-29. The author retracted that theory himself on [2026-07-25](https://github.com/intel/vision-drivers/issues/37) -- "SOLVED -- and the cause was on my side, not in the bridge":

> **Root cause: my sensor driver published `V4L2_CID_LINK_FREQ = 360,960,000`.** That number is correct, but it is the per-lane **MIPI bit rate**, whereas V4L2 defines `LINK_FREQ` as the **DDR clock** -- half of it, `180,480,000`. So `ipu7-isys` programmed the CSI-2 D-PHY at ~721 Mbps against a sensor transmitting at ~361. The clock lane came up, but no data packet ever framed: zero SOF on the port, indefinitely.

A receiver clocked 2x fast sees clock-lane transitions but never validates a packet -- from the host side, indistinguishable from a firmware block, which is why it read as one for months. The `0x0830` MIPI payload wasn't the fix either: per the fix pack's `IR-FINDINGS.md`, Windows never sends an IR `HOST_SET_MIPI_CONFIG` at all (all nine `0x0830` buffers across every capture carry RGB geometry).

One more from the fix-pack component table you reproduce, since it would mislead a Nix port: **don't port `int3472-patched`.** It's harmful on kernels >= ~7.1 -- mainline gained `skl_int3472_register_led` and registers `<sensor>::ir_flood_led` itself, while the patched module maps the GPIO to the sensor and registers no `led_classdev`, so installing it *removes* the node Howdy writes to. The fix pack's installer now skips it on such kernels.

## What I'd add to the profile

Happy to send these as a PR against your branch -- your call.

1. **`hm1092` vendored module**, same shape as your `./intel-cvs`, from the fix pack's `dkms/hm1092`. Stays vendored indefinitely: `drivers/media/i2c/hm1092.c` is 404 on mainline master.

2. **The fix pack's fork of `intel_cvs`, as a second, option-gated module package.** This is the piece that actually makes IR work, and it is *not* interchangeable with your existing `./intel-cvs`. `hm1092.c:39` declares `extern int __weak cvs_send_mipi_ir_config(void)` and calls it at stream start (`hm1092.c:1057`) to get CSI-2 port-2 forwarding configured. The rev your derivation pins, `intel/vision-drivers` `845d6f8`, has **zero** `EXPORT_SYMBOL` of any kind in `drivers/misc/icvs/`, and zero hits for `cvs_send_mipi_ir_config`, `HOST_SET_MIPI_CONFIG` or `mipi_cfg` -- that whole 0x0830 path is the fix pack's own code, exported at its `intel_cvs.c:159`. Because the reference is weak, `hm1092` links against either build: with upstream's the pointer is simply NULL, the call is skipped, port 2 is never configured, and every frame comes back dark. It fails silently at runtime, not at build.

   Both build an `intel_cvs.ko`, so exactly one may ever be in the module tree -- they need to be mutually exclusive by construction (drop the base profile's entry when the IR option is on), not merely both listed.

3. **`HIMX1092` in `ipu-bridge`, version-gated.** Mainline master line 53 has `IPU_SENSOR_CONFIG("HIMX1092", 2, 180000000, 180480000)` -- the `2` is `nr_link_freqs`, not a lane count (this sensor is 1-lane); both values are DDR clocks, for the 19.2 MHz and 24 MHz EXTCLK variants, and Dell uses 180480000. It landed in `4fdb0342f05e` on 2026-07-30, **after** the v7.2 tag: `grep -c HIMX1092 drivers/media/pci/intel/ipu-bridge.c` is 0 at v7.1 *and* 0 at v7.2. So the gate is `versionOlder ... "7.3"`, not `"7.2"` -- 7.2 still needs the backport, and without the entry ipu-bridge builds no software node, so the sensor never joins the media graph. Either `boot.kernelPatches` or an external `ipu-bridge` built from the target kernel's own source -- I lean external to avoid a full rebuild per profile user, but that's a detail for the diff.

   Don't vendor the fix pack's bridge wholesale. Its link frequency is fine (it carries 180480000, the same DDR-clock correction as the root cause above); the problem is that the fork isn't additive. Diffed against v7.1 it's +15/-9, and two of those hunks regress unrelated machines:
   - it **deletes** `IPU_SENSOR_CONFIG("OVTI5675", 1, 450000000)` (present at v7.1, v7.2 and master), so OV5675 laptops lose their bridge entry outright;
   - it **overwrites** the four entries of `upside_down_sensor_dmi_ids` instead of appending, so **XPS 13 9340** and **XPS 14 9440** lose their existing upside-down quirk to make room for the two Panther Lake boards.

   A one-entry backport of the mainline commit avoids both.

4. **IR-LED udev rule** into your existing `services.udev.extraRules` -- recognition runs unprivileged under a lock screen, so without it the illuminator silently never fires and every frame is too dark (the 100%-near-black row above). The fix pack's rule calls `/bin/chgrp` and `/bin/chmod`; NixOS activation creates exactly one entry in `/bin`, `sh`, so on NixOS those need store paths or both `RUN+=` lines are a silent no-op:
   ```
   SUBSYSTEM=="leds", KERNEL=="*ir_flood_led*", ACTION=="add|change", \
       RUN+="${pkgs.coreutils}/bin/chgrp video /sys%p/brightness", \
       RUN+="${pkgs.coreutils}/bin/chmod 0664 /sys%p/brightness"
   ```
   udev's `GROUP=`/`MODE=` can't substitute for those calls: the illuminator is a LED class device with no `/dev` node, so only its `brightness` sysfs attribute matters and those keys don't reach it.

5. **Load-order softdep** -- **helps RGB too**, so worth taking regardless of IR. `boot.extraModprobeConfig` goes multi-line with:
   ```
   softdep intel_ipu7 pre: usbio gpio_usbio i2c_usbio intel_cvs intel_skl_int3472_discrete
   ```
   Without it the bridge can come up after sensor discovery and loop on USB re-enumeration (`usbio-bridge 3-3:1.0: Bulk out failed: -110` / `usb 3-3: USB disconnect`), taking **both** sensors down. Your profile relies on boot timing here; so does mine -- I'm missing the file too and it works by luck.

6. Optional UX: the fix pack's `wireplumber/50-disable-v4l2-ipu7.conf` hides the ~30 dead raw-Bayer ipu7 V4L2 nodes from PipeWire so pickers show only the loopback. Applies to your RGB-only profile as-is.

## Caveats

- IR is verified on exactly **two** machines: the fix-pack author's XPS 16 DA16260 and this DA14260. Worth saying so in the README rather than implying universality.
- **The floor is 7.1**, for a reason independent of the bridge entry: `int3472/discrete.c` gained the `INT3472_GPIO_TYPE_STROBE -> "ir_flood"` mapping and the `skl_int3472_register_led()` call at v7.1 -- v7.0 has zero occurrences of all three. Below that the illuminator is never exposed as a LED device at all, so the udev rule matches nothing and every frame is dark no matter what else is right. With the bridge entry above, that puts the tested window at 7.1.x.
- All of it is 7.1.x with the fix pack's out-of-tree `intel_cvs`. IR on your `>= 7.2` in-tree-`cvs` path is **untested anywhere** -- but untested, not known-broken. The in-tree `drivers/media/i2c/cvs` exports no symbols either, and doesn't need to: it's a full bridge subdev that sends `ICVS_HOST_SET_MIPI_CONFIG` itself from `csi_set_link_cfg()` (`v4l2.c:56`), taking `link_freq` from the sensor's own control via `v4l2_get_link_freq()`. And `06CB:0701` is in its quirk table but is **not** given `ICVS_NO_MIPI_CONFIG` (only the Lattice NX33 is) -- so what `hm1092` asks for through the weak symbol may simply happen anyway. The one mechanism I can see that might genuinely block it: `ipu_bridge_instantiate_ivsc()` ends with `set_secondary_fwnode()` on the shared `csi_dev`, once per sensor and overwriting unconditionally, while in-tree cvs models exactly one remote -- and both sensors sit behind the same `INTC10E1:00`. Whether that touches the IR sensor depends on HIMX1092's ACPI `_DEP`, which I haven't read off a DSDT. First 7.3 kernel (in-tree HIMX1092) plus vendored `hm1092` is still the clean experiment.
- Howdy/PAM itself is user-side, out of scope for a nixos-hardware profile -- a README pointer to the fix pack's `howdy/` dir is probably the right depth.

One nit unrelated to IR, worth knowing if you debug this: `find-ir-node.sh` and `lib-detect.sh` resolve everything from the graph, but `ir_reader.py`'s four `media-ctl` calls swallow both return code and stderr -- if a `--set-v4l2` or the `-l` fails, it proceeds to `STREAMON` anyway and you get an opaque Howdy timeout with no hint that setup failed. Bit me once.

cooparo:
Thanks for the incredibly thorough writeup. We'd welcome a PR against this branch with those additions.

HritwikSinghal:
> Thanks for the incredibly thorough writeup. We'd welcome a PR against this branch with those additions.

Hi @cooparo , i have made some PRs for the fix (with help of claude). 

**PRs**, split so the RGB half can land independently -- it helps your profile with or without IR:

- cooparo/nixos-hardware#1 -- items 5 and 6 above (load-order softdep, wireplumber node hiding). One file, no new options.
- cooparo/nixos-hardware#2 -- items 1-4, opt-in behind `hardware.dell-xps-14-da14260.irCamera.enable`. Stacked on #1; I'll rebase once that lands.

Both template checkboxes are deliberately unticked: this machine runs Arch, so the profile builds and evaluates on both channels but has never been booted on NixOS. Everything measured above is Arch 7.1.8 against the same module sources.

HritwikSinghal:
We can also import speaker tunings from omarchy. see https://github.com/basecamp/omarchy/commit/aa9f0c54c5c3bd8141f1d30ecf52e6b377a45fe5 . This has made my speakers much better than stock and quite close to macbook levels.

cooparo:
History is written by the victors (jk)

cooparo:
## Update — infrared camera and face unlock now work (both opt-in)

Two additions since the last review, built on @HritwikSinghal's fix-pack work.

**The infrared camera works.** The description used to say it was blocked by Intel firmware and would not be fixed — that was wrong. The camera was being told to run its clock twice as fast as it actually does, so it never produced a picture. Corrected (plus a one-line kernel patch for this specific sensor), it streams normally.

It is **off by default**, because enabling it has to compile the kernel locally — the sensor patch only reached mainline Linux after 7.2, and the supported kernel here is 7.1. So it is one opt-in switch, and people who don't want infrared don't pay the build cost:

```nix
hardware.dell-xps-14-da14260.irCamera.enable = true;
```

**Face unlock (Howdy).** A second switch sets up Howdy, the Linux counterpart to Windows Hello, end to end:

```nix
hardware.dell-xps-14-da14260.irCamera.howdy.enable = true;
```

It reads the infrared image directly (the normal tools mishandle this sensor's raw format), switches the infrared light on and off around each scan, and is set up as a *convenience* — a failed face scan always falls back to your password, so no lock-out. Result: face unlock at the login screen, on `sudo`, and on the lock screen.

**For reviewers:**
- Infrared is gated to kernel 7.1.x by an assertion (reasoning in the message).
- Verified on two machines: this DA14260 and an XPS 16 DA16260.
- The Howdy wiring only activates when `services.howdy` exists in nixpkgs (early 2026 onwards); older channels evaluate unchanged.
- `nix run .#run-tests` passes on both channels.

The RGB camera, audio and microphone are unchanged from the last review.

cooparo:
> We can also import speaker tunings from omarchy. see [basecamp/omarchy@aa9f0c5](https://github.com/basecamp/omarchy/commit/aa9f0c54c5c3bd8141f1d30ecf52e6b377a45fe5) . This has made my speakers much better than stock and quite close to macbook levels.

I can try now to address this

cooparo:
## Speaker tuning ported — needs ears

@HritwikSinghal pushed the omarchy tuning you suggested, as an opt-in option:

```nix
hardware.dell-xps-14-da14260.speakerTuning.enable = true;
```

Same graph as basecamp/omarchy@aa9f0c5 (13 biquads + LSP limiter), in front of the internal speaker sink. Off by default.

**It works, but I can't tell if it sounds better.** That's why I'm asking. The plumbing is tested — the sink appears, becomes the default output, passes audio. The voicing isn't: I haven't A/B'd it and I don't trust my ears after a day in the config.

Two things I'd love you to check, since you have the reference in your ears:

1. **Does it sound like yours on Omarchy?**
2. **Does the volume drop match?** I measure the chain at **~9 dB quieter** than bypass (−4 dB from the EQ, −5.3 dB from the limiter's `g_in`). So you have to turn it up to compare fairly. If yours *doesn't* feel that much quieter, our setups differ and that's a bug worth finding rather than a matter of taste.

Two known quirks, both documented: a second output device appears ("Laptop Speakers"), and its volume compounds with the raw sink's — set the raw one to 100% once.

A WirePlumber smart filter (`node.software-dsp` + `hide-parent`, as the Framework 13 profile uses) would remove both, but omarchy reports it passing audio through unprocessed on PipeWire 1.6.8 / WirePlumber 0.5.15 — the versions I'm on. I took their word for it; left as a follow-up. If anyone has it actually working on 1.6.x, I'll happily rewrite it that way.

PR #1933, opened 2026-07-03
apple/t2: improve firmware extraction

mkorje:
###### Description of changes

It turns out 7z can extract the macOS recovery image in userspace, so we can get rid of the linux VM code (which could fail occasionally).

###### Things done


soopyc (approved):

soopyc:
built successfully and looks correct in vm.

Issue #1946, opened 2026-07-14
raspberry-pi: unify device-tree overlays


JamieMagee:
Raspberry Pi support currently has two incompatible overlay paths. `hardware.deviceTree.overlays` merges overlays into kernel DTBs at build time, while `configtxt.settings` asks Raspberry Pi firmware to apply them at boot. Loading a generation DTB drops firmware overlays and runtime fixups, passing through the firmware DT ignores build-time overlays.

Make Raspberry Pi firmware the only overlay path. Add an ordered `hardware.raspberry-pi.configtxt.overlays` API for stock and custom overlays, conditional filters, and scoped parameters. Compile custom DTS sources to DTBOs, then stage them with raw kernel DTBs, the stock overlay directory, firmware, U-Boot, and generated `config.txt` as one managed bundle.

Migrate the Pi 4 modules to the new API, then remove the custom `dtmerge` backend and Pi-specific `hardware.deviceTree.overlays` usage.

sorki:
I would love if the old path remained as is / backwards compatible. Lots of users already use that, it works well and is as close to upstream way of doing things as possible.

doronbehar:
> upstream way of doing things

What upstream do you refer to exactly?

doronbehar:
Anything else left here?

JamieMagee:
I'm still planning to migrate the rest of the custom overlays under [`raspberry-pi/4`](https://github.com/NixOS/nixos-hardware/tree/master/raspberry-pi/4).

PR #1947, opened 2026-07-14
raspberry-pi: add ordered config.txt overlays

JamieMagee:
###### Description of changes

Adds `hardware.raspberry-pi.configtxt.deviceTreeOverlays`, so an overlay can be declared together with the parameters scoped to it. Conditional filters nest exactly as they do in `settings`, and the leaf is an ordered list naming one overlay per element.

```nix
hardware.raspberry-pi.configtxt.deviceTreeOverlays.pi4 = [
  { dwc2.dr_mode = "host"; }
  {
    gpio-fan = {
      gpiopin = 14;
      temp = 80000;
    };
  }
];
```

```ini
[all]
[pi4]
dtoverlay=dwc2
dtparam=dr_mode=host
dtoverlay=
dtoverlay=gpio-fan
dtparam=gpiopin=14
dtparam=temp=80000
dtoverlay=
```

`settings` groups values by key, so it cannot keep an overlay next to the `dtparam` lines belonging to it. That matters because a `dtparam` applies to the most recently loaded overlay, and an overlay's parameters stay in scope only until the next one loads. Order between overlays can matter too, since one may build on another.

Parameters become separate `dtparam` lines rather than being appended to the `dtoverlay` line, which keeps them clear of the line length limit. Booleans render as `on`/`off`, and a `null` parameter is dropped so `mkForce` can remove one.

###### Fixes an existing scope bug

Splitting overlays out of `settings` also fixes a bug in the shipped defaults. `dtoverlay` sorts before `dtparam`, so grouping by key emitted `dtoverlay=vc4-kms-v3d` ahead of `dtparam=audio=on`. But `vc4-kms-v3d` exports its own `audio` parameter, and the firmware resolves a `dtparam` against the most recently loaded overlay before falling back to the base DTB. That line has therefore been configuring HDMI audio rather than enabling the onboard audio it was meant to. Overlays now render after `settings`, which keeps base parameters out of any overlay's scope.

Part of #1946.

###### Things done

### DIFF
--- a/raspberry-pi/README.md
+++ b/raspberry-pi/README.md
@@ -49,11 +49,11 @@ Board profiles import `hardware.raspberry-pi.configtxt`, which renders `config.t
 {
   hardware.raspberry-pi.configtxt.settings = {
     all = {
-      dtparam = [ "audio=on" ];
-      dtoverlay = [
-        "vc4-kms-v3d"
-        "disable-bt"
+      dtparam = [
+        "audio=on"
+        "i2c_arm=on"
       ];
+      disable_overscan = true;
     };
     pi5.arm_freq = 2400;
     cm4.otg_mode = true;
@@ -61,14 +61,60 @@ Board profiles import `hardware.raspberry-pi.configtxt`, which renders `config.t
 }
 ```
 
-List values become repeated keys in the rendered file, so the `dtoverlay` above expands to:
+List values become repeated keys in the rendered file, so the `dtparam` above expands to:
 
-```
-dtoverlay=vc4-kms-v3d
-dtoverlay=disable-bt
+```ini
+dtparam=audio=on
+dtparam=i2c_arm=on
 ```
 
 Top-level attrs are conditional sections (`all`, `pi4`, `pi5`, `cm4`, and so on). Nesting stacks filters. To drop a default, set the key to `null` with `mkForce`.
+
+Every rendered group opens with `[all]` and then its own filters. Filters of different types stack rather than replace, so `[all]` is what clears the previous group before the next one starts.
+
+The firmware applies these filters at boot, not Nix at evaluation time. This matters in two cases. One image can boot on more than one board, such as a CM4 and a CM5 swapped into the same carrier. Filters like `[EDID=...]`, `[gpio4=1]`, `[tryboot]` and `[bootvar0=42]` also depend on state that Nix cannot see: the attached monitor, a jumper, an EEPROM value.
+
+### Device tree overlays
+
+Overlays go in `configtxt.deviceTreeOverlays`, not in a `dtoverlay` key under `settings`. Filters nest the same way, and the leaf is an ordered list where each element names one overlay and its parameters:
+
+```nix
+{
+  hardware.raspberry-pi.configtxt.deviceTreeOverlays.pi4 = [
+    { dwc2.dr_mode = "host"; }
+    {
+      gpio-fan = {
+        gpiopin = 14;
+        temp = 80000;
+      };
+    }
+  ];
+}
+```
+
+This renders after `configtxt.settings` as:
+
+```ini
+[all]
+[pi4]
+dtoverlay=dwc2
+dtparam=dr_mode=host
+dtoverlay=
+dtoverlay=gpio-fan
+dtparam=gpiopin=14
+dtparam=temp=80000
+dtoverlay=
+```
+
+A separate option is necessary because `settings` groups values by key. That grouping cannot keep an overlay next to the `dtparam` lines that belong to it. The order is functional. A `dtparam` applies to the overlay that loaded last, and the parameters of an overlay stay in scope only until the next overlay loads. The bare `dtoverlay=` line closes that scope, so later parameters apply to the base device tree. The order between overlays can also matter, because one overlay can build on another.
+
+Overlays render after `settings`. This order keeps base parameters such as `dtparam=audio=on` out of the scope of an overlay. An overlay can export a parameter with the same name as a base parameter. While that overlay is in scope, the firmware uses the parameter of the overlay.
+
+Each parameter becomes its own `dtparam` line rather than an addition to the `dtoverlay` line, which keeps them clear of the 98-character line limit. Booleans render as `on` or `off`. Use `null` with `mkForce` to remove a parameter.
+
+The module concatenates lists from separate modules, but the order is not the order of definition. If one overlay must load before another, set the order with `mkBefore` or `mkAfter`.
+
+To supply your own file, set `configtxt.file`. The module then ignores `settings` and `deviceTreeOverlays`.
 
 ## Current limits
 
--- a/raspberry-pi/common/config-txt.nix
+++ b/raspberry-pi/common/config-txt.nix
@@ -43,14 +43,14 @@ let
     path: value:
     if value == null then
       [ ]
-    else if builtins.isAttrs value && !(builtins.isList value) && !(value ? _type) then
-      lib.flatten (lib.mapAttrsToList (name: recurse ([ name ] ++ path)) value)
-    else
+    else if !(builtins.isAttrs value) then
       {
         conditionals = lib.sort builtins.lessThan (lib.filter (k: k != "all") (lib.tail path));
         name = lib.head path;
         inherit value;
-      };
+      }
+    else
+      lib.flatten (lib.mapAttrsToList (name: recurse ([ name ] ++ path)) value);
 
   # Group flattened items by their conditional filter set, then render.
   groupItems =
--- a/raspberry-pi/common/config-txt.nix
+++ b/raspberry-pi/common/config-txt.nix
@@ -82,8 +82,16 @@ in
   options.hardware.raspberry-pi.configtxt = {
     settings = lib.mkOption {
       type =
-        with lib.types;
         let
+          inherit (lib.types)
+            attrsOf
+            bool
+            int
+            listOf
+            nullOr
+            oneOf
+            str
+            ;
           atom = nullOr (oneOf [
             str
             int
--- a/raspberry-pi/common/config-txt.nix
+++ b/raspberry-pi/common/config-txt.nix
@@ -1,9 +1,13 @@
 # config.txt generation module for Raspberry Pi
 #
-# Generates config.txt from structured Nix options. The type system models
-# config.txt as nested attrs: each nesting level adds a conditional filter,
-# and leaves are config values. Repeated keys (dtparam, dtoverlay, gpio)
-# are supported via lists.
+# This module renders config.txt from structured Nix options. The type models
+# config.txt as nested attrs: each level of nesting adds a conditional filter,
+# and the leaves are config values. Lists produce repeated keys for dtparam,
+# dtoverlay and gpio.
+#
+# Device tree overlays have their own option. A dtparam applies to the overlay
+# that loaded last, and values grouped by key cannot express that order.
+# Overlays render after settings, so base parameters apply first.
 #
 # Based on work from nvmd/nixos-raspberrypi (MIT License) and rendering
 # approach suggested by @quentinmit. Follows RFC 42.
@@ -35,48 +39,156 @@ let
 
   mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
 
-  # Recursively flatten nested attrs into a list of { conditionals, name, value } records.
-  # Each nesting level (except the leaf) is a conditional filter.
-  # "all" filters are omitted from the conditionals list since they reset state.
-  # null values are filtered out, allowing `mkForce null` to remove defaults.
-  recurse =
+  # Overlay parameters use on/off for booleans. This follows the Raspberry Pi
+  # documentation and the overlay README. Plain settings keep 1/0.
+  mkParamValue = v: if builtins.isBool v then (if v then "on" else "off") else mkValueString v;
+
+  # An "all" filter resets state rather than narrowing it, so it never has to be
+  # recorded. The rest are sorted, which makes two paths that stack the same
+  # filters in a different order compare equal.
+  mkConditionals =
+    path:
+    lib.pipe path [
+      (lib.filter (k: k != "all"))
+      (lib.sort builtins.lessThan)
+    ];
+
+  # Flatten settings into { conditionals, items } groups, one group per leaf.
+  # Every level above the leaf is a conditional filter. A null value drops out,
+  # so `mkForce null` removes a default.
+  recurseSettings =
     path: value:
     if value == null then
       [ ]
     else if !(builtins.isAttrs value) then
-      {
-        conditionals = lib.sort builtins.lessThan (lib.filter (k: k != "all") (lib.tail path));
-        name = lib.head path;
-        inherit value;
-      }
+      [
+        {
+          conditionals = mkConditionals (lib.init path);
+          items = [
+            {
+              name = lib.last path;
+              inherit value;
+            }
+          ];
+        }
+      ]
     else
-      lib.flatten (lib.mapAttrsToList (name: recurse ([ name ] ++ path)) value);
+      lib.pipe value [
+        (lib.mapAttrsToList (name: recurseSettings (path ++ [ name ])))
+        lib.concatLists
+      ];
 
-  # Group flattened items by their conditional filter set, then render.
-  groupItems =
-    items:
-    lib.mapAttrsToList (groupJSON: groupItems: {
-      conditionals = builtins.fromJSON groupJSON;
-      items = builtins.listToAttrs groupItems;
-    }) (builtins.groupBy (x: builtins.toJSON x.conditionals) items);
+  # The same flattening for overlays. Here the leaf is already the ordered
+  # overlay list for that filter set, so every level of nesting is a filter.
+  recurseOverlays =
+    path: value:
+    if value == null then
+      [ ]
+    else if builtins.isList value then
+      [
+        {
+          conditionals = mkConditionals path;
+          items = value;
+        }
+      ]
+    else
+      lib.pipe value [
+        (lib.mapAttrsToList (name: recurseOverlays (path ++ [ name ])))
+        lib.concatLists
+      ];
 
-  mkGroup =
+  # Merge groups that share a filter set. Two paths can reduce to the same set
+  # (all.pi4 and pi4.all both yield [ "pi4" ]), so merging is needed even though
+  # each recursion returns one group per leaf.
+  #
+  # Sorting first puts equal sets next to each other, so a single pass coalesces
+  # them. Folding from the right lets each step prepend, which keeps the result
+  # in order without a second pass. builtins.sort is stable, so items keep their
+  # original relative order, which overlays depend on.
+  #
+  # The sort is not redundant with mkConditionals, which only orders the filters
+  # within one group. Groups arrive in attrs walk order, driven by the raw names
+  # before "all" is stripped, so all.z.foo walks before b.bar.
+  groupByConditionals =
+    groups:
+    lib.pipe groups [
+      (lib.sort (a: b: a.conditionals < b.conditionals))
+      (lib.foldr (
+        group: merged:
+        let
+          next = lib.head merged;
+        in
+        if merged != [ ] && next.conditionals == group.conditionals then
+          [ (group // { items = group.items ++ next.items; }) ] ++ lib.tail merged
+        else
+          [ group ] ++ merged
+      ) [ ])
+    ];
+
+  # Every group opens with [all] plus its own filters. For settings that leading
+  # [all] is only a reset. For overlays it is required: a bare `dtoverlay=` ends
+  # overlay scope but leaves the conditional filters set, and only [all] clears
+  # them.
+  mkFilterHeader = conditionals: lib.concatMapStrings (f: "[${f}]\n") ([ "all" ] ++ conditionals);
+
+  renderSettingsGroup =
     group:
-    lib.concatMapStrings (k: "[${k}]\n") group.conditionals
-    + lib.generators.toKeyValue {
-      inherit mkKeyValue;
-      listsAsDuplicateKeys = true;
-    } group.items;
+    mkFilterHeader group.conditionals
+    + lib.pipe group.items [
+      builtins.listToAttrs
+      (lib.generators.toKeyValue {
+        inherit mkKeyValue;
+        listsAsDuplicateKeys = true;
+      })
+    ];
 
-  toConfigTxt =
-    attrs:
-    let
-      groups = lib.sort (a: b: a.conditionals < b.conditionals) (
-        groupItems (lib.flatten (recurse [ ] attrs))
-      );
-    in
-    lib.concatMapStringsSep "\n[all]\n" mkGroup groups;
+  # Each entry ends with a bare `dtoverlay=`. This terminator closes the parameter
+  # scope of that overlay, so a later dtparam applies to the base DTB instead. A
+  # terminator must never come first in the file, where it means "load no HAT
+  # overlay". Settings render first, and every terminator here follows a
+  # dtoverlay=<name>, so no terminator comes first today. A change to the section
+  # order can break this guarantee.
+  renderOverlayGroup =
+    group:
+    mkFilterHeader group.conditionals
+    + lib.concatMapStrings (
+      entry:
+      let
+        name = lib.head (lib.attrNames entry);
+      in
+      lib.concatMapStrings (line: line + "\n") (
+        [ "dtoverlay=${name}" ]
+        ++ lib.pipe entry.${name} [
+          (lib.filterAttrs (_: v: v != null))
+          (lib.mapAttrsToList (p: v: "dtparam=${p}=${mkParamValue v}"))
+        ]
+        ++ [ "dtoverlay=" ]
+      )
+    ) group.items;
 
+  render =
+    conf: renderSectionFunc: recurseFunc:
+    lib.pipe conf [
+      (recurseFunc [ ])
+      groupByConditionals
+      (map renderSectionFunc)
+      (lib.concatStringsSep "\n")
+    ];
+
+  # Settings render first, so base DT parameters apply before any overlay loads.
+  # The overlay groups follow. A blank line separates the two sections, and the
+  # filter keeps that separator from becoming a stray leading or trailing line
+  # when one of the two is empty.
+  generatedConfig =
+    lib.pipe
+      [
+        (render cfg.settings renderSettingsGroup recurseSettings)
+        (render cfg.deviceTreeOverlays renderOverlayGroup recurseOverlays)
+      ]
+      [
+        (lib.filter (section: section != ""))
+        (lib.concatStringsSep "\n")
+      ];
 in
 {
   options.hardware.raspberry-pi.configtxt = {
@@ -115,6 +227,10 @@ in
         Booleans render as `0`/`1`. Use `null` with `mkForce` to remove a
         default.
 
+        Set device tree overlays through {option}`deviceTreeOverlays` rather
+        than a `dtoverlay` key here. A `dtoverlay` written here gets no
+        terminator, so it captures every `dtparam` that follows it.
+
         See <https://www.raspberrypi.com/documentation/computers/config_txt.html>
       '';
       example = lib.literalExpression ''
@@ -131,13 +247,95 @@ in
       '';
     };
 
+    deviceTreeOverlays = lib.mkOption {
+      type =
+        let
+          inherit (lib.types)
+            addCheck
+            attrsOf
+            bool
+            int
+            listOf
+            nullOr
+            oneOf
+            str
+            ;
+          params = attrsOf (
+            nullOr (oneOf [
+              str
+              int
+              bool
+            ])
+          );
+          # Each list element names one overlay. Without this check one element
+          # can name two overlays, as in [ { a = { }; b = { }; } ]. Nix sorts
+          # attribute names, so the order between them is lost.
+          entry = addCheck (attrsOf params) (e: builtins.length (builtins.attrNames e) == 1) // {
+            description = "attribute set naming exactly one overlay, mapped to its parameters";
+          };
+          node = oneOf [
+            (listOf entry)
+            (attrsOf node)
+          ];
+        in
+        attrsOf node // { description = "config.txt device tree overlay type"; };
+      default = { };
+      description = ''
+        Device tree overlays for the Raspberry Pi firmware, rendered in list
+        order.
+
+        Keys are conditional filter sections and nest exactly like
+        {option}`settings`. The leaf is an ordered list, and each element names
+        one overlay mapped to its parameters.
+
+        Use this instead of a `dtoverlay` key in {option}`settings` when an
+        overlay has parameters. `settings` groups values by key and cannot keep
+        an overlay next to the `dtparam` lines that belong to it. A parameter
+        stays in scope only until the next overlay loads.
+
+        Each parameter becomes its own `dtparam` line rather than an addition to
+        the `dtoverlay` line, which keeps them clear of the 98-character limit
+        on `config.txt` lines. Booleans render as `on` or `off`. Use `null` with
+        `mkForce` to remove a default.
+
+        Overlays render after {option}`settings`, so base device tree parameters
+        set there apply before any overlay loads. The example renders as:
+
+        ```ini
+        [all]
+        [pi4]
+        dtoverlay=dwc2
+        dtparam=dr_mode=host
+        dtoverlay=
+        dtoverlay=gpio-fan
+        dtparam=gpiopin=14
+        dtparam=temp=80000
+        dtoverlay=
+        ```
+      '';
+      example = lib.literalExpression ''
+        {
+          pi4 = [
+            { dwc2.dr_mode = "host"; }
+            {
+              gpio-fan = {
+                gpiopin = 14;
+                temp = 80000;
+              };
+            }
+          ];
+        }
+      '';
+    };
+
     file = lib.mkOption {
       type = lib.types.path;
-      default = pkgs.writeText "config.txt" (toConfigTxt cfg.settings);
-      defaultText = lib.literalExpression ''pkgs.writeText "config.txt" (generated from settings)'';
+      default = pkgs.writeText "config.txt" generatedConfig;
+      defaultText = lib.literalExpression ''pkgs.writeText "config.txt" (generated from settings and deviceTreeOverlays)'';
       description = ''
-        Path to the generated config.txt file. Defaults to the rendered output
-        of `settings`, but can be overridden to supply a custom config.txt.
+        Path to the config.txt file to install. Defaults to the rendered output
+        of {option}`settings` and {option}`deviceTreeOverlays`. A custom file
+        replaces that output, so the module then ignores both options.
       '';
     };
   };
--- a/raspberry-pi/common/config-txt.nix
+++ b/raspberry-pi/common/config-txt.nix
@@ -31,7 +31,7 @@ let
     else if false == v then
       "0"
     else
-      builtins.abort "config.txt: unsupported value type: ${builtins.typeOf v}";
+      abort "config.txt: unsupported value type: ${builtins.typeOf v}";
 
   mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
 
--- a/tests/config-txt-tests.nix
+++ b/tests/config-txt-tests.nix
@@ -1,0 +1,246 @@
+# Rendering tests for config.txt. Board profile builds only prove the module
+# evaluates, not that the output is right.
+#
+# Each case diffs the generated file against an expected one at build time,
+# which avoids import-from-derivation.
+#
+# nix build .#checks.<system>.raspberry-pi-config-txt
+
+{
+  lib,
+  pkgs,
+  runCommand,
+  writeText,
+}:
+
+let
+  configtxt = ../raspberry-pi/common/config-txt.nix;
+
+  # Render without a full NixOS evaluation.
+  render =
+    module:
+    (lib.evalModules {
+      modules = [
+        configtxt
+        { _module.args = { inherit pkgs; }; }
+        module
+      ];
+    }).config.hardware.raspberry-pi.configtxt.file;
+
+  cases = {
+    empty = {
+      module = { };
+      expected = "";
+    };
+
+    settingsOnly = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings.all.arm_boost = true;
+      };
+      expected = ''
+        [all]
+        arm_boost=1
+      '';
+    };
+
+    settingsValueTypes = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings.all = {
+          yes = true;
+          no = false;
+          num = 2;
+          text = "abc";
+        };
+      };
+      expected = ''
+        [all]
+        no=0
+        num=2
+        text=abc
+        yes=1
+      '';
+    };
+
+    stackedFilters = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings.pi4."gpio4=1".arm_freq = 1000;
+      };
+      expected = ''
+        [all]
+        [gpio4=1]
+        [pi4]
+        arm_freq=1000
+      '';
+    };
+
+    # "all" is stripped before sorting, so all.z lands after b.
+    groupSortOrder = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings = {
+          all.z.foo = 1;
+          b.bar = 2;
+        };
+      };
+      expected = ''
+        [all]
+        [b]
+        bar=2
+
+        [all]
+        [z]
+        foo=1
+      '';
+    };
+
+    # This is what makes mkForce null remove a default.
+    nullDropsOut = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings.all = {
+          kept = true;
+          dropped = null;
+        };
+      };
+      expected = ''
+        [all]
+        kept=1
+      '';
+    };
+
+    listsRepeatKeys = {
+      module = {
+        hardware.raspberry-pi.configtxt.settings.all.dtparam = [
+          "audio=on"
+          "i2c_arm=on"
+        ];
+      };
+      expected = ''
+        [all]
+        dtparam=audio=on
+        dtparam=i2c_arm=on
+      '';
+    };
+
+    overlayWithParams = {
+      module = {
+        hardware.raspberry-pi.configtxt.deviceTreeOverlays.all = [
+          { dwc2.dr_mode = "host"; }
+        ];
+      };
+      expected = ''
+        [all]
+        dtoverlay=dwc2
+        dtparam=dr_mode=host
+        dtoverlay=
+      '';
+    };
+
+    # Overlay parameters use on/off, unlike settings.
+    overlayBooleans = {
+      module = {
+        hardware.raspberry-pi.configtxt.deviceTreeOverlays.all = [
+          {
+            example = {
+              yes = true;
+              no = false;
+              dropped = null;
+            };
+          }
+        ];
+      };
+      expected = ''
+        [all]
+        dtoverlay=example
+        dtparam=no=off
+        dtparam=yes=on
+        dtoverlay=
+      '';
+    };
+
+    duplicateOverlays = {
+      module = {
+        hardware.raspberry-pi.configtxt.deviceTreeOverlays.all = [
+          { mcp251xfd.interrupt = 25; }
+          { mcp251xfd.interrupt = 24; }
+        ];
+      };
+      expected = ''
+        [all]
+        dtoverlay=mcp251xfd
+        dtparam=interrupt=25
+        dtoverlay=
+        dtoverlay=mcp251xfd
+        dtparam=interrupt=24
+        dtoverlay=
+      '';
+    };
+
+    overlayOrder = {
+      module = {
+        hardware.raspberry-pi.configtxt.deviceTreeOverlays.all = [
+          { first = { }; }
+          { second = { }; }
+        ];
+      };
+      expected = ''
+        [all]
+        dtoverlay=first
+        dtoverlay=
+        dtoverlay=second
+        dtoverlay=
+      '';
+    };
+
+    # all.pi4 and pi4.all are the same filter set, so they merge.
+    mergedFilterPaths = {
+      module = {
+        hardware.raspberry-pi.configtxt.deviceTreeOverlays = {
+          all.pi4 = [ { first = { }; } ];
+          pi4.all = [ { second = { }; } ];
+        };
+      };
+      expected = ''
+        [all]
+        [pi4]
+        dtoverlay=first
+        dtoverlay=
+        dtoverlay=second
+        dtoverlay=
+      '';
+    };
+
+    # dtoverlay sorts before dtparam, so rendering settings first is what keeps a
+    # base parameter out of an overlay's scope.
+    settingsRenderBeforeOverlays = {
+      module = {
+        hardware.raspberry-pi.configtxt = {
+          settings.all.dtparam = [ "audio=on" ];
+          deviceTreeOverlays.all = [ { vc4-kms-v3d = { }; } ];
+        };
+      };
+      expected = ''
+        [all]
+        dtparam=audio=on
+
+        [all]
+        dtoverlay=vc4-kms-v3d
+        dtoverlay=
+      '';
+    };
+  };
+
+  # One derivation per case, so a single test can be built on its own.
+  tests = lib.mapAttrs (
+    name: case:
+    runCommand "config-txt-${name}" { } ''
+      diff -u ${writeText "${name}-expected.txt" case.expected} ${render case.module}
+      touch $out
+    ''
+  ) cases;
+in
+# Building this builds every case, because naming them here makes them inputs.
+# passthru re-exposes each one, so a single case can be built on its own with
+# `nix build .#checks.<system>.raspberry-pi-config-txt.<name>`.
+runCommand "raspberry-pi-config-txt-tests" {
+  cases = lib.attrValues tests;
+  passthru = tests;
+} "touch $out"
--- a/flake.nix
+++ b/flake.nix
@@ -553,6 +553,9 @@
         pkgs.lib.optionalAttrs (self.formatter ? ${system}) {
           formatting = treefmtEval.config.build.check self;
         }
+        // {
+          raspberry-pi-config-txt = pkgs.callPackage ./tests/config-txt-tests.nix { };
+        }
         // nixosTests
       );
     };
--- a/raspberry-pi/common/config-txt-defaults.nix
+++ b/raspberry-pi/common/config-txt-defaults.nix
@@ -8,30 +8,37 @@
 { lib, ... }:
 
 {
-  hardware.raspberry-pi.configtxt.settings = {
-    all = {
-      camera_auto_detect = lib.mkDefault true;
-      display_auto_detect = lib.mkDefault true;
-      max_framebuffers = lib.mkDefault 2;
-      disable_fw_kms_setup = lib.mkDefault true;
-      disable_overscan = lib.mkDefault true;
-      arm_boost = lib.mkDefault true;
-      dtparam = lib.mkDefault [ "audio=on" ];
-      dtoverlay = lib.mkDefault [ "vc4-kms-v3d" ];
-      # U-Boot needs the UART enabled to boot (see arch/arm/mach-bcm283x/Kconfig
-      # in the U-Boot tree). Without it U-Boot hangs on the Pi 4.
-      enable_uart = lib.mkDefault true;
+  hardware.raspberry-pi.configtxt = {
+    settings = {
+      all = {
+        camera_auto_detect = lib.mkDefault true;
+        display_auto_detect = lib.mkDefault true;
+        max_framebuffers = lib.mkDefault 2;
+        disable_fw_kms_setup = lib.mkDefault true;
+        disable_overscan = lib.mkDefault true;
+        arm_boost = lib.mkDefault true;
+        # A base DTB parameter rather than an overlay one, so it stays here. It
+        # has to apply before any overlay loads: vc4-kms-v3d exports its own
+        # `audio` parameter, which would otherwise capture this line and set
+        # HDMI audio instead of enabling the onboard audio.
+        dtparam = lib.mkDefault [ "audio=on" ];
+        # U-Boot needs the UART enabled to boot (see arch/arm/mach-bcm283x/Kconfig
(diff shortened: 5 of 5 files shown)


JamieMagee:
@doronbehar could I ask for a review of this please, if you have some time? It's the first part in a series to fix Raspberry Pi device trees once and for all 😅 

I do really appreciate all the time and effort you've put into to reviewing my changes. Thank you!

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +98,47 @@ let
...
+    overlay:
+    lib.concatStringsSep "\n" (
+      [ "[all]" ]
+      ++ map (filter: "[${filter}]") overlay.filters
+      ++ [ "dtoverlay=${overlay.name}" ]
+      ++ map (param: "dtparam=${param}") overlay.params
+      ++ [ "dtoverlay=" ]
+    );
+
+  legacyConfig = toConfigTxt cfg.settings;
```
doronbehar: Why call this legacy?

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +98,47 @@ let
...
+    lib.filter (overlay: !validOverlayName overlay.name) cfg.overlays
+  );
+  invalidFilters = lib.concatMap (
+    overlay: lib.filter (filter: !validFilter filter) overlay.filters
+  ) cfg.overlays;
+  invalidParams = lib.concatMap (overlay: lib.filter hasLineBreak overlay.params) cfg.overlays;
+  longLines = lib.filter (line: builtins.stringLength line > 98) (
+    lib.splitString "\n" generatedConfig
+  );
+  usesGeneratedFile = toString cfg.file == toString generatedFile;
```
doronbehar: This whole code is a bit hard to read and I haven't delved to it yet. Do you think you might be able to improve it with `lib.pipe`? And please, more comments maybe?

raspberry-pi/common/config-txt.nix
JamieMagee: I’ve renamed this to `settingsConfig`.

raspberry-pi/common/config-txt.nix
JamieMagee: I’ve split the rendering and validation into `lib.pipe` stages and added comments for settings compatibility, filter reset, validation errors, the line limit, and custom files.

raspberry-pi/common/config-txt.nix
```diff
@@ -123,14 +204,59 @@ in
...
 
+    overlays = lib.mkOption {
+      type = lib.types.listOf overlayType;
+      default = [ ];
+      description = ''
+        Ordered Raspberry Pi firmware overlays.
+
+        Entries are rendered after {option}`settings`. Each one writes its
+        filters, `dtoverlay`, and `dtparam` values in order, then resets the
+        overlay scope.
```
doronbehar: Could you please explain here why someone would need `hardware.raspberry-pi.configtxt.overlays` and not be satisfied with `hardware.raspberry-pi.configtxt.settings`? I kind of failed to understand that from https://github.com/NixOS/nixos-hardware/issues/1946 ... Also, could you show how the Nix expression in the `example` is eventually rendered in the `config.txt` file? It'd help understand the structure of the expected Nix expressions.

JamieMagee: `settings` is fine for ordinary values and simple overlays, including parameters placed on the same `dtoverlay` line. Its limitation is that the renderer groups values by key, so it cannot preserve an interleaved sequence such as:

```ini
dtoverlay=first
dtparam=...
dtoverlay=second
dtparam=...
```

That order matters because each `dtparam` applies to the most recently loaded overlay. Overlay application order can matter too.

`configtxt.overlays` uses a list, so each overlay stays with its filters and parameters and is rendered in list order. It also resets the overlay scope after each entry.

I've expanded the option description and README with a two-overlay example and the exact generated `config.txt`.

raspberry-pi/README.md
```diff
@@ -63,13 +63,54 @@ Board profiles import `hardware.raspberry-pi.configtxt`, which renders `config.t
...
+    }
+    {
+      name = "gpio-fan";
+      filters = [ "pi4" ];
+      params = [
+        "gpiopin=14"
+        "temp=80000"
+      ];
+    }
+  ];
```
doronbehar: I don't like how the `[<sec>]` filters have winded up in a `filters` attribute inside each `overlay`. Also, I don't see a reason for a NixOS module to make such an effort to filter any device, since we can do that straight with Nix and not with `config.txt`'s specification etc. I also personally find it a bit ugly how you have multiple `[<sec>]` lines in that example, although I understand that "it works".

In fact, I tend to suggest that we should even remove the `configtxt.settings.all` v.s e.g `configtxt.settings.pi4` etc, and simply put everything under `[all]`, and instruct users to manage their filters with Nix.

I understand the need for specifying device tree overlays and parameters in conjunction, and how that's not possible with the current API. However, I don't understand why is it have to be ordered. If the order doesn't matter, and the only reason to have an ordered list is the fact you need `dtoverlay=` and then a `dtparam=` line, then you could support that via a special `configtxt.settings` key (assuming `[all]` was flatten already). Here's an example (again, assuming `[all]` was flatten):

```nix
configtxt.settings = {
  arm_boost = true;
  disable_overscan = true;
  deviceTreeOverlays = {
    "vc4-kms-v3d" = {
      audio = true;
    };
    dwc2 = {
      dr_mode = "host";
    };
    gpio-fan = {
      gpiopin = 14;
      temp = 80000;
    };
  };
};
```

Which will render as:

```ini
arm_boost=on
disable_overscan=on
dtoverlay=dwc2
dtparam=dr_mode=host
dtoverlay=
dtoverlay=gpio-fan
dtparam=gpiopin=14
dtparam=temp=80000
dtoverlay=
```

Also TBH I'm not sure I understand the `dtoverlay=` (no value) lines, but I guess this is justified (please document that in a comment).

If you think the above design is good, here's how I'd plan to implement it:

1. Make this PR a draft.
2. Create a PR that will simply flatten any `configtxt.settings.${filter}` into `configtxt.settings`, and explain that everything is being put into `[all]` in the docs, and that Nix allows users to make their own filters etc hence it is not needed. Since the `configtxt.settings` API hasn't been in nixos-hardware for a while, I don't think it is too bad for it to break.
3. Implement this PR by simply taking special care of the `deviceTreeOverlays` attribute in `configtxt.settings`. Make `configtxt.settings.dtoverlay` and `configtxt.settings.dtparam` raise an evaluation error telling the user to use `deviceTreeOverlays`.

doronbehar (changes requested):

JamieMagee:
@quentinmit, if you have time, could you take a look at this API discussion? You proposed the recursive conditional-filter representation in #1788 and commented that `dtparam` and `dtoverlay` might eventually need a higher-level configuration surface. We are deciding whether overlay applications should remain an ordered list with optional firmware-time filters, or use a named attrset and flatten conditional sections. Your perspective would be helpful.

doronbehar:
@JamieMagee if you ask me, if @quentinmit won't reply in a week or so, we should go ahead and start implementing at least points 1 & 2 in [the plan I outlined before](https://github.com/NixOS/nixos-hardware/pull/1947#discussion_r3612932166).

raspberry-pi/README.md
JamieMagee: Three of your points I agree with, but two I don't. This response ended up being a little long, so I've tried to break it up into chunks.

---

I agree that `params = [ "dr_mode=host" ]` is clumsy. `dr_mode = "host"` reads better and lets me reuse the same value coercion as `settings`, so ints and bools render identically in both places.

The bare `dtoverlay=` does require a little more explanation. From the firmware docs[^1][^2]:

> As a special case, if called with no value – `dtoverlay=` – the option marks the end of a list of overlay parameters. If used before any other dtoverlay or dtparam setting, it prevents the loading of any HAT overlay.

> Overlay parameters are only in scope until the next overlay is loaded. [...] To expose the parameter exported by the base DTB instead, end the current overlay scope using: `dtoverlay=`

So the terminator is what stops a later `dtparam` binding to the wrong overlay. The second part is a bit of a footgun: a bare `dtoverlay=` before any other overlay line silently disables HAT autoloading. That can't happen today because `settings` always renders first, but it will bite whoever changes the render order.

The `[all]` clutter is fair. The renderer emits the reset unconditionally, which is lazy. I'll see if I can rewrite it to emit `[all]` only when the previous entry actually left a filter set, and drop the filter lines when the list is empty. An unfiltered overlay would then only render its `dtoverlay=` and `dtparam=` lines.

---

Regarding keying by overlay name, an attrset can't express the same overlay loaded twice. From the overlays README[^3]:

> When you have additional hardware with more than one device of a kind, you end up using the same overlay multiple times with other parameters, e.g.
>
> ```ini
> dtoverlay=mcp251xfd,spi0-0,interrupt=25
> dtoverlay=mcp251xfd,spi0-1,interrupt=24
> ```

An attrset keyed on `mcp251xfd` would only get one of them.

There's a smaller problem as well. Nix sorts attribute names, so an attrset also gives up control of output order:

```sh
$ nix-instantiate --eval -E 'builtins.attrNames { vc4-kms-v3d = 1; dwc2 = 2; gpio-fan = 3; }'
[ "dwc2" "gpio-fan" "vc4-kms-v3d" ]
```

Your example writes `vc4-kms-v3d` first and it would render third.

---

Model filters I'll concede. Nix knows the board at eval time, so `[pi4]` and `[pi5]` buy us nothing that an `if` doesn't.

But a lot aren't known until boot, so Nix can't stand in for them:

- `[EDID=DEL-DELL_U2422H]` which monitor is plugged in
- `[gpio4=1]` a jumper or switch
- `[bootvar0=42]` and `[cust_otp0&0x1]` EEPROM and OTP values
- etc.

On API shape, @quentinmit's position in #1788 was close to the opposite of this proposal. He asked for arbitrary nested `attrsOf` [specifically so combined filters were expressible][1], and [separately argued][2] that overlays belong beside `settings` rather than inside it:

> I think it would be more in keeping with RFC 42 to keep `settings` a direct map to the `config.txt` file, and put `dt-params` one level up.

Nesting `deviceTreeOverlays` under `settings` means the recursion special-cases one key, and `settings` stops being a direct mapping of the file.

---

I'll do the mechanical parts now: attrset params, the `dtoverlay=` documentation, and the `[all]` cleanup. I'll mark this as draft while I rework it.

If you still want the filters gone after the above, could we do that as its own PR? It breaks the API that shipped in #1788, and I'd still like to give @quentinmit a change to weigh in.

[1]: https://github.com/NixOS/nixos-hardware/pull/1788#discussion_r2903116334
[2]: https://github.com/NixOS/nixos-hardware/pull/1788#discussion_r2903098763

[^1]: https://www.raspberrypi.com/documentation/computers/config_txt.html#dtoverlay
[^2]: https://www.raspberrypi.com/documentation/computers/configuration.html#part3.2
[^3]: https://github.com/raspberrypi/firmware/blob/78e81e2cd6e00efeb79c169bff29ec457fa14b11/boot/overlays/README#L87-L97

doronbehar: Thanks @JamieMagee for the detailed answer. As for this example:

> ```ini
> dtoverlay=mcp251xfd,spi0-0,interrupt=25
> dtoverlay=mcp251xfd,spi0-1,interrupt=24
> ```

Is this equivalent to this?

```ini
dtoverlay=mcp251xfd
dtparam=spi0-0,interrupt=25
dtoverlay=
dtoverlay=mcp251xfd
dtparam=spi0-1,interrupt=24
dtoverlay=
```

Because with my design from my previous comment, you can render the above long `dtoverlay` lines with:

```nix
hardware.raspberry-pi.configtxt.settings.all = { # let's assume we keep the filters as is now...
  deviceTreeOverlays = {
    "mcp251xfd,spi0-0,interrupt=25" = { };
    "mcp251xfd,spi0-1,interrupt=24" = { };
  };
};
```

> The `[all]` clutter is fair. The renderer emits the reset unconditionally, which is lazy. I'll see if I can rewrite it to emit `[all]` only when the previous entry actually left a filter set, and drop the filter lines when the list is empty. An unfiltered overlay would then only render its `dtoverlay=` and `dtparam=` lines.
> 
> [...]
> 
> Model filters I'll concede. Nix knows the board at eval time, so `[pi4]` and `[pi5]` buy us nothing that an `if` doesn't.
> 
> But a lot aren't known until boot, so Nix can't stand in for them:

Thanks for this info, I wasn't aware of that. Maybe to make things simple, we can keep the filters as attributes of `hardware.raspberry-pi.configtxt.settings.` for now. I don't mind that too much to be honest.

quentinmit: There are a ton of different topics in this thread. Hopefully I've managed to reply to each one at least once:

> Also, I don't see a reason for a NixOS module to make such an effort to filter any device, since we can do that straight with Nix and not with `config.txt`'s specification etc. I also personally find it a bit ugly how you have multiple `[<sec>]` lines in that example, although I understand that "it works".
> 
> In fact, I tend to suggest that we should even remove the `configtxt.settings.all` v.s e.g `configtxt.settings.pi4` etc, and simply put everything under `[all]`, and instruct users to manage their filters with Nix.

> Model filters I'll concede. Nix knows the board at eval time, so `[pi4]` and `[pi5]` buy us nothing that an `if` doesn't.

I don't think it is reasonable to assume that a Nix configuration will be deployed on exactly one single machine with a particular set of unchanging hardware. For example, an SD card image might need to be booted on multiple generations of Pi. (This isn't just hypothetical - I personally have a uConsole in which I swap out a CM4 and CM5.)

> The `[all]` clutter is fair. The renderer emits the reset unconditionally, which is lazy. I'll see if I can rewrite it to emit `[all]` only when the previous entry actually left a filter set, and drop the filter lines when the list is empty. An unfiltered overlay would then only render its `dtoverlay=` and `dtparam=` lines.

I agree it's a little cluttered but I don't think there's a way around it. You would have to parse the individual filters and understand which ones override which other ones in order to figure out when it would be safe to omit `[all]`. Note that RPi's [own documentation recommends](https://www.raspberrypi.com/documentation/computers/config_txt.html#the-all-filter) an `[all]` at the end of each section, which is basically what we have now. (Just at the start of the next section instead of the end of the previous section.)

> I agree that `params = [ "dr_mode=host" ]` is clumsy. `dr_mode = "host"` reads better and lets me reuse the same value coercion as `settings`, so ints and bools render identically in both places.

I also agree with using an attrset for params; AFAIK param order never matters.

> Regarding keying by overlay name, an attrset can't express the same overlay loaded twice. From the overlays README[3](#user-content-fn-3-ea7f946c66aeb8c8af8ee548fd2ac183):
> 
> > When you have additional hardware with more than one device of a kind, you end up using the same overlay multiple times with other parameters, e.g.
> > dtoverlay=mcp251xfd,spi0-0,interrupt=25
> > dtoverlay=mcp251xfd,spi0-1,interrupt=24
> 
> An attrset keyed on `mcp251xfd` would only get one of them.
> 
> There's a smaller problem as well. Nix sorts attribute names, so an attrset also gives up control of output order:

I agree that we need to support duplicate overlays. I believe it's also possible for one overlay to depend on another, which means that ordering matters between overlays with different names. I certainly dislike this because it prevents merging of overlays and parameters across modules. (You can't have one module enable the `audremap` overlay and then another module set the output pin, for instance, because you can't modify the list entry added by the first module.) But I don't really see a way around that. I guess we could provide yet another option that uses an attrset with merge semantics for overlays that have no ordering constraints and are only instantiated once, and a second option that takes a list, but that feels too complicated to me.

> On API shape, @quentinmit's position in #1788 was close to the opposite of this proposal. He asked for arbitrary nested `attrsOf` [specifically so combined filters were expressible](https://github.com/NixOS/nixos-hardware/pull/1788#discussion_r2903116334), and [separately argued](https://github.com/NixOS/nixos-hardware/pull/1788#discussion_r2903098763) that overlays belong beside `settings` rather than inside it:
> 
> > I think it would be more in keeping with RFC 42 to keep `settings` a direct map to the `config.txt` file, and put `dt-params` one level up.
> 
> Nesting `deviceTreeOverlays` under `settings` means the recursion special-cases one key, and `settings` stops being a direct mapping of the file.

This one is extra tricky. I think the obvious way to apply RFC 42's recommendation is to have `deviceTreeOverlays` live outside `settings`, and have the implementation of `deviceTreeOverlays` set the necessary keys in `settings`. But we can't just do that here because the needed configuration (`dtoverlay` and `dtparam` interleaved in the correct sequence) cannot be expressed in terms of the attrset structure in `settings`. (And the parameters can't just be set as `dtoverlay=foo,pin1=a,pin2=b,...` because the lines are limited to 80 characters.)

So I think I agree with this PR that the next best option is to construct `config.txt` from the combination of `settings` and `deviceTreeOverlays`. This pattern is also fairly common in nixpkgs.

> I don't like how the `[<sec>]` filters have winded up in a `filters` attribute inside each `overlay`.

Instead of a single list with `name`, `filters`, and `params` keys, I think we should match the same conditional filter syntax as used by settings. How about using the overlay name as a key and the parameters as the value ala `lib.types.attrTag`? For example,

```nix
hardware.raspberry-pi.configtxt = {
    settings.pi4."gpio4=1".audio_pwm_mode = 1;
    deviceTreeOverlays.pi4."gpio4=1" = [
      { audremap.enable_jack = true; }
      { ghostamp = {}; }
    ];
};
```

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +98,66 @@ let
...
+    (map (overlay: overlay.name))
+  ];
+  invalidFilters = lib.pipe cfg.overlays [
+    (lib.concatMap (overlay: overlay.filters))
+    (lib.filter (filter: !validFilter filter))
+  ];
+  invalidParams = lib.pipe cfg.overlays [
+    (lib.concatMap (overlay: overlay.params))
+    (lib.filter hasLineBreak)
+  ];
```
quentinmit: Was there a particular reason you added this validation for overlay names, parameters, etc.? I'm generally of the opinion that we should just trust the user and if they want to do something wacky like put a newline or bracket in their filter they probably know what they are doing.

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +98,66 @@ let
...
+
+  # Raspberry Pi firmware rejects config.txt lines longer than 98 characters.
+  longLines = lib.pipe generatedConfig [
+    (lib.splitString "\n")
+    (lib.filter (line: builtins.stringLength line > 98))
+  ];
+
+  # A custom file bypasses generatedConfig. It cannot be combined with ordered
+  # overlays, and its lines are not validated here.
+  usesGeneratedFile = toString cfg.file == toString generatedFile;
```
quentinmit: A pattern I've seen used in nixpkgs is:
```nix
options.hardware.raspberry-pi.configtxt.file.highestPrio < (lib.mkOptionDefault { }).priority
```
and then you don't have to do awkward string comparisons on store paths. (`opt` is often set to `options.hardware.rasberry-pi.configtxt` ala how `cfg` is set)

raspberry-pi/common/config-txt.nix
```diff
@@ -123,14 +204,87 @@ in
...
+          }
+        ]
+      '';
+    };
+
     file = lib.mkOption {
       type = lib.types.path;
-      default = pkgs.writeText "config.txt" (toConfigTxt cfg.settings);
-      defaultText = lib.literalExpression ''pkgs.writeText "config.txt" (generated from settings)'';
+      default = generatedFile;
```
quentinmit: If you change `usesGeneratedFile` as I suggest above, this can go back to a literal `pkgs.writeText` call (with `generatedConfig` as the argument).

raspberry-pi/README.md
doronbehar: @quentinmit thanks for your detailed reply.

> I don't think it is reasonable to assume that a Nix configuration will be deployed on exactly one single machine with a particular set of unchanging hardware. For example, an SD card image might need to be booted on multiple generations of Pi. (This isn't just hypothetical - I personally have a uConsole in which I swap out a CM4 and CM5.)

Very cool. It might be nice to describe something similar in our `configtxt` README or something to justify this design detail of the module.

> > Regarding keying by overlay name, an attrset can't express the same overlay loaded twice. From the overlays README[3](#user-content-fn-3-ea7f946c66aeb8c8af8ee548fd2ac183):
> > > When you have additional hardware with more than one device of a kind, you end up using the same overlay multiple times with other parameters, e.g.
> > > ```
> > > dtoverlay=mcp251xfd,spi0-0,interrupt=25
> > > dtoverlay=mcp251xfd,spi0-1,interrupt=24
> > > ```
> > 
> > An attrset keyed on `mcp251xfd` would only get one of them.
> > There's a smaller problem as well. Nix sorts attribute names, so an attrset also gives up control of output order:
> 
> I agree that we need to support duplicate overlays. I believe it's also possible for one overlay to depend on another, which means that ordering matters between overlays with different names. I certainly dislike this because it prevents merging of overlays and parameters across modules. (You can't have one module enable the `audremap` overlay and then another module set the output pin, for instance, because you can't modify the list entry added by the first module.) But I don't really see a way around that.

I kind of lost you here. What most bothers me, is whether order matters in the above `dtoverlay=mcp251xfd,spi0-*` example. I don't like the idea that we consider how to do flipflops in the air just because we _think_, that for a very advanced usage of our module, order matters.

And @quentinmit, I wanted to get your opinion on the idea that a user wishing to set this would have to use the attribute set key `mcp251xfd,spi0-0,interrupt=25` or `mcp251xfd,spi0-1,interrupt=24`.

> Instead of a single list with `name`, `filters`, and `params` keys, I think we should match the same conditional filter syntax as used by settings. How about using the overlay name as a key and the parameters as the value ala `lib.types.attrTag`? For example,
> 
> ```nix
> hardware.raspberry-pi.configtxt = {
>     settings.pi4."gpio4=1".audio_pwm_mode = 1;
>     deviceTreeOverlays.pi4."gpio4=1" = [
>       { audremap.enable_jack = true; }
>       { ghostamp = {}; }
>     ];
> };
> ```

That's interesting, but could you please tell us how would you expect this Nix example to get rendered?

quentinmit: > I kind of lost you here. What most bothers me, is whether order matters in the above `dtoverlay=mcp251xfd,spi0-*` example. I don't like the idea that we consider how to do flipflops in the air just because we _think_, that for a very advanced usage of our module, order matters.

I think in the example above order does not matter. That was an example of instantiating the same overlay multiple times.

I have a concrete example of ordering on my uConsole. [This base overlay](https://github.com/robertjakub/oom-hardware/blob/main/raspberry-pi/overlays/source/cpi-pmu.dts) defines the `axp22x` charge controller and the `battery` object, and then [this overlay](https://github.com/quentinmit/isz/blob/main/uconsole/configuration.nix#L56-L79) sets the expected voltage and capacity of the battery. The second overlay must be applied after the first overlay.

> And @quentinmit, I wanted to get your opinion on the idea that a user wishing to set this would have to use the attribute set key `mcp251xfd,spi0-0,interrupt=25` or `mcp251xfd,spi0-1,interrupt=24`.

I think there are two problems with that.
- config.txt lines are limited to 98 characters, so all your parameters + overlay name + `dtoverlay=` would need to fit in 98 characters.
- We would lose the serialization of parameter values, so it would be on the user to figure out how to encode booleans/strings/etc.

> > Instead of a single list with `name`, `filters`, and `params` keys, I think we should match the same conditional filter syntax as used by settings. How about using the overlay name as a key and the parameters as the value ala `lib.types.attrTag`? For example,
> > ```nix
> > hardware.raspberry-pi.configtxt = {
> >     settings.pi4."gpio4=1".audio_pwm_mode = 1;
> >     deviceTreeOverlays.pi4."gpio4=1" = [
> >       { audremap.enable_jack = true; }
> >       { ghostamp = {}; }
> >     ];
> > };
> > ```
> 
> That's interesting, but could you please tell us how would you expect this Nix example to get rendered?

I would expect this to render as
```ini
[all]
[pi4]
[gpio4=1]
audio_pwm_mode=1

[all]
[pi4]
[gpio4=1]
dtoverlay=audremap
dtparam=enable_jack=on
dtoverlay=
dtoverlay=ghostamp
dtoverlay=
```

Obviously, a more clever serializer could remove some of the redundancy - this config would be equivalent to:

```ini
[pi4]
[gpio4=1]
audio_pwm_mode=1
dtoverlay=audremap
dtparam=enable_jack
dtoverlay=ghostamp
```

IMO it's not worth making this Nix module more complicated to eliminate this redundancy, though.

doronbehar: OK Not bad. I have a small improvement to your suggestion: Require the filters to come after `configtxt`, so the hierarchy is first and foremost based upon filters:

```nix
hardware.raspberry-pi.configtxt.pi4."gpio4=1" = {
    settings.audio_pwm_mode = 1;
    deviceTreeOverlays = [
      { audremap.enable_jack = true; }
      { ghostamp = {}; }
    ];
};
```

Rendered to:

```ini
[pi4]
[gpio4=1]
audio_pwm_mode=1
dtoverlay=audremap
dtparam=enable_jack
dtoverlay=ghostamp
```

What do you think? @JamieMagee I'm also interested in your input too.

JamieMagee: Apologies for the radio silence, I've been busy with work.

A couple of things put me off about hoisting the filters.

RFC 42 keeps the freeform part inside `settings`, with declared options as siblings ([the example in the RFC][1] has `freeformType` inside the `settings` submodule and `enable` next to it). Hoisting inverts that. `configtxt.pi4` would sit next to `configtxt.file`, so `configtxt` needs its own `freeformType` and user-chosen filter names end up sharing a namespace with real options.

It also stops `settings` mapping to config.txt. Right now `settings` is the file. Under your shape it's a fragment of one section. That was @quentinmit's reasoning for putting `deviceTreeOverlays` beside `settings` rather than inside it, and I think it applies to filters too.

So I'd like to go with @quentinmit's shape: `deviceTreeOverlays` alongside `settings`, filters by nesting, an ordered list at the leaf with one overlay name per element. Filters stay exactly where they are in `settings`, and the renderer keeps the `[all]` resets and terminators.

I'll pick up the three other threads at the same time (dropping the validation, `highestPrio` instead of the store path comparison, and documenting the bare `dtoverlay=`), and add a README note on the swappable-board case, since that's the justification for keeping filters at all.

Sound okay?

[1]: https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-1-1

raspberry-pi/common/config-txt.nix
JamieMagee: No good reason, and you're right that there isn't one. Dropped along with the line-length assertion below it.

Worth noting for anyone reading later: the RPi docs give two different limits. The [file format section](https://www.raspberrypi.com/documentation/computers/config_txt.html#file-format) says 98 characters, while the [`dtparam` section](https://www.raspberrypi.com/documentation/computers/config_txt.html#dtparam) says 80.

raspberry-pi/common/config-txt.nix
JamieMagee: Nice, thanks. I hadn't come across `highestPrio` before.

It turns out I don't need it any more. `usesGeneratedFile` only existed to gate the two assertions (overlays combined with a custom file, and the line-length check), and both are gone now.

raspberry-pi/common/config-txt.nix
JamieMagee: Done.

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +81,78 @@ let
...
+
+  # Two paths can reduce to the same filter set (all.pi4 and pi4.all both yield
+  # [ "pi4" ]), so merge their entries first. The sort matches the one used for
+  # settings groups, which keeps the output stable.
+  overlayGroups = lib.pipe cfg.deviceTreeOverlays [
+    (recurseOverlays [ ])
+    (builtins.groupBy (group: builtins.toJSON group.conditionals))
+    (lib.mapAttrsToList (
+      groupJSON: groups: {
+        conditionals = builtins.fromJSON groupJSON;
```
doronbehar: I'm not sure exactly how this works, but there's no way I'll be approving a Nix function that serializes to json and then from json. I'm pretty sure this is just noot needed.

raspberry-pi/common/config-txt.nix
```diff
@@ -123,13 +199,83 @@ in
       '';
     };
 
+    deviceTreeOverlays = lib.mkOption {
+      type =
+        with lib.types;
```
doronbehar: Please avoid `with lib.types;` See https://nix.dev/guides/best-practices.html#with-scopes .

raspberry-pi/common/config-txt.nix
```diff
@@ -123,13 +199,83 @@ in
...
+        let
+          params = attrsOf (
+            nullOr (oneOf [
+              str
+              int
+              bool
+            ])
+          );
+          entry = addCheck (attrsOf params) (e: builtins.length (builtins.attrNames e) == 1) // {
```
doronbehar, line 273: Could you explain this length check? Maybe we should simply rely upon the leaf value being of type list?

raspberry-pi/common/config-txt.nix
doronbehar: Please avoid `with lib.types` here too please, if you can (sorry I missed this in the previous PR...).

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +81,78 @@ let
...
+    );
+
+  # Settings render first, so base DT parameters apply before any overlay loads.
+  # The overlay groups follow.
+  settingsConfig = toConfigTxt cfg.settings;
+  overlayConfig = lib.concatMapStringsSep "\n\n" renderOverlayGroup overlayGroups;
+  generatedConfig =
+    settingsConfig
+    + lib.optionalString (overlayGroups != [ ]) (
+      lib.optionalString (settingsConfig != "") "\n" + overlayConfig + "\n"
```
doronbehar: Why do you use a conditional inside a conditional here? And why do we need to add `overlayConfig` only if `settingsConfig != ""`? In general, why do we need the conditionals? The `overlayConfig` string will simply become an empty harmless string I think.

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +81,78 @@ let
...
+    ))
+    (lib.sort (a: b: a.conditionals < b.conditionals))
+  ];
+
+  # A bare `dtoverlay=` closes the parameter scope of the current overlay, so a
+  # later dtparam applies to the base DTB instead. It must never be the first
+  # dtoverlay or dtparam line in the file, where it means "load no HAT overlay".
+  # Settings render first, and every terminator here follows a dtoverlay=<name>,
+  # so that cannot happen now. A change to the section order can reintroduce it.
+  renderEntry =
```
doronbehar: Could you simply avoid defining this function and clutter the namespace, and simply write it inline below? This wouldn't have bothered me that much unless this function name was so unspecific :).

doronbehar (changes requested):

raspberry-pi/common/config-txt.nix
quentinmit: This is to ensure that the input is:
```nix
[
  { overlay1 = {}; }
  { overlay2 = {}; }
]
```
and not
```nix
[
  { overlay1 = {}; overlay2 = {}; }
]
```
(i.e. the list elements are `types.attrTag`-style single-key attrsets)

raspberry-pi/common/config-txt.nix
quentinmit: Unfortunately, the `toJSON` is needed. `builtins.groupBy` requires a string grouping key, so the list of filters needs to be turned into a string to use as the grouping key. I guess the `fromJSON` call could be replaced with `inherit (builtins.elemAt groups 0) conditionals;`, if you think that's slightly cleaner.

raspberry-pi/common/config-txt.nix
```diff
@@ -77,6 +81,78 @@ let
...
     lib.concatMapStringsSep "\n[all]\n" mkGroup groups;
 
+  # Overlay parameters use on/off for booleans. This follows the Raspberry Pi
+  # documentation and the overlay README. Plain settings keep 1/0.
+  mkParamString = v: if builtins.isBool v then (if v then "on" else "off") else mkValueString v;
+
+  # Flatten deviceTreeOverlays into { conditionals, entries } records. Unlike
+  # settings, there is no leaf key. Every level of nesting is a conditional
+  # filter, and the leaf is the ordered overlay list for that filter set.
+  recurseOverlays =
```
quentinmit, line 83: I have a hilarious (terrible?) idea. This module serializes string values verbatim, right? What if we just turn

```nix
deviceTreeOverlays.pi4 = [ { overlay1 = { param="value"; }; } ];
```
into
```nix
settings.pi4.dtoverlay = [ "overlay1\ndtparam=param=value\ndtoverlay=" ];
```
?

We don't do any quoting of setting values, so this would just render as
```ini
dtoverlay=overlay1
dtparam=value
dtoverlay=
```
and this would interleave correctly with any user-specified `dtoverlay` or `dtparam` values. Then we wouldn't need any changes or additions to the serialization code.

JamieMagee: I like it more than I expected to, and it does work for the defaults. Lowering `vc4-kms-v3d` that way renders:

```ini
arm_boost=1
dtoverlay=vc4-kms-v3d
dtoverlay=
dtparam=audio=on
```

What stops me is the case where a user adds an ordinary overlay of their own:

```nix
settings.all = {
  dtparam = [ "audio=on" ];
  dtoverlay = [ "vc4-kms-v3d\ndtoverlay=" "disable-bt" ];   # second one written by the user
};
```

```ini
dtoverlay=vc4-kms-v3d
dtoverlay=
dtoverlay=disable-bt
dtparam=audio=on
```

`audio=on` is now inside the scope of `disable-bt`. Any raw `dtoverlay` that lands last captures every `dtparam` after it, and where the names collide the overlay silently wins. That is the same failure this PR started from.

So I would rather keep the separate renderer.

raspberry-pi/common/config-txt.nix
JamieMagee: Half of it can go. As @quentinmit says, `builtins.groupBy` only takes a string key, so the list of filters has to be serialised to group by it. The `fromJSON` on the way back out was avoidable though: every member of a group has the same `conditionals` value, so the first one gives it back.

```nix
inherit (builtins.elemAt groups 0) conditionals;
```

Output is identical. I have applied it to the `settings` renderer as well, where the same pattern came in with #1788, so `fromJSON` is now gone from the file entirely. That one also had an inner lambda argument shadowing the binding it was being assigned to, which I renamed while I was in there.

raspberry-pi/common/config-txt.nix
JamieMagee: Both replaced with `inherit (lib.types) ...` in the existing `let`.

raspberry-pi/common/config-txt.nix
JamieMagee: I put it in a comment next to the `addCheck`.

raspberry-pi/common/config-txt.nix
JamieMagee: Yeah, this is a little overengineered 😅 Dropping this guard only ever adds blank lines to the output, which the firmware ignores anyway. It's now just a filter over non-empty sections.

raspberry-pi/common/config-txt.nix
JamieMagee: Fair. Inlined into the `lib.concatMap` in `renderOverlayGroup`.

raspberry-pi/common/config-txt.nix
quentinmit: I don't understand your example. The user would have that problem with or without the rendered overlay, no? i.e.

```nix
settings.all = {
  dtparam = [ "audio=on" ];
  dtoverlay = [ "disable-bt" ];
};
```
will already produce the broken
```ini
dtoverlay=disable-bt
dtparam=audio=on
```
Adding the `"vc4-kms-v3d\ndtoverlay="` overlay doesn't change the behavior - either way that parameter is (mis?)interpreted as a parameter for the overlay instead of a parameter for the base devicetree.

doronbehar (commented):
@JamieMagee while trying to explain what I still don't like about the code, I eventually wrote the following code. Please tell me what do you think:

```nix
let
  cfg = config.hardware.raspberry-pi.configtxt;

  mkValueString =
    v:
    if builtins.isInt v then
      toString v
    else if builtins.isString v then
      v
    else if true == v then
      "1"
    else if false == v then
      "0"
    else
      abort "config.txt: unsupported value type: ${builtins.typeOf v}";

  mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";

  # Overlay parameters use on/off for booleans. This follows the Raspberry Pi
  # documentation and the overlay README. Plain settings keep 1/0.
  mkParamValue = v: if builtins.isBool v then (if v then "on" else "off") else mkValueString v;

  # "all" filters are omitted since they reset state, and the remaining
  # filters are sorted so that stacking order doesn't matter for grouping.
  mkConditionals =
    path:
    lib.pipe path [
      (lib.filter (k: k != "all"))
      (lib.sort builtins.lessThan)
    ];

  # Recursively flatten nested attrs into a list of { conditionals, items } groups,
  # one group per leaf. Each nesting level (except the leaf) is a conditional
  # filter. null values are filtered out, allowing `mkForce null` to remove
  # defaults.
  recurseSettings =
    path: value:
    if value == null then
      [ ]
    else if !(builtins.isAttrs value) then
      [
        {
          conditionals = mkConditionals (lib.init path);
          items = [
            {
              # `path` is appended per call, so the last element is the leaf's own key.
              name = lib.last path;
              inherit value;
            }
          ];
        }
      ]
    else
      lib.pipe value [
        (lib.mapAttrsToList (name: recurseSettings (path ++ [ name ])))
        lib.concatLists
      ];

  # Flatten deviceTreeOverlays into { conditionals, items } groups, one group
  # per leaf. Unlike settings, the leaf itself is already the ordered overlay
  # list for that filter set, so every level of nesting is a conditional filter.
  recurseOverlays =
    path: value:
    if value == null then
      [ ]
    else if builtins.isList value then
      [
        {
          conditionals = mkConditionals path;
          items = value;
        }
      ]
    else
      lib.pipe value [
        (lib.mapAttrsToList (name: recurseOverlays (path ++ [ name ])))
        lib.concatLists
      ];

  # Merge groups whose `conditionals` lists are equal by concatenating their
  # `items`, then sort the merged groups for stable output. Two different
  # attrs paths can reduce to the same filter set (all.pi4 and pi4.all both
  # yield [ "pi4" ]), so merging is necessary even though `recurseSettings`/
  # `recurseOverlays` already return one group per leaf.
  #
  # The final sort is not redundant with `mkConditionals`: that only sorts the
  # filters *within* one group's `conditionals`, not the groups relative to
  # each other. Groups arrive in attrs tree-walk order, which is driven by the
  # raw, pre-filter attribute names - so `all.z.foo` walks before `b.bar` since
  # "all" < "b", even though "all" is later stripped and the resulting
  # `conditionals` should sort as [ "b" ] < [ "z" ]. Sorting here, after
  # merging, is what fixes that up.
  groupByConditionals =
    groups:
    lib.pipe groups [
      (lib.foldl' (
        merged: group:
        let
          idx = lib.lists.findFirstIndex (g: g.conditionals == group.conditionals) null merged;
        in
        if idx == null then
          merged ++ [ group ]
        else
          lib.imap0 (i: g: if i == idx then g // { items = g.items ++ group.items; } else g) merged
      ) [ ])
      (lib.sort (a: b: a.conditionals < b.conditionals))
    ];

  # Every group opens with [all] plus its own filters. For settings this is
  # only ever a no-op reset. For overlays it's required: a bare `dtoverlay=`
  # ends overlay scope but leaves the conditional filters set, and only [all]
  # resets them.
  mkFilterHeader = conditionals: lib.concatMapStrings (f: "[${f}]\n") ([ "all" ] ++ conditionals);

  # Settings render first, so base DT parameters apply before any overlay
  # loads. The overlay groups follow.
  generatedConfig =
    lib.pipe cfg.settings [
      (recurseSettings [ ])
      groupByConditionals
      (map (
        group:
        mkFilterHeader group.conditionals
        + lib.pipe group.items [
          builtins.listToAttrs
          (lib.generators.toKeyValue {
            inherit mkKeyValue;
            listsAsDuplicateKeys = true;
          })
        ]
      ))
      (lib.concatMapStrings (group: group + "\n"))
    ]
    + lib.pipe cfg.deviceTreeOverlays [
      (recurseOverlays [ ])
      groupByConditionals
      # Each entry ends with a bare `dtoverlay=`. This terminator closes the
      # parameter scope of that overlay, so a later dtparam applies to the
      # base DTB instead. A terminator must never come first in the file,
      # where it means "load no HAT overlay" - settings render first, and
      # every terminator here follows a dtoverlay=<name>, so no terminator
      # comes first today. A change to the section order can break this
      # guarantee.
      (map (
        group:
        mkFilterHeader group.conditionals
        + lib.concatMapStringsSep "\n" (
          entry:
          let
            name = lib.head (lib.attrNames entry);
          in
          lib.concatStringsSep "\n" (
            [ "dtoverlay=${name}" ]
            ++ lib.pipe entry.${name} [
              (lib.filterAttrs (_: v: v != null))
              (lib.mapAttrsToList (p: v: "dtparam=${p}=${mkParamValue v}"))
            ]
            ++ [ "dtoverlay=" ]
          )
        ) group.items
      ))
      (lib.concatMapStrings (group: group + "\n\n"))
    ];

in
```

I think it is much more readable and maintainable, because:

- No functions are defined and used once - every function declared is declared in order to be used for both `cfg.settings` and `cfg.deviceTreeOverlays`.
- `cfg.deviceTreeOverlays` and `cfg.settings` are handled similarly and equivalently. A single grouping function is used for both, and each attribute set is recursed with a similar function with a very similar structure and name.

Disclaimer: I used Claude-code to write this. It is pretty thoroughly tested by it. Tell me if you think it'd be nice to write evaluation tests for this code.

raspberry-pi/common/config-txt.nix
JamieMagee: Ah yeah, you're right. On the branch:

```nix
settings.all = {
  dtparam = [ "audio=on" ];
  dtoverlay = [ "disable-bt" ];
};
```

```ini
dtoverlay=disable-bt
dtparam=audio=on
```

Already broken, no lowering involved. I had conflated "a raw `settings.dtoverlay` entry is unsafe", which predates this PR, with "lowering would introduce that", which it would not.

Entries the module lowers carry their own terminator and are safe wherever they land. Handwritten ones do not, and the module cannot add one without parsing the string. So lowering would not create the problem, it would put safe and unsafe entries in one list. That is an argument for discouraging raw `settings.dtoverlay`, not an argument against your idea.

I kept the separate renderer for now, mostly because @doronbehar's refactor landed on top of it and shrank the code the lowering would have deleted. But if you still think it is worth doing, say so and I can reopen it.

JamieMagee:
@doronbehar thanks for that. I took your suggestions, with one change.

Double checking your version against the current branch and a bunch of different tests, everything was functionally equivalent. The only two changes were:

- a leading `[all]` on the first settings group
- a trailing blank line, from `group + "\n\n"`

I kept the `[all]` and dropped the blank line.

The one change I made is `groupByConditionals`. I rewrote it as:

```nix
groupByConditionals =
  groups:
  lib.pipe groups [
    (builtins.groupBy (group: builtins.toJSON group.conditionals))
    (lib.mapAttrsToList (
      _: matching: {
        inherit (builtins.elemAt matching 0) conditionals;
        items = lib.concatMap (group: group.items) matching;
      }
    ))
    (lib.sort (a: b: a.conditionals < b.conditionals))
  ];
```

On tests: yes, done, in `tests/config-txt-tests.nix`, wired into `checks`. Run with `nix build .#checks.x86_64-linux.raspberry-pi-config-txt`.

doronbehar:
> The one change I made is `groupByConditionals`. I rewrote it as:

Why bring back `groupyBy` + `toJSON` :sob: ?

> On tests: yes, done, in `tests/config-txt-tests.nix`, wired into `checks`. Run with `nix build .#checks.x86_64-linux.raspberry-pi-config-txt`.

Nice, thanks. Could you convert these to an attribute set please? map each attribute of test to an attribute. Running `nix build .#checks.x86_64-linux.raspberry-pi-config-txt` will still build all of them, and it will still be possible to build individual tests.

JamieMagee:
> Why bring back `groupBy` + `toJSON` :sob: ?

Okay, I think we can do it without `toJSON` and still be linear. How's this?

```nix
groupByConditionals =
  groups:
  lib.pipe groups [
    (lib.sort (a: b: a.conditionals < b.conditionals))
    (lib.foldl' (
      merged: group:
      let
        previous = lib.head merged;
      in
      if merged != [ ] && previous.conditionals == group.conditionals then
        [ (previous // { items = previous.items ++ group.items; }) ] ++ lib.tail merged
      else
        [ group ] ++ merged
    ) [ ])
    lib.reverseList
  ];
```

> Could you convert these to an attribute set please?

Done.

```sh
nix build .#checks.x86_64-linux.raspberry-pi-config-txt                # all
nix build .#checks.x86_64-linux.raspberry-pi-config-txt.overlayOrder   # one
```

raspberry-pi/common/config-txt.nix
```diff
@@ -35,55 +39,165 @@ let
...
+        ++ lib.pipe entry.${name} [
+          (lib.filterAttrs (_: v: v != null))
+          (lib.mapAttrsToList (p: v: "dtparam=${p}=${mkParamValue v}"))
+        ]
+        ++ [ "dtoverlay=" ]
+      )
+    ) group.items;
+
+  renderSection =
+    recurse: render: value:
```
doronbehar: I find the following naming much more readable:

```suggestion
  render =
    config: renderSectionFunc: recurseFunc:
```

raspberry-pi/common/config-txt.nix
```diff
@@ -35,55 +39,165 @@ let
...
+    lib.pipe value [
+      (recurse [ ])
+      groupByConditionals
+      (map render)
+      (lib.concatStringsSep "\n")
+    ];
+
+  # Settings render first, so base DT parameters apply before any overlay loads.
+  # The overlay groups follow. A blank line separates the two sections.
+  sections = lib.filter (section: section != "") [
```
doronbehar: Again, why filter empty sections? What is the amount of extra unnecessary new lines we will experience? If it is less then 3, I think it'd be better to simplify things by writing `generatedConfig` straight away.

raspberry-pi/common/config-txt.nix
```diff
@@ -35,55 +39,165 @@ let
...
+        merged: group:
+        let
+          previous = lib.head merged;
+        in
+        if merged != [ ] && previous.conditionals == group.conditionals then
+          [ (previous // { items = previous.items ++ group.items; }) ] ++ lib.tail merged
+        else
+          [ group ] ++ merged
+      ) [ ])
+      lib.reverseList
```
doronbehar: Is this really needed?

doronbehar:
> Okay, I think we can do it without `toJSON` and still be linear.

So this whole time you insisted upon `toJSON` + `builtins.groupBy` to still be linear and not what? If the answer is yes, you could have been more communicative about it.

JamieMagee:
Yeah, that was the reason, apologies for not saying it earlier.

I said `toJSON` was needed, but what was I should have said it's needed *for `builtins.groupBy`*, and the reason I
wanted `groupBy` was to avoid rescanning the accumulator on every merge. I never said that second part. Then when you sent your version I swapped it back without giving the reason either, so from where you're sitting it's just been me undoing your work twice with no reason attached. Sorry, that's on me.

raspberry-pi/common/config-txt.nix
JamieMagee: Done, but with one change: I used `value` rather than `config` for the first argument. `config` is already the module argument that `cfg` comes from, so shadowing it inside the module felt like a footgun for whoever comes next.

raspberry-pi/common/config-txt.nix
JamieMagee: Good catch! No it isn't 😅 It was only there because `foldl'` builds the accumulator backwards. Folding from the right instead means each step prepends, so the result comes out in order and the reverse goes away.

```nix
(lib.foldr (
  group: merged:
  let
    next = lib.head merged;
  in
  if merged != [ ] && next.conditionals == group.conditionals then
    [ (group // { items = group.items ++ next.items; }) ] ++ lib.tail merged
  else
    [ group ] ++ merged
) [ ])
```

raspberry-pi/common/config-txt.nix
JamieMagee: I ran the tests, and it only saves one extra newline, but twelve of the thirteen tests change. Every case that exercises only settings or only overlays picks up an extra leading or trailing blank line, and only the case with both populated stays the same. The newline itself is harmless in the file. But I'm worried that it looks like a typo, which someone will try tidy up and break.

So I've dropped the `sections` binding (which I think was the ceremony you were pointing at) and inlined the filter:

```nix
generatedConfig = lib.concatStringsSep "\n" (
  lib.filter (section: section != "") [
    (render cfg.settings renderSettingsGroup recurseSettings)
    (render cfg.deviceTreeOverlays renderOverlayGroup recurseOverlays)
  ]
);
```

Happy to drop the filter as well and take the blank lines if you'd still rather.

tests/config-txt-tests.nix
```diff
@@ -0,0 +1,245 @@
...
+      touch $out
+    ''
+  ) cases;
+in
+# Building this builds every case. Individual cases are attributes on it, so
+# `nix build .#checks.<system>.raspberry-pi-config-txt.overlayOrder` works too.
+runCommand "raspberry-pi-config-txt-tests" { passthru = tests; } ''
+  ${lib.concatMapStringsSep "\n" (test: "cat ${test} > /dev/null") (lib.attrValues tests)}
+  touch $out
+''
```
doronbehar: What do you need this `runCommand` for? And what does `cat ${test} > /dev/null` is supposed to mean?? Why not simply:


```suggestion
  # One derivation per case, so a single test can be built on its own.
  lib.mapAttrs (
    name: case:
    runCommand "config-txt-${name}" { } ''
      diff -u ${writeText "${name}-expected.txt" case.expected} ${render case.module}
      touch $out
    ''
  ) cases;
```

raspberry-pi/common/config-txt.nix
```diff
@@ -35,55 +39,167 @@ let
...
+
+  # Settings render first, so base DT parameters apply before any overlay loads.
+  # The overlay groups follow. A blank line separates the two sections, and the
+  # filter keeps that separator from becoming a stray leading or trailing line
+  # when one of the two is empty.
+  generatedConfig = lib.concatStringsSep "\n" (
+    lib.filter (section: section != "") [
+      (render cfg.settings renderSettingsGroup recurseSettings)
+      (render cfg.deviceTreeOverlays renderOverlayGroup recurseOverlays)
+    ]
```
doronbehar: More readable IMO:

```suggestion
  generatedConfig = lib.pipe [
    (render cfg.settings renderSettingsGroup recurseSettings)
    (render cfg.deviceTreeOverlays renderOverlayGroup recurseOverlays)  
  ] [
    (lib.filter (section: section != ""))
    (lib.concatStringsSep "\n")
  ];
```

raspberry-pi/common/config-txt.nix
```diff
@@ -35,55 +39,167 @@ let
...
+          (lib.filterAttrs (_: v: v != null))
+          (lib.mapAttrsToList (p: v: "dtparam=${p}=${mkParamValue v}"))
+        ]
+        ++ [ "dtoverlay=" ]
+      )
+    ) group.items;
+
+  render =
+    value: renderSectionFunc: recurseFunc:
+    lib.pipe value [
```
doronbehar: I understand what you wrote earlier about the name `config` being taken, but `value` is definitely a bad name. We are talking about a complex, full configuration attribute set, it is not a single `value`. How about `conf` or `confAttrSet`?

raspberry-pi/README.md
```diff
@@ -49,27 +49,73 @@ Board profiles import `hardware.raspberry-pi.configtxt`, which renders `config.t
...
 
 Top-level attrs are conditional sections (`all`, `pi4`, `pi5`, `cm4`, and so on). Nesting stacks filters. To drop a default, set the key to `null` with `mkForce`.
 
+Every rendered group opens with `[all]` and then its own filters. Filters of different types stack rather than replace, so `[all]` is what clears the previous group before the next one starts.
+
+The firmware applies these filters at boot, not Nix at evaluation time. This matters in two cases. One image can boot on more than one board, such as a CM4 and a CM5 swapped into the same carrier. Filters like `[EDID=...]`, `[gpio4=1]`, `[tryboot]` and `[bootvar0=42]` also depend on state that Nix cannot see: the attached monitor, a jumper, an EEPROM value.
+
+### Device tree overlays
+
+Overlays go in `configtxt.deviceTreeOverlays`, not in a `dtoverlay` key under `settings`. Filters nest the same way, and the leaf is an ordered list where each element names one overlay and its parameters:
```
doronbehar, line 79: I think we discussed this, but I'm not sure what was the resolution: Do we ban `dtoverlay` from `settings`? Maybe we should? And instruct the user to use `deviceTreeOverlays`?

doronbehar (commented):
Looking better :).

raspberry-pi/README.md
JamieMagee: I think we should, but not in this PR. I'd like to migrate all the existing overlays under `raspberry-pi/4` first, so we have a lot of good examples of how to use `deviceTreeOverlays`, then add the assertion banning `dtoverlay`.

raspberry-pi/common/config-txt.nix
JamieMagee: Yeah, sounds good. Renamed to `conf`.

raspberry-pi/common/config-txt.nix
JamieMagee: Agreeded, done.

tests/config-txt-tests.nix
JamieMagee: Yeah, the `cat` was ugly. Removed. The wrapper needs to stay though, or `nix build` on the parent stops working:

```
error: expected flake output attribute 'checks.x86_64-linux.raspberry-pi-config-txt'
to be a derivation or path but found a set
```

A flake check has to be a derivation, and an attrset of them doesn't work. Naming each test case in the derivation environment does the same job without being too confusing:

```nix
runCommand "raspberry-pi-config-txt-tests" {
  cases = lib.attrValues tests;
  passthru = tests;
} "touch $out"
```

doronbehar (approved):
Thanks for all the work and cooperation @JamieMagee !

doronbehar:
@JamieMagee I split the changes here a bit to semantic changes that should be applied independently of the main change of the PR. I'm sure you'd be OK with it.

doronbehar:
@quentinmit it'd be nice to get your approval of the PR before merging.

JamieMagee:
@doronbehar looks great, thank you!

doronbehar:
@quentinmit I'd like to merge this in 12-24 hours if you have no more comments to give.

PR #1948, opened 2026-07-15
Add Keylightd derivation and module, make available to Framework 13 and Framework 16

twoolie:
###### Description of changes

Adds a simple wrapper to build [keylightd](https://github.com/jonas-schievink/keylightd) and provide it as a systemd service. Has options to set the brightness and timeout, and control the power light.

This is a huge help when watching videos in dark rooms when the non-dimmable power light can be quite distracting.

Module is only imported for Framework 13 and 16 as the Framework 12 does not have a keyboard backlight.

###### Things done

### DIFF
--- a/framework/13-inch/common/default.nix
+++ b/framework/13-inch/common/default.nix
@@ -5,6 +5,7 @@
     ../../../common/pc/ssd
     ../../kmod.nix
     ../../framework-tool.nix
+    ../../keylightd.nix
     ./classic-audio.nix
   ];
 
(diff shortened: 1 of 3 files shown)


dantefromhell:
There's also https://gitlab.com/mamarley/keylightc which seems to be maintained more regularly, but it does not support changing the light status of the power button from what I can see.

JohnAZoidberg:
This looks like it should go into nixpkgs not nixos-hardware. It's a package and a service that a user can enable/install if they like.

twoolie:
I'm not against upstreaming the package,  as framework-tool is already in nixpkgs.

I would contend that the module that exposes "hardware.framework.keylightd" should live here with the rest of the "hardware.framework.*" options, given that it only works with this particular framework hardware.

PR #1952, opened 2026-07-17
macbookpro-12-1 replace power up/down commands with systemd oneshot services

### LINKED ISSUE #1794: macbook-12-1:  powerManagement.powerUpCommands is going to be deprecated in nixos 26.11
```
evaluation warning: powerManagement.powerUpCommands is deprecated due to it having unclear ordering semantics.
                    It will be removed in NixOS 26.11.
                    It is recommended to create an explicit systemd oneshot service instead,
                    that is pulled in at the right time during the boot process.
                    See https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html
                    for more information on possible targets that can be used for this.

                    If you also want to run this service upon waking up from resume, the recommended
                    method to do so is described here:
                    https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#sleep.target
```

DinoLeung:
###### Description of changes

- updated `powerManagement.powerUpCommands` and `powerManagement.powerDownCommands` to systemd oneshort services (fixes #1794)
- updated the `powerManagement.powerUpCommands` in readme as well

###### Things done

### DIFF
--- a/apple/macbook-pro/12-1/README.md
+++ b/apple/macbook-pro/12-1/README.md
@@ -3,12 +3,18 @@
 ## Wireless won't get reconnected after resume/hibernate
 
 The wifi driver is unloaded before suspend/hibernate to workaround driver issues.
-This means it might be required to restart your wifi deamon i.e. wpa_supplicant:
+This means it might be required to restart your wifi daemon i.e. wpa_supplicant:
 
 ```nix
-powerManagement.powerUpCommands = ''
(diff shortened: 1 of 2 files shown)


DinoLeung:
@kpy3 would you mind having a look?

kpy3:
@DinoLeung unfortunately I don't have this old MacBook anymore

PR #1963, opened 2026-07-24
raspberry-pi: add wireless firmware to Pi 3 and 5

JamieMagee:
###### Description of changes

Pi 4 already installs nixos-hardware's pinned wireless firmware package. Pi 3 and Pi 5 only received firmware when another module, such as the generic SD image, enabled the broader redistributable firmware set.

Add the same package to both profiles. Systems importing only a board profile now use the 2026-03-21 firmware instead of relying on nixpkgs' older 2025-04-08 package.

cc: @doronbehar 

###### Things done

### DIFF
--- a/raspberry-pi/3/default.nix
+++ b/raspberry-pi/3/default.nix
@@ -35,4 +35,5 @@
     "console=tty0"
   ];
 
+  hardware.firmware = [ (pkgs.callPackage ../common/raspberry-pi-wireless-firmware.nix { }) ];
 }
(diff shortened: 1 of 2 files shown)


doronbehar (commented):
What about raspberry pi 2? Maybe add this to `common.nix`? For reference, this is the commit that added firmware only for Raspberry Pi 4:

https://github.com/NixOS/nixos-hardware/commit/2a3a409ac57bbe31cc4e51a14799a34927066499

JamieMagee:
The Raspberry Pi 2 doesn't have WiFi or Bluetooth.

PR #1975, opened 2026-08-03
framework: Remove dhowett's driver

JohnAZoidberg:
###### Description of changes
Most of it (Battery Charge Control and LED control) has been upstream for a while. Only fan control and privacy switches hasn't.

The module hasn't been updated in two years.

###### Things done

### DIFF
--- a/framework/kmod.nix
+++ b/framework/kmod.nix
@@ -4,32 +4,16 @@
   ...
 }:
 {
-  options.hardware.framework.enableKmod =
-    (lib.mkEnableOption "the community-created Framework kernel module that allows interacting with the embedded controller from sysfs.")
-    // {
-      # enable by default on NixOS >= 24.05 and kernel >= 6.10
-      default = lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.10";
-      defaultText = "enabled by default if kernel >= 6.10";
-    };
-
-  config = lib.mkIf config.hardware.framework.enableKmod {
-    assertions = [
-      {
-        assertion = lib.versionAtLeast config.boot.kernelPackages.kernel.version "6.10";
-        message = "The framework laptop kernel module requires Linux 6.10 or above";
-      }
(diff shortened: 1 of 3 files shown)


JohnAZoidberg:
cc @dhowett

DHowett:
as the @dhowett in question: yes absolutely, upstream is much more capable now than it was in 2022 when I started work on framework-laptop-kmod.

It has no place in a modern distro :)

framework/kmod.nix
```diff
@@ -4,32 +4,16 @@
...
-
-      # https://github.com/DHowett/framework-laptop-kmod?tab=readme-ov-file#usage
-      kernelModules = [
-        "cros_ec"
-        "cros_ec_lpcs"
-      ];
-    };
+    # Make sure that the charge control driver is loaded
+    extraModProbeConfig = ''
+      options cros_charge_control probe-with_fwk_charge_control=1
```
JohnAZoidberg, line 16: @t-8ch this is still needed/recommended, right?

t-8ch: AFAIK the upstream interface used by this driver (and therefore the driver itself) is still completely broken on current Framework firmware versions.
And given that the command `EC_CMD_CHARGE_LIMIT_CONTROL 0x3E03` is still present in the latest firmware, the module parameter is still necessary.

But this parameter is meant to be set *by the end user*. As they will need to make sure not to use `EC_CMD_CHARGE_LIMIT_CONTROL` in any way, as that will also break the driver. Setting the parameter automatically behind their backs defeats its whole purpose.

t-8ch:
> Only fan control and privacy switches hasn't.

Fan control is on my todo list.

Privacy switches in its current form would be trivial to do.
But given that there is no update notification implemented,
I am not really sure what the point of it would be.
(Given that it is a Framework-specific interface, personally I don't plan on implementing this)

PR #1983, opened 2026-08-08
TUXEDO: load Motorcomm YT6801 only if kernel is below v7.0

rtucek:
###### Description of changes

TUXEDO used to recommend to install the Motorcomm YT6801 driver
manually, unless using one of their WebFAI-supported distributions.[^1]

However, starting with version v7.0[^2], the kernel has first-class
supported for the driver.

It's worth pointing out, that users may still manually apply
`linuxKernel.packages.*.yt6801` in order to load the external kernel
module, provided by the vendor directly.[^3]

[^1]: https://www.tuxedocomputers.com/en/Motorcomm-YT6801-txtmotorcomm-yt6801-paket.tuxedo
[^2]: https://github.com/torvalds/linux/commit/02ff155ea2812ea25a086665522fd4fa196ef0ce
[^3]: https://github.com/NixOS/nixpkgs/blob/nixos-26.05/pkgs/os-specific/linux/yt6801/default.nix

###### Things done

### DIFF
--- a/tuxedo/infinitybook/pro14/gen10/default.nix
+++ b/tuxedo/infinitybook/pro14/gen10/default.nix
@@ -4,8 +4,10 @@
     ../../.
   ];
 
-  # Add Motorcomm YT6801 Driver if available
+  # Add Motorcomm YT6801 Driver if available and kernel version is below 7.0
   boot.extraModulePackages =
     with config.boot;
-    lib.lists.optional (kernelPackages ? yt6801) kernelPackages.yt6801;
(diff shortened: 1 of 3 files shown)


rtucek:
I've e2e tested it on Linux v7.1.7 (latest) and v6.18.43 (LTS) with my TUXEDO InfinityBook Pro14 Gen9 AMD (see rtucek/dotfiles#12).

For good measure, I've kept Pro14 Gen 10 and Pro15 Gen 10 models in sync with the same gate.

PR #1993, opened 2026-08-18
raspberry-pi: add gpio-fan overlay

JamieMagee:
###### Description of changes

Adds `hardware.raspberry-pi.gpio-fan`, using the upstream [`gpio-fan` device tree overlay][1].
 
The module exposes the GPIO pin, temperature threshold, and hysteresis settings through `hardware.raspberry-pi.configtxt.deviceTreeOverlays`. This avoids the in-tree DTS proposed in #1826 and makes the option available to every Raspberry Pi profile.

cc: @matiasyocca @doronbehar 

###### Things done

[1]: https://github.com/raspberrypi/linux/blob/rpi-6.18.y/arch/arm/boot/dts/overlays/gpio-fan-overlay.dts

### DIFF
--- a/raspberry-pi/common/default.nix
+++ b/raspberry-pi/common/default.nix
@@ -3,6 +3,7 @@
     ./config-txt.nix
     ./config-txt-defaults.nix
     ./firmware.nix
+    ./gpio-fan.nix
   ];
 
   boot.initrd.availableKernelModules = [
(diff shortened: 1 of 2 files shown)


doronbehar:
I'm totally against this PR. It is too trivial - users who install a gpio fan on their Raspberry Pi should be competent enough to figure out that overlay you transcribed here. Also, @matiasyocca had enough time to reply if they are not satisfied with the general solution available with #1947 (and demonstrated here).

JamieMagee:
Fair enough. I opened this as a replacement for #1826 after #1947 landed, but I see your point that `deviceTreeOverlays` already covers the use case (I documented `gpio-fan` in the README 😅). Thanks for the review.

PR #1994, opened 2026-08-19
framework: Add Laptop 12 Intel Core Ultra Series 3

JohnAZoidberg:
###### Description of changes
Add support for https://frame.work/blog/framework-laptop-12-now-with-core-series-3

###### Things done

### DIFF
--- a/framework/12-inch/intel-ultra-s3/README.md
+++ b/framework/12-inch/intel-ultra-s3/README.md
@@ -1,0 +1,33 @@
+# [Framework Laptop 12](https://frame.work/laptop12)
+
+## Updating Firmware
+
+Everything is updateable through fwupd, so it's enabled by default.
+
+To get the latest firmware, run:
+
+```sh
+$ fwupdmgr refresh
+$ fwupdmgr update
+```
+
+- [Latest Update](https://example.com/TODO)
+
+## Touchscreen support in initrd (for unl0kr)
+
+To unlock your LUKS disk encryption with an onscreen touch keyboard, you can use unl0kr.
+
+This module will automatically included the necessary kernel modules in initrd to make touchpad and touchscreen work when `boot.initrd.unl0kr.enable = true`.
+
+Example configuration:
+
+```nix
+{
+  boot.initrd.systemd.enable = true;
+  boot.initrd.unl0kr.enable = true;
+
+  boot.initrd.luks.devices."cryptroot" = {
+    device = "/dev/disk/by-uuid/YOUR-UUID-HERE";
+  };
+}
+```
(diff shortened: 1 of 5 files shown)


framework/12-inch/intel-ultra-s3/README.md
```diff
@@ -0,0 +1,33 @@
...
+Everything is updateable through fwupd, so it's enabled by default.
+
+To get the latest firmware, run:
+
+```sh
+$ fwupdmgr refresh
+$ fwupdmgr update
+```
+
+- [Latest Update](https://example.com/TODO)
```
JohnAZoidberg, line 14: Currently no firmware update available. Factory will ship the latest firmware.

JohnAZoidberg:
Other dependencies are in nixpkgs already:

- https://github.com/NixOS/nixpkgs/pull/488943
- https://github.com/NixOS/nixpkgs/pull/548976

Linux 7.2 has another audiojack/mic fix, but at this point probably too early to make people upgrade.

garretraziel:
I investigated the failing validation and reproduced it locally. Both failing checks (`stable-framework-12-intel-ultra-s3` and `unstable-framework-12-intel-ultra-s3`) fail because the new module uses `pkgs` without accepting it as a module argument:

```text
error: undefined variable 'pkgs'
at framework/12-inch/intel-ultra-s3/default.nix:9:52
```

The immediate fix is:

```diff
-{ config, lib, ... }:
+{ config, lib, pkgs, ... }:
```

After applying that change locally, the new profile evaluates under both the stable and unstable test inputs.

I also noticed a few things worth addressing before merge:

* Framework calls this product **Intel Core Series 3**, not Core Ultra Series 3. The announcement explains that it is derived from Core Ultra Series 3, but its actual branding is Core Series 3: https://frame.work/blog/framework-laptop-12-now-with-core-series-3. It may be better to rename the directory and public flake attribute before users depend on them, e.g. `intel-core-series3` / `framework-12-intel-core-series3`.
* The added `softdep soc_button_array post: ...` declarations appear to express the opposite of the stated ordering requirement. If pinctrl must load before `soc_button_array`, that would be `pre:`, not `post:`. They are also inside the `< 6.13` kernel condition, so they cannot affect this new model, which requires kernel 7.1 or newer. Since the profile already puts `pinctrl_intel_platform` in `boot.initrd.kernelModules`, the softdep additions may simply be unnecessary.
* `services.fprintd.enable` should probably use `lib.mkDefault true`, as the other Framework profiles do, so users can disable it normally.
* Explicitly adding `fprintd` to `environment.systemPackages` is redundant; the NixOS fprintd module already adds `services.fprintd.package` there.
* The README still contains `https://example.com/TODO`; it would be better to remove that line until a firmware page exists.
* Since Framework documents kernel 7.1 as the minimum, an assertion on the resulting selected kernel version could prevent an older pinned nixpkgs from silently selecting a `linuxPackages_latest` that is still too old.

Also, maintainer edits are currently disabled on this PR. Enabling them would let a nixos-hardware maintainer apply small fixes directly.

PR #2030, opened 2026-09-09
pc/laptop: do not enable tlp if tuned is enabled

thiagokokada:
###### Description of changes

TuneD is a maintained alternative for power-profile-daemon. If the user manually enables it we should not try to enable TLP, otherwise they will conflict with each other.

###### Things done


thiagokokada:
CC @SuperSandro2000.

SuperSandro2000 (approved):

PR #2034, opened 2026-09-10
raspberry-pi/kernel: enable kexec

fkokosinski:
###### Description of changes

This PR enables `kexec` in the kernel. This enables the kernel to be used with e.g. [nixos-anywhere](https://github.com/nix-community/nixos-anywhere).

###### Things done

### DIFF
--- a/raspberry-pi/common/kernel.nix
+++ b/raspberry-pi/common/kernel.nix
@@ -70,6 +70,11 @@ in
       NET_CLS_BPF = lib.mkForce yes;
       NLS_CODEPAGE_437 = lib.mkForce yes;
       FB_SIMPLE = yes;
+
+      # Required by nixos-anywhere
+      KEXEC = yes;
+      KEXEC_FILE = yes;
+      PROC_KCORE = yes;
     }
     # nixpkgs defaults to lazy preempt on kernel version >=6.18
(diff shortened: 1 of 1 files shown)


fkokosinski:
@JamieMagee @doronbehar -

Could you take a look at this PR in your spare time, since you've been active in this area recently? Thanks!

JamieMagee:
Why do you need to run nixos-anywhere from a Raspberry Pi already running NixOS? I thought that was for converting other Linux Distros to NixOS.

EDIT: I generated the kernel configs before and after this change. On 64-bit Pi 3, 4, and 5 systems, `CONFIG_KEXEC_FILE` is already enabled, `CONFIG_KEXEC` is ignored because its arm64 dependency is not met (it depends on `PM_SLEEP_SMP`), and the only effective change is `CONFIG_PROC_KCORE=y`.

Pi 2 gets no effective change, while Pi 1 also gains the crash dump options selected through `KEXEC_CORE`.

Since the nixos-anywhere image uses `--kexec-syscall-auto` on this kernel, it should already try `kexec_file_load()` first. If that fails, enabling `/proc/kcore` alone does not make the fallback work because `CONFIG_KEXEC` is still disabled.

Could you include the exact kexec failure output and a before and after runtime test showing which syscall requires these settings? Without that, I do not think these overrides should be merged.

PR #2044, opened 2026-09-20
raspberry-pi/firmware: stage vendor overlays independent of dtbSrc

hatch01:
## Description of changes

`install-rpi-firmware` uses the same `$dtbSrc` variable to decide both
where the base `.dtb` files come from and where the `overlays/`
directory it copies to the firmware partition comes from.

When `useGenerationDeviceTree = true`, `$dtbSrc` is
`/run/current-system/dtbs`. That path never contains an `overlays/`
subdirectory unless the generation was built with
`hardware.deviceTree.overlays`, so the overlay-staging step becomes a
no-op, and the stale-file prune then removes any vendor `.dtbo` that
might already be present in the firmware partition's `overlays/`.

Concretely, this means `hardware.raspberry-pi.configtxt.deviceTreeOverlays`
never works once `useGenerationDeviceTree = true` — which is exactly
the combination needed for a Pi 4 config that requires
`apply-overlays-dtmerge` (e.g. `tc358743`, which needs the generation's
dtmerge-produced dtb as the base since U-Boot keeps whatever dtb the
firmware handed it) together with a stock `config.txt` overlay like
`dwc2`. Right now that combination silently drops the `dwc2` overlay
on every switch.

This PR decouples the two: vendor overlays are staged from
`$firmwareBoot/overlays` unconditionally (they're matched against
`configtxt.deviceTreeOverlays` entries by filename, independent of the
base dtb source), and `$dtbSrc/overlays` is staged in addition when it
differs from `$firmwareBoot`, so overlays produced on the generation
side via `hardware.deviceTree.overlays` still take precedence.

This doesn't touch the default (`useGenerationDeviceTree = false`)
case at all — it only fixes the currently-broken `true` case. It's a
narrow fix, not an attempt to resolve the broader tension between the
two device-tree mechanisms tracked in #1946.

Related: #1946

## Things done

- [x] Tested the changes end-to-end by using your fork of nixos-hardware and
     (tested on github.com/hatch01/nixos-pikvm)


doronbehar:
cc @JamieMagee

